PR #1, opened 2018-04-17
update amis to 2017.09.1

Vad1mo:
The patch updates to The latest Amazon Linux AMI 2017.09.1 -  released on 2018-01-17.
https://aws.amazon.com/amazon-linux-ami/


npalm:
@Vad1mo thanks for the update, sorry for the late reply. I was on a long term leave.

PR #21, opened 2018-11-13
make iam instance profile configurable

msvechla:
Hello, 

Thanks for this awesome module, this saved us a lot of work! This PR adds the possibility to specify the `runners_iam_instance_profile`, which will attach an IAM instance profile to the spawned runners.

I tested it already and it works, as long as the permissions of the gitlab-runner role are adjusted.


npalm:
@msvechla thanks for the change, looks all good! I only prefer to rename `runners_iam_instance_profile` to `runners_iam_instance_profile_name` so it is clear the name should be provided an not the arn.

msvechla:
@npalm correct, thats what I meant with " as long as the permissions of the gitlab-runner role are adjusted". 

I will update the PR accordingly. Thanks for getting back that quickly!

msvechla:
I just pushed the changes. I also added another variable for specifying the `runners_pre_build_script`, which would have been another PR from me.

With those changes we are now able to inject the IAM role of the executors to the docker containers, which is why we created this PR in the first place.

Please check it out and let me know what you think.

msvechla:
@npalm thanks for working on this that quickly!

PR #23, opened 2018-12-03
Add instance profile to spot instances and auto register

PatrickWolleb:
Provides better security by adding instance profiles to docker machine command in config.toml.
Provides better developer experience by adding gitlab registration command in user-data.
Provides capability of adding additional docker volume mounts


PatrickWolleb:
Great @npalm, thanks.
It is still needs finalising. Testing different configs.

npalm:
> Great @npalm, thanks.
> It is still needs finalising. Testing different configs.

@PatrickWolleb did you was able to finish testing?

PatrickWolleb:
I am have new approach of storing the runner token Secure Param. Store. I will try to finalise over Christmas.

npalm:
@PatrickWolleb Thx!

PR #25, opened 2018-12-14
Add `cache_bucket_prefix`, to allow for unique bucket names

rsrchboy:
Sometimes you may want to run more than one of these instances per AWS account :)  Allowing an optional prefix enables unique bucket names to be configured, without breaking existing installations.


npalm:

PR #26, opened 2018-12-20
Add pre/post install user-data snippets for runners

rsrchboy:
This commit introduces two new variables, intended to be used to allow
scriptlets to be inserted before and after the runner install (but
before the runner is started) in the user-data configuration.


npalm:
@rsrchboy thx will check the PR in the next days

npalm:

PR #27, opened 2018-12-28
Configurable docker image for runner

mkalmykov:
Hi, @npalm!

This PR introduces configurable docker image for runners.

In our case we need the ability to use custom images, because we work with private docker registries and need to configure credential helpers to perform pulls from them. To handle this, I've added the variable `runners_image` with default value `docker:18.03.1-ce`. So without setting this variable everything should work as before.


PanfilovDenis:
+1 I really need it too.

npalm:
@mkalmykov thx will check your PR in the next days

mkalmykov:
@npalm Thank you for the response. We've noticed that we need one more item to be configurable. I will add the code tomorrow and notify you when PR is ready.

mkalmykov:
@npalm 

We've found that there is already a PR #26 with changes that we need, so this PR is ready.

npalm:

PR #32, opened 2019-02-11
Add possibility to use simple docker executor and multiple subnets for gitlab runner

msvechla:
Hello, 

as we are only running some lightweight tasks on AWS with gitlab-runners, we need a more simplified setup by simply using the `docker` executor, rather than using the full-blown `docker+machine` solution.

Therefore I made the executor configurable and left the default with `docker+machine`.
In addition I had to install docker on the gitlab-runner instance.

If this gets merged, there probably should be 1-2 sentences about this in the readme, as this is not using spot instances, and only a single gitlab-runner instance with docker. For lightweight use-cases this would be perfect though. Let me know what you think!

As we also need AZ failover, I adapted `subnet_id_gitlab_runner` to use an array of subnets. This worked fine on our instance. I also adjusted the examples accordingly, however I did not test them.


npalm:
Thanks for taking the time to submit the change. I will do my best to review the PR in the next week. Please can you check the build failure as well. The build only verify the formatting and validate the terraform (aka lint).

template/gitlab-runner.tpl
```diff
@@ -13,8 +13,11 @@ curl -L https://github.com/docker/machine/releases/download/v${docker_machine_ve
   chmod +x /tmp/docker-machine && \
   cp /tmp/docker-machine /usr/local/bin/docker-machine && \
   ln -s /usr/local/bin/docker-machine /usr/bin/docker-machine
+yum install docker -y
```
npalm: Can you make this port conditional?

variables.tf
```diff
@@ -54,6 +54,12 @@ variable "runners_name" {
   type        = "string"
 }
 
+variable "runners_executor" {
```
npalm: Can you add an dedicated example for newly added executor, looks like you can simply base the example on the default and then just enable the docker

npalm: FIxed

template/gitlab-runner.tpl
npalm: Fixed

msvechla:
Oh thanks a lot for your work here! I will have a little bit more time in a few days if this needs additional work.

Thanks a lot!

npalm:
Thanks for the contribution. Your PR is merged manually into develop. I will close the PR

PR #52, opened 2019-04-05
allow setting ENV variables in runner config

reach3r:
Hi there,

I think the title basically says it all. We needed to be able to set environment variables for the runners and thought it might be useful for the project as well.

If this PR needs more work, please let me know.

Best regards,
Dominik

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -158,6 +158,7 @@ data "template_file" "runners" {
     runners_root_size                 = "${var.runners_root_size}"
     runners_iam_instance_profile_name = "${var.runners_iam_instance_profile_name}"
     runners_use_private_address       = "${var.runners_use_private_address}"
+    runners_environment               = "${jsonencode(var.runners_environment)}"
     runners_pre_build_script          = "${var.runners_pre_build_script}"
     runners_post_build_script         = "${var.runners_post_build_script}"
     runners_pre_clone_script          = "${var.runners_pre_clone_script}"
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -158,6 +158,7 @@ data "template_file" "runners" {
     runners_root_size                 = "${var.runners_root_size}"
     runners_iam_instance_profile_name = "${var.runners_iam_instance_profile_name}"
     runners_use_private_address       = "${var.runners_use_private_address}"
+    runners_environment               = "${jsonencode(var.runners_environment)}"
```
npalm: May suggest to rename the var to `runners_environment_vars` slight more clear.

variables.tf
```diff
@@ -151,6 +151,12 @@ variable "runners_iam_instance_profile_name" {
   default     = ""
 }
 
+variable "runners_environment" {
+  description = "Environment variables in the pipeline, will be used in the runner config.toml"
```
npalm: Could be helping to add here a sample in the description, such as: For example `["VAR=VALUE"]`

reach3r: I would like to do that, but I run into trouble:
If I put a hcl list into the description that's a syntax error, then if I escape it it's not anymore but it will look wrong in the generated docs :(
```hcl
variable "runners_environment_vars" {
  description = "Environment variables during build execution, e.g. `[\"KEY=Value\",\"TOKEN=xyz\"]`, see runner-public example. Will be used in the runner config.toml"
  type        = "list"
  default     = []
}
```
Any advice?

main.tf
reach3r: Good idea, done.

Issue #47, opened 2019-04-01
Disable SSH access for runner instance


npalm:
Add an option to disable SSH (ingress) connection to the runner instance. The default could be disable access.

PR #55, opened 2019-04-11
Add option to disable SSH to the runner instance

arrpee:
This PR addresses issue #47 and also adds a way of limiting SSH ingress to a list of CIDR blocks.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -3,33 +3,33 @@ resource "aws_key_pair" "key" {
   public_key = "${var.ssh_public_key}"
 }
 
+locals {
+  # workaround for "conditional operator cannot be used with list values"
+  runner_ssh_config = {
+    enabled  = "${var.gitlab_runner_ssh_cidr_blocks}"
+    disabled = "${list()}"
+  }
+}
+
 resource "aws_security_group" "runner" {
   name_prefix = "${var.environment}-security-group"
   vpc_id      = "${var.vpc_id}"
 
+  ingress {
+    from_port   = 22
+    to_port     = 22
+    protocol    = "tcp"
+    cidr_blocks = "${local.runner_ssh_config["${var.enable_gitlab_runner_ssh_access == 1 ? "enabled" : "disabled"}"]}"
+  }
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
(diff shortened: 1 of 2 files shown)


npalm:
@arrpee thanks for the PR will check soon in detail. A quick check results in the follow error while updating my current deployment:

Reproduce as follow:
```
git checkout develop
cd examples/runners-public
terraform init && terraform apply
```
Next merge in your PR
```
terraform apply
```
error:
```

Error: Error applying plan:

1 error(s) occurred:

* module.runner.aws_security_group_rule.out_all_runner: 1 error(s) occurred:

* aws_security_group_rule.out_all_runner: [WARN] A duplicate Security Group rule was found on (sg-03e725283069d8ec2). This may be
a side effect of a now-fixed Terraform issue causing two security groups with
identical attributes but different source_security_group_ids to overwrite each
other in the state. See https://github.com/hashicorp/terraform/pull/2376 for more
information and instructions for recovery. Error message: the specified rule "peer: 0.0.0.0/0, ALL, ALLOW" already exists

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

arrpee:
I can't think of a clean way to make this change backward compatible without forcing terraform to recreate the auto scaling group and the launch configuration. Since this security group is declared in the attribute-as-blocks mode, we'll probably have to wait until terraform 0.12 arrives with dynamic blocks to make this change. Thoughts?

npalm:
@arrpee the only option I see for the short term is to remain the inline blocks. And add the variable to the inline block:
```
resource "aws_security_group" "runner" {
  name_prefix = "${var.environment}-security-group"
  vpc_id      = "${var.vpc_id}"

  ingress {
    from_port   = 22
    to_port     = 22
    protocol    = "tcp"
    cidr_blocks = ["${var.gitlab_runner_ssh_cidr_blocks}"]
  }

  egress {
    from_port   = 0
    to_port     = 0
    protocol    = "-1"
    cidr_blocks = ["0.0.0.0/0"]
  }

  tags = "${local.tags}"
}
```

Just tested it and it works.

arrpee:
@npalm I think I've got a working change which is backward compatible, keeps both variables and doesn't recreate everything. Can you test it out?

Issue #75, opened 2019-05-31
Runner cannot reach Internet in public subnet


int128:
I am trying to use this module in the public subnet. It seems the runner EC2 instance cannot reach the Internet because it does not have a public IP address due to the following config:

https://github.com/npalm/terraform-aws-gitlab-runner/blob/2f45f1d73ba0cffd310a859cf30f08e88b8a1b03/main.tf#L218

I changed the above to `true` and finally the runner has been registered successfully. Is this expected design? Some extra config needed?

Thank you for the great work.

int128:
I think it may be nice to respect `runners_use_private_address` as follows:

```tf
  associate_public_ip_address = "${!var.runners_use_private_address}"
```

npalm:
@int128 thanks for reporting the issue. Personally I use the runner only in a VPC with private / public setup. So never face the issue. The sample in the example dir (https://github.com/npalm/terraform-aws-gitlab-runner/blob/develop/examples/runner-public/main.tf#L24) is configures so it uses public IP address: `runners_use_private_address = false`

If I remember correctly this will assign also public ip addresses to the docker-machines.

Feel free to suggest your improvement in a PR

int128:
Yes, that will assign a public IP address to the docker-machines, but it will not do to the runner.
I will open a pull request for this later.

int128:
It turned out that the runner-public example works well because the module `terraform-aws-modules/vpc/aws` assigns a public IP by default.

https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/158b7d79ece06815cea20e64aa5bd15a5b197efe/variables.tf#L649-L653.

I think it should assign a public IP regardless of the subnet config when `runners_use_private_address 
 = false`.

npalm:
@int128 Yes I agree.

I suggest we update the public example with the `map_public_ip_on_launch` set to `false` and implement on the runner your suggestion. Let me know if you would like to create a PR?

int128:
This issue occurs only if we create a subnet manually and does not occur with the module `terraform-aws-modules/vpc/aws`. So we don't need to change the public example.

Would you take a look at the PR?

npalm:
Thanks will do in the next days

PR #79, opened 2019-06-04
Fix runner is not reachable when runners_use_private_address = false

int128:
## Description
This fixes the issue #75 the runner is not reachable when the following conditions satisfy:

- `runners_use_private_address = false`
- the subnet is public (i.e. the default route is to an Internet Gateway)
- `map_public_ip_on_launch` of the subnet is default

## Migrations required
No.

## Verification
Tested with the example project: https://github.com/int128/aws-gitlab-runner-starter.

## Documentation
No change.


npalm:
Fix #75

PR #86, opened 2019-06-18
Fix misleading variable description

solutionDrive-Alt:
## Description
changes the description for the variable `enable_manage_gitlab_token`, since the second part of the description is misleading

## Migrations required
NO

## Verification
tried to build the infrastructure with this module and believed the wrong part of the description. So, I had chosen the wrong value for this variable.

## Documentation
README was updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ terraform destroy
 | docker\_machine\_version | Version of docker-machine. | string | `"0.16.1"` | no |
 | enable\_cloudwatch\_logging | Boolean used to enable or disable the CloudWatch logging. | string | `"true"` | no |
 | enable\_gitlab\_runner\_ssh\_access | Enables SSH Access to the gitlab runner instance. | string | `"false"` | no |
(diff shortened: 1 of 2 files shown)


npalm:
@solutionDrive-Alt thanks for taking the time with creating the pull request. I also checked the code again.

I think the name of the variable `enable_manage_gitlab_token` is ok, but I agree we can come up with a better description. True means the SSM parameter is managed by terraform. False means that the parameter in SSM is still used by the the user-data script but the resource is in SSM is not managed. So once a terraform destoy is applied the token will remain in SSM. In case of true i will be remove.

What doe you think of the following proposal?


Boolean to enable the management of the GitLab token in SSM. If `true` the token will be stored in SSM however, which means the SSM property is a terraform managed resource. If `false` the Gitlab token will be stored in the SSM by the user-data script which during creation of the the instance. However the SSM parameter is not managed an will remain in SSM after a terraform delete.

solutionDrive-Alt:
@npalm no problem. I will probably work some more on our forked module. Maybe there will also be some changes this module can benefit from.

Now I understand this variable much better and I agree that your proposal for the description fits better than my just changed description. I will updated this PR later that day.

npalm:
@solutionDrive-Alt thanks, looking forward to other improvements. I also have a branch prepared to move to TF 0.12. I will keep a branch 0.11 available.

PR #88, opened 2019-07-02
use latest ubuntu 18.04 for gitlab runner instances

roock:
## Description
The default image used for the docker-machine EC2 instances is rather old. In this PR this is changed to use the latest AMI of Ubuntu 18.04 available on AWS EC2.

## Migrations required
NO

## Verification
Works without problems in our installation for a few weeks

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -147,6 +147,7 @@ data "template_file" "runners" {
     runners_subnet_id                 = "${var.subnet_id_runners}"
     runners_aws_zone                  = "${var.aws_zone}"
     runners_instance_type             = "${var.docker_machine_instance_type}"
+    runners_ami                       = "${data.aws_ami.docker-machine.id}"
     runners_spot_price_bid            = "${var.docker_machine_spot_price_bid}"
     runners_security_group_name       = "${aws_security_group.docker_machine.name}"
     runners_monitoring                = "${var.runners_monitoring}"
@@ -197,6 +198,14 @@ resource "aws_autoscaling_group" "gitlab_runner_instance" {
         data.null_data_source.tags.*.outputs,
         list(map("key", "Name", "value", local.name_runner_instance, "propagate_at_launch", true)))}",
   ]
+}
+
+data "aws_ami" "docker-machine" {
+  most_recent = true
+
+  filter = "${var.runner_ami_filter}"
+
+  owners = ["${var.runner_ami_owners}"]
 }
 
 data "aws_ami" "runner" {
--- a/variables.tf
+++ b/variables.tf
@@ -315,6 +315,24 @@ variable "ami_owners" {
   default     = ["amazon"]
 }
 
+variable "runner_ami_filter" {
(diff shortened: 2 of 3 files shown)


npalm:
@roock thanks for the improvement. I will try it the next days. Aksed my self the question should a be a configuration parameter or not?

npalm:
@roock works like a charm, please can you have a look on the one comment I made. Just about the naming.

roock:
@npalm which comment about naming do you mean?
about adding a parameter for this: This came also to my mind, I'll try to add it to the MR

npalm:
@roock I mean the name of the terraform resource 

```data "aws_ami" "latest-ubuntu" {
 ``

I think name the data resource `docker-machine-ami` is a more clear name. And indeed having the filter configurable would be great.

**Comment by roock** [comment]
@npalm  I've added a parameter for the docker machine ami defaulting to latest Ubuntu 16.04. Or should we set default to Ubuntu 18.04?
also changed the name of the terraform resource as requested.

File: variables.tf [hunk 1]
```diff
@@ -315,6 +315,22 @@ variable "ami_owners" {
...
+  default = [{
+    name   = "name"
+    values = ["ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*"]
+  }]
+}
+
+variable "runner_ami_owners" {
+  description = "The list of owners used to select the AMI of Gitlab runner docker-machine instances."
+  type        = "list"
+  default     = ["Canonical"]
```
npalm: A look up by name is not supported, I get the following error
```
Error: Error refreshing state: 1 error occurred:
	* module.runner.data.aws_ami.docker-machine: 1 error occurred:
	* module.runner.data.aws_ami.docker-machine: data.aws_ami.docker-machine: InvalidUserID.Malformed: Invalid user id: "Canonical"
	status code: 400, request id: 
```

Replacing Canonical by 099720109477 does the trick. Suggest that you replace Canonical by the id and add a comment line that it is the id of Canonical

main.tf
```diff
@@ -198,6 +199,22 @@ resource "aws_autoscaling_group" "gitlab_runner_instance" {
   ]
 }
 
+data "aws_ami" "latest-ubuntu" {
```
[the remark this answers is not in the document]
roock: already changed to "docker-machine"

variables.tf
roock: fixed in b08e1f8e5cb641c8d6ceb95cff2ce292bce1aee7

npalm:
@roock Cool for creating this PR. I dit a quick check and now the runners are using `ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20180228.1 (ami-0b541372)` instead of `ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20190628 (ami-03746875d916becc0)
`. Which means it looks like the default behaviour is not matching the docs at the docker site (https://docs.docker.com/machine/drivers/aws/) Which mention the default should be the daily ubuntu image.

PR #89, opened 2019-07-02
Adding docker pull policy Parameter

roock:
## Description
Adding additional parameter to specify the Docker Pull Policy for the Gitlab Runner

## Migrations required
NO

## Verification
has been in production in our setup for a few weeks with parameter if-not-present. default value is the default from Gitlab

## Documentation
Documentation was updated to include the additional parameter. Also it was updated from upstream and fixed minor autodoc issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Terraform module for GitLab auto scaling runners on AWS spot instances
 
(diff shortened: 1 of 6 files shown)


npalm:
@roock thanks, look good will check the change in detail the next days.

PR #93, opened 2019-07-09
fix: availability zone in AWS region

theBenForce:
## Description
Use valid availability zones when using a region other than `eu-west-1`.

## Migrations required
NO - If yes please describe the mirgration.

## Verification
Please mentioned the examples you have verified.

## Documentation
Please ensure you pdate the README in `_docs/README.md`. The README.md in the root can be update by running the script `ci/bin/autodocs.sh`

### DIFF
--- a/examples/runner-default/main.tf
+++ b/examples/runner-default/main.tf
@@ -5,7 +5,7 @@ module "vpc" {
   name = "vpc-${var.environment}"
   cidr = "10.0.0.0/16"
 
-  azs             = ["eu-west-1a"]
+  azs             = ["${var.aws_region}a"]
   private_subnets = ["10.0.1.0/24"]
   public_subnets  = ["10.0.101.0/24"]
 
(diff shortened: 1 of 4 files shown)


examples/runner-default/main.tf
```diff
@@ -5,7 +5,7 @@ module "vpc" {
   name = "vpc-${var.environment}"
   cidr = "10.0.0.0/16"
 
-  azs             = ["eu-west-1a"]
+  azs             = ["${var.aws_region}a"]
```
npalm: maybe even a more generic approach it the next one:

```
data "aws_availability_zones" "available" {
  state = "available"
}
module "vpc" {
  ...
  azs = ["${data.aws_availability_zones.available.names[0]}"]
}
```

What do you think?

theBenForce: That looks like a much better way of doing it. I'll update my pull request with these changes.

PR #95, opened 2019-07-11
Mount docker socket

hatemosphere:
## Description
This will add an optional Docker socket mount as an alternative (opinionated to be better) option to privileged mode

## Migrations required
NO

## Verification
It works with default example in our setup

## Documentation
Updated

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,10 +2,10 @@
 A few sentences describing the overall goals of the pull request's commits.
 
 ## Migrations required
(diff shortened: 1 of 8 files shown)


npalm:
Great contribution @hatemosphere 
What do you think to update the runner-public example to use the socket instead of the privileged mode?

hatemosphere:
@npalm done that! please let me know if you want to change something else. sorry for additional commits, was a bit lazy to open more smaller PRs while testing :)

PR #98, opened 2019-07-16
Add ssh_key_pair override

gordonbondon:
## Description
Allow using existing key pair instead of generated

## Migrations required
NO


npalm:
@gordonbondon nice addition, thanks!

PR #99, opened 2019-07-19
Added IAM outputs, additional volumes mounts into runner, fix ssh var name

alexandrst88:
## Description
A few sentences describing the overall goals of the pull request's commits.
- Added option to mount additional volumes
- Added output for : runner_agent_role_arn, runner_role, runner_role_name, runner_role_arn
- Fix for ssh var name

## Migrations required
No

## Verification
- default and public

## Documentation
Please ensure you update the README in `_docs/README.md`. The README.md in the root can be update by running the script `ci/bin/autodocs.sh`


outputs.tf
```diff
@@ -8,12 +8,27 @@ output "runner_cache_bucket_arn" {
...
+  value       = aws_iam_role.instance.name
+}
+
+output "runner_role_arn" {
   description = "ARN of the rule used for the docker machine runners."
   value       = aws_iam_role.docker_machine.arn
 }
+
+output "runner_role_name" {
+  description = "Name of the rule used for the docker machine runners."
```
npalm: rule -> role (type)

outputs.tf
```diff
@@ -8,12 +8,27 @@ output "runner_cache_bucket_arn" {
...
+}
+
+output "runner_agent_role_arn" {
   description = "ARN of the rule used for the ec2 instance for the GitLab runner agent."
   value       = aws_iam_role.instance.arn
 }
 
-output "runner_role" {
+output "runner_agent_role_name" {
+  description = "Name of the rule used for the ec2 instance for the GitLab runner agent."
```
npalm: Typo rule -> role

outputs.tf
```diff
@@ -8,12 +8,27 @@ output "runner_cache_bucket_arn" {
...
+}
+
+output "runner_role_arn" {
   description = "ARN of the rule used for the docker machine runners."
   value       = aws_iam_role.docker_machine.arn
 }
+
+output "runner_role_name" {
+  description = "Name of the rule used for the docker machine runners."
+  value       = aws_iam_role.docker_machine.arn
```
npalm: The ARN is returned instead of the name. Name of variable not matching the return value

template/runner-config.tpl
```diff
@@ -18,7 +18,7 @@ check_interval = 0
     image = "${runners_image}"
     privileged = ${runners_privileged}
     disable_cache = false
-    volumes = ["/cache"%{ if runners_mount_docker_socket == "true" },"${runners_docker_socket}"%{ endif ~}]
+    volumes = ["/cache"${runners_additional_volumes}]
```
npalm: Thanks, I like the more generic approach.

variables.tf
```diff
@@ -121,16 +122,10 @@ variable "runners_privileged" {
...
-}
-
-variable "runners_docker_socket" {
-  description = "Location of Docker socket on host if socket mount if enabled, will be used in the runner config.toml"
-  type        = "string"
-  default     = "/var/run/docker.sock:/var/run/docker.sock"
+variable "runners_additional_volumes" {
+  description = "Additional volumes that will be used in the runner config.toml, e.g Docker socket"
+  type        = list
+  default     = []
```
npalm: Maybe we could update the runner-public example to add an example usages

main.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_key_pair" "key" {
-  count      = var.ssh_key_pair == "" ? 0 : 1
+  count      = var.ssh_public_key == "" ? 0 : 1
```
npalm: `var.ssh_key_pair` was introduced to re-use existing keys. If set no key will be generated by the module. Your change looks like doing the same but for some reason the default example fails no.

```
Error: Invalid count argument

  on ../../main.tf line 2, in resource "aws_key_pair" "key":
   2:   count      = var.ssh_public_key == "" ? 0 : 1

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

npalm: I solved the problem with key on in a slight other way.

hatemosphere: @npalm so what we should leave in this PR and what should be removed to resolve the conflict?

alexandrst88:
@npalm made the fixes that you requested, seems like build failed due to missed region

hatemosphere:
@npalm hey, any chances to cut a new tag from the current branch state so we can move from our fork to upstream?

npalm:
@hatemosphere new release 4.0.0_M4 - https://github.com/npalm/terraform-aws-gitlab-runner/releases/tag/4.0.0_M4

I hope to switch in the coming weeks by default to the TF 0.12 branch and release the 4.0.0 version.

PR #101, opened 2019-07-19
Add variable docker_machine_ssh_cidr_blocks allowing ssh ingress restriction

kevinrambaud:
## Description
My company uses CloudSploit (https://cloudsploit.com/) to monitor our AWS account, the current configuration causes alerts because of the ingress configuration of the docker machine instance. As such, I added the `docker_machine_ssh_cidr_blocks` to allow the configuration of this value like the `gitlab_runner_ssh_cidr_blocks` variable does.

## Migrations required
No

## Verification
Deployed on a private instance.

## Documentation
Documentation has been updated.


npalm:
@kevinrambaud fully agree the docker machine runners do not require world ingress in general. So good catch. I also spot that the current naming is not that clear. For example `ssh` for the resource meant for the docker machine.

bishtawi:
@kevinrambaud @npalm What value do you guys recommend we set on the cidr? My company's infosec team also does not like the open ingress security group on ports 2376 and 22 but I dont know what value I should be setting. What needs SSH access? What needs docker port access?

PR #105, opened 2019-07-24
Add runners_volumes variable

kevinrambaud:
## Description
One of the needs of my team is to build docker image from Gitlab, to do that we currently use this option https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#use-docker-socket-binding and therefore need to be able to customize the volumes set in the `config.toml` template.

This pull request adds support for that and default to `["/cache"]` if not provided to support the existing behavior.

Usage in my case:

```
runners_volumes = ["/cache", "/var/run/docker.sock:/var/run/docker.sock"]
```

## Migrations required
No

## Verification
Deployed and validated on a private gitlab instance.

## Documentation
Updated.


npalm:
Thanks @kevinrambaud will check the PR later today

npalm:
Verified with default example.

PR #113, opened 2019-08-06
Cache bucket name optional account

Glen-Moonpig:
## Description

The IAM roles that we use for deployment require us to use certain resource name prefixes, along with the account ID and suffix this was pushing the cache bucket name over the 63 character limit imposed by AWS.
I have added an input variable to make the account ID in the bucket name optional as we did not require it, but wanted to maintain backwards compatibility for others.

## Migrations required
NO

## Verification

We have deployed the module to our environments.

## Documentation

### DIFF
--- a/cache/variables.tf
+++ b/cache/variables.tf
@@ -11,8 +11,8 @@ variable "cache_bucket_prefix" {
 
 variable "cache_bucket_name_include_account_id" {
   description = "Boolean to add current account ID to cache bucket name."
-  type        = string
-  default     = "true"
+  type        = bool
+  default     = true
 }
 
 variable "cache_bucket_versioning" {
--- a/cache/variables.tf
+++ b/cache/variables.tf
@@ -9,6 +9,12 @@ variable "cache_bucket_prefix" {
   default     = ""
 }
 
+variable "cache_bucket_name_include_account_id" {
+  description = "Boolean to add current account ID to cache bucket name."
+  type        = string
+  default     = "true"
+}
+
 variable "cache_bucket_versioning" {
   description = "Boolean used to enable versioning on the cache bucket, false by default."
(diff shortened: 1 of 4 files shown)


cache/variables.tf
```diff
@@ -9,6 +9,12 @@ variable "cache_bucket_prefix" {
   default     = ""
 }
 
+variable "cache_bucket_name_include_account_id" {
+  description = "Boolean to add current account ID to cache bucket name."
+  type        = string
```
npalm: Please update the type to `bool` (without quotes). Next you can remove the quotes form the default value as well.

variables.tf
```diff
@@ -261,6 +261,12 @@ variable "cache_bucket_prefix" {
   default     = ""
 }
 
+variable "cache_bucket_name_include_account_id" {
+  description = "Boolean to add current account ID to cache bucket name."
+  type        = string
```
npalm: Same here, update the type to bool

main.tf
```diff
@@ -269,6 +269,7 @@ module "cache" {
 
   create_cache_bucket     = var.cache_bucket["create"]
   cache_bucket_prefix     = var.cache_bucket_prefix
+  cache_bucket_name_include_account_id = var.cache_bucket_name_include_account_id
```
npalm: Can you run terraform fmt on `main.tf` this will format the code according hashicorp standards. CI build checks for this.

Glen-Moonpig: Formatting fixed

variables.tf
Glen-Moonpig: Changed to bool

cache/variables.tf
Glen-Moonpig: Changed to bool

PR #117, opened 2019-08-13
Switch to Go modules

loafoe:
## Description
Switch to Go modules

## Migrations required
NO

## Verification
Ran `go test ./...` and terraform ran for some time

## Documentation
Will update the README after you had a go with the changes as well.


npalm:
@loafoe thanks!

PR #122, opened 2019-08-19
Add MaxBuilds variable to gitlab runner config

gertjanmaas:
## Description
Adds the Gitlab Runner `MaxBuild` configuration option, which terminates instances that have run a certain number of builds.

## Migrations required
NO 

## Verification
Ran `runner-default` which works with both `0` `MaxBuild` (which disables the option) and `1` `MaxBuild` where the instance is destroyed after 1 build.

## Documentation
Updated :)

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -118,7 +118,7 @@ variable "runners_idle_count" {
 }
 
 variable "runners_max_builds" {
-  description = "Max builds for each runner after which it will be removed, will be used in the runner config.toml."
+  description = "Max builds for each runner after which it will be removed, will be used in the runner config.toml. By default set to 0, no maxBuilds will be set in the configuration."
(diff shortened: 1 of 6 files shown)


variables.tf
```diff
@@ -117,6 +117,12 @@ variable "runners_idle_count" {
   default     = 0
 }
 
+variable "runners_max_builds" {
+  description = "Max builds for each runner after which it will be removed, will be used in the runner config.toml."
```
npalm:
```suggestion
  description = "Max builds for each runner after which it will be removed, will be used in the runner config.toml. By default set to 0, no maxBuilds will be set in the configuration."
```

npalm:
@gertjanmaas thanks

Issue #80, opened 2019-06-05
Volume size


daanielcarvalho:
Hi

How can i change a volume size of the instance created? there is no option!

Thanks in advance
Daniel

npalm:
@daanielcarvalho you can set the variable `runners_root_size`. It is used for setting the size of the disk of the docker-machine instances.

daanielcarvalho:
Hi

I use runners_root_size = "30" and I still get 8gb machine running.
![Screenshot 2019-06-06 at 09 14 15](https://user-images.githubusercontent.com/6240971/59017539-a689af80-883b-11e9-8d49-d87ba2202faa.png)

Thanks in advance
Daniel

npalm:
Hi Daniel @daanielcarvalho 

Is this the instance that is used for the agent? Or is this an instance used for hosting the runners. Typically a spot instances managed via docker-machine? In case this spot instance disk is not changed, please can you check if the root size is set in the user_data script in the launch coniguration.

If you would like to change the size of the disk of the instance hosting the agent, you have to create a change in the source. Feel free to propose a PR

daanielcarvalho:
Hello

this is for the instances that will run the docker image with runner specifications!

Thanks
Daniel

npalm:
@daanielcarvalho a bit late response. Just verified and it works as documented. By setting `runners_root_size = 25` the seize of the volume is 25 GB for example

![image](https://user-images.githubusercontent.com/11609620/61958862-77cdc100-afc2-11e9-8f6c-ab6ce852854c.png)

npalm:
I will close the issue for now, Thanks for reporting

bsuv:
@npalm this is not resolved! I just set the value of `runners_root_size` in a v4.1.0 module instance and *still* got only 8 GB. 

I hope it's fine with you if I submit a PR replacing LCs with LTs and also using an input variable for setting root block device size at LT-time instead of boot time.

npalm:
@bsuv Thank you for reporting. The runner setup supports two modes, See the scenario's mentionin the readme.

In the first mode the builds are scheduled via a docker machine on spot instances. The `runners_root_size` sized is used for the spot instances hosting docker machine (see sample runner-default). In the thirds scenario builds are running on the the instance created by the auto scaling group via docker. for that instance the size is currently not configurable (see example runner-docker).

So feel free to propose PR's for:
- make the size of the runner agent instance configurable / instance used for build when executor is `docker`.
[...]

PR #123, opened 2019-08-20
Allow for configurable root block size

bsuv:
## Description
This fixes https://github.com/npalm/terraform-aws-gitlab-runner/issues/80 by allowing to use specific/dedicated input variables for configuring the EBS volume types and size attached to the instance.

## Migrations required
YES - while I'm setting the default to `8`GB drift might still be created depending on the existing/running launch configuration.

## Verification
`plan` and `apply`.

## Documentation

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -246,6 +246,11 @@ resource "aws_launch_configuration" "gitlab_runner_instance" {
   instance_type        = var.instance_type
   spot_price           = var.runner_instance_spot_price
   iam_instance_profile = aws_iam_instance_profile.instance.name
+  root_block_device {
+    volume_type = "${var.ec2_volume_type}"
+    volume_size = "${var.ec2_volume_size}"
+
+  }
 
   associate_public_ip_address = false == var.runners_use_private_address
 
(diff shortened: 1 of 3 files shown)


bsuv:
@npalm please have a look at this as I'm currently encountering this issue very frequently

main.tf
```diff
@@ -266,6 +266,11 @@ resource "aws_launch_configuration" "gitlab_runner_instance" {
   instance_type        = var.instance_type
   spot_price           = var.runner_instance_spot_price
   iam_instance_profile = aws_iam_instance_profile.instance.name
+  root_block_device {
```
npalm: This trigger a recreate as you mention which can be avoid by terraform 0.12 dynamic block. I have tested the following config.

```
  dynamic "root_block_device" {
    for_each = [var.runner_root_block_device]
    content {
      delete_on_termination = lookup(root_block_device.value, "delete_on_termination", true)
      volume_type           = lookup(root_block_device.value, "volume_type", "gp2")
      volume_size           = lookup(root_block_device.value, "volume_size", 8)
      iops                  = lookup(root_block_device.value, "iops", null)
    }
  }

```

And as variable add:
```
variable "runner_root_block_device" {
  type    = map(string)
  default = {}
}
```

Now you can set the root block device as following where you declare the module:
```
module "runner" {
   ....
  root_block_device = {
    volume_type = "gp2"
    volume_type = 8
  }
```

bsuv: Good idea! See follow-up commits please.

PR #132, opened 2019-09-10
Encryption security issues

hendrixroa:
## Description
Enable ebs encryption option, default:false

## Migrations required
NO

## Verification
NO

## Documentation
The docs are updated successfully with new variables description


hendrixroa:
TODO:
- user data encryption

bishtawi:
Does the MR as it is now complete the task of enabling encryption on the root EBS? If it does, can we merge this as it is now, and create a separate MR for encrypting user data (where does that exist, if not in the root device)? I only ask because I am looking to enable encryption on the EBS and it would be nice to get that in, if it really is just a single boolean. If the boolean by itself does nothing, and it requires additional work for it to actually encrypt, then carry on.

npalm:
@hendrixroa any update on this PR? Are you still working on it?

hendrixroa:
@npalm Yes, I'm testing with some security tools with this module, getting some insights from `scout-suite`

hendrixroa:
@npalm  I don't underatand why variables.tf has conflicts, can you review it pls?

npalm:
@hendrixroa I was not able to run the default example. I have made some comments. Please can you update the default example to so it works with KMS encryption. It looks like the changes in the cache module are missing. As well configuration for cloudwatch to grant access to the KMS keuy

npalm:
@hendrixroa inspired by this PR I have created a slightly different approach for the log encryption. Please could you have a look at PR #156

hendrixroa:
@npalm that is the right way to go, the encryption should be optional.

PR #134, opened 2019-09-23
Add access_level option to registration call in runner template

willychenchen:
## Description
Allows setting access_level to register protected runners. Protected runners will only run jobs triggered on protected branches allowing use of different IAM roles to control publish/deploy permissions available to branches.

## Migrations required
YES
The current gitlab_runner_registration_config pattern will require that all users add "access_level" to their config.
Example:
```
module "runner" {
    registration_token = var.registration_token
    tag_list = "docker_spot_runner"
    description = "runner public - auto"
    locked_to_project = "true"
    run_untagged = "false"
    maximum_timeout = "3600"
    access_level = "not_protected"
  }
}
```

## Verification
Verified runner-public for both `not_protected` and `ref_protected` values. Confirmed that runner is accurately registered to project as not protected and protected in each case.
One finding was that `terraform plan` fails due to aws_key_pair.key being unable to calculate `count` because `local_file.public_ssh_key.content` is not available during plan. Resolved by first running `terraform apply -target local_file.public_ssh_key` to make resource available during plan.

## Documentation
`_docs/README.md` updated to include `access_level` where other runner registration config values are found. `autodocs.sh` was executed and any changes were committed.


npalm:
@willychenchen thanks will check the PR later this week

npalm:
> @willychenchen thanks will check the PR later this week

Sorry too busy this week. Will check early next week. I also see the GitHub actions are not triggered on the PR. There looks like still some ref to travis :(

main.tf
```diff
@@ -128,6 +128,7 @@ data "template_file" "gitlab_runner" {
     gitlab_runner_locked_to_project         = var.gitlab_runner_registration_config["locked_to_project"]
     gitlab_runner_run_untagged              = var.gitlab_runner_registration_config["run_untagged"]
     gitlab_runner_maximum_timeout           = var.gitlab_runner_registration_config["maximum_timeout"]
+    gitlab_runner_access_level              = var.gitlab_runner_registration_config["access_level"]
```
npalm: We could change this line as follow so adding the property to the map become optional and the change will remain in that case backwards compatible

```
gitlab_runner_access_level              = lookup(var.gitlab_runner_registration_config, "access_level", "not_protected")
```

npalm:
@willychenchen I have changed the PR so the merge goes to `develop` the default branch

main.tf
willychenchen: Awesome, added! Great reason why I should look through the TF functions that are available.

willychenchen:
Validated the following 3 cases:
* no access_level key -> registered runner is not protected and no error/warning from TF
* access_level = 'not_protected' -> registered runner is not protected
* access_level = 'ref_protected' -> registered runner is protected
runner-public example now demonstrates the backwards compatible case of not providing access_level as well as the new capability of registering as protected.

PR #139, opened 2019-10-03
Add variable docker_machine_docker_cidr_blocks allowing docker ingress restriction

bishtawi:
## Description
Related to https://github.com/npalm/terraform-aws-gitlab-runner/pull/101, make the ingress CIDR configurable for the docker port 2376 security group rule.

## Migrations required
No, default value of the new variable is the original value.


bishtawi:
What I dont know, and would like to get some clarity on, is what is the minimal IP range that needs SSH and docker port access? Is it simply the Runner? If it is, instead of using `cidr_blocks`, could we use `source_security_group_id` and set that to `aws_security_group.runner.id`?

PR #140, opened 2019-10-04
Lock down docker port access to only the runner security group

bishtawi:
## Description
Alternative to https://github.com/npalm/terraform-aws-gitlab-runner/pull/139 as I dont think we need to expose the docker port to the world. Just to the gitlab runner instance and maybe the other docker machines?

## Migrations required
NO


bishtawi:
@npalm Does it make sense to make a similar change to the docker machine ssh security group rule? The docker machine instances only receive SSH connections from the runner right?

npalm:
@bishtawi yes the ssh connection is used by docker-machine on the runner agent instance. Feel free to make the same change for this security rule

PR #142, opened 2019-10-04
Add option to limit docker machine ssh ingress access to only the runner

bishtawi:
## Description
Similar to https://github.com/npalm/terraform-aws-gitlab-runner/pull/140, limit the docker machine ssh ingress access to only the runner via security groups instead of cidr ranges.

I kept backwards compatibility by making this feature optional (defaults to false) as I do not know if people actually need external SSH access to the docker machines. @npalm Do you think external SSH access is necessary? If it doesnt make sense, we could remove the CIDR based security rule and simply the code. Otherwise, this should work to support both cases.

## Migrations required
No

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -215,7 +215,7 @@ data "template_file" "runners" {
     runners_root_size                 = var.runners_root_size
     runners_iam_instance_profile_name = var.runners_iam_instance_profile_name
     runners_use_private_address_only  = var.runners_use_private_address
-    runners_use_private_address       = var.runners_use_private_address ? false : true
+    runners_use_private_address       = ! var.runners_use_private_address
     runners_environment_vars          = jsonencode(var.runners_environment_vars)
     runners_pre_build_script          = var.runners_pre_build_script
     runners_post_build_script         = var.runners_post_build_script
--- a/main.tf
+++ b/main.tf
@@ -47,18 +47,6 @@ resource "aws_security_group" "docker_machine" {
   )
 }
 
-resource "aws_security_group_rule" "docker_machine_docker_external" {
-  count = var.runners_use_private_address ? 0 : 1
-
-  type        = "ingress"
-  from_port   = 2376
-  to_port     = 2376
-  protocol    = "tcp"
-  cidr_blocks = var.docker_machine_docker_cidr_blocks
-
-  security_group_id = aws_security_group.docker_machine.id
-}
-
 resource "aws_security_group_rule" "docker_machine_docker_runner" {
   type                     = "ingress"
   from_port                = 2376
@@ -75,18 +63,6 @@ resource "aws_security_group_rule" "docker_machine_docker_self" {
   to_port   = 2376
   protocol  = "tcp"
   self      = true
-
-  security_group_id = aws_security_group.docker_machine.id
-}
-
-resource "aws_security_group_rule" "docker_machine_ssh_external" {
-  count = var.runners_use_private_address ? 0 : 1
-
-  type        = "ingress"
-  from_port   = 22
-  to_port     = 22
-  protocol    = "tcp"
-  cidr_blocks = var.docker_machine_ssh_cidr_blocks
 
   security_group_id = aws_security_group.docker_machine.id
 }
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -309,18 +309,18 @@ variable "gitlab_runner_ssh_cidr_blocks" {
   default     = ["0.0.0.0/0"]
 }
 
-variable "docker_machine_docker_cidr_blocks" {
```
bishtawi: FYI, this variable is not being used as the docker machine docker port ingress access is controlled via security group instead of CIDR.

npalm:
@bishtawi I think in only some debug cases there is a need for ssh acces, but in that case you use the runner instance to get into. Alternatively the session manager can be used. So I think you right the CIDR can be removed as well.

variables.tf
npalm: Thanks

main.tf
```diff
@@ -68,6 +68,8 @@ resource "aws_security_group_rule" "docker_machine_docker_self" {
 }
 
 resource "aws_security_group_rule" "docker_machine_ssh" {
+  count = var.docker_machine_ssh_restrict_ingress_to_runner ? 0 : 1
```
npalm: @bishtawi I like the work you did to get it backwards compatible. But your change is not breaking anything at all. I would suggest the we remove the variable `docker_machine_ssh_restrict_ingress_to_runner` and remove the cidr rule as well.

roock:
@npalm @bishtawi Did you check that this change also works for a public runner (this is also true for the change concerning the docker-machine port #140 ?
I tried to do something similar (https://github.com/roock/terraform-aws-gitlab-runner/commit/5e12e8851d16fe78d45e302f4e439a359fe80f0e), but it wasn't able to get it working using public IPs on the runners (security groups do not work with external IPs).
I think we need the docker-machine option "amazonec2-use-private-address" in this case (see 
https://docs.docker.com/machine/drivers/aws/ )

bishtawi:
@roock Yup, I tested my changes against a repo hosted on Gitlab.com and had no problems. My Gitlab runners were deployed on a private subnet in my VPC, but I dont see why it would be any different if they were in a public subnet. I dont see any reason why the docker machines need fully open public ingress access on ports 22 and 2376. The machines communicate solely with the runner ec2 box, and not with Gitlab.com. And unless I am mistaken, I am pretty sure the Gitlab Runners poll gitlab.com for CI jobs to run; Gitlab.com doesnt push anything to the runners, so I really dont see why unrestricted ingress access is necessary.

Looking at your changes, the only thing that is different is that I added a rule to allow `self` traffic through. Not sure if that matters? Or maybe I am misunderstanding what you mean by "public runner"?

bishtawi:
>  but it wasn't able to get it working using public IPs on the runners (security groups do not work with external IPs)

This line confuses me. Even if your gitlab runners were on the public subnet, you should still be able to use security groups to specify allowed ingress traffic. In fact, thats usually when security groups are the most important. Unless your docker machines are hosted in a different VPC than the runner? Or you arnt using a VPC? Hah, maybe I am just misunderstanding your comment.

bishtawi:
Welp, just tested it out on a public subnet and set `runners_use_private_address = false` and you are absolutely correct that it does not work. Did some [searching and yup](https://serverfault.com/questions/938758/aws-security-group-how-to-allow-public-ip-from-another-security-group), looks like using security groups as the source only allows private IP traffic and not public. You really do learn something new every day. 

@npalm How should we best handle this? Probably the best solution would be to introduce some variables allowing users to both either set CIDRs or enable locking ingress access via security group source.

bishtawi:
Alright, updated PR to handle both public and private IP addresses. Now variables `docker_machine_docker_cidr_blocks` and `docker_machine_ssh_cidr_blocks` allow users to set custom CIDR ranges, but these ranges are only used when `runners_use_private_address` is false. If `runners_use_private_address` is true then we use the security group source to limit ingress.

roock:
I think it would be best to use the private IP for communicating between Runner and Docker Machine in any case, also because traffic within the same VPC is cheaper and faster than going over the public IPs
see https://github.com/roock/terraform-aws-gitlab-runner/commit/64f9e7971232e03bdc8d85235d4b073c03a87b6e

bishtawi:
@npalm @roock Alright done, ready for review

npalm:
@bishtawi thanks, wil start checking the changes

npalm:
Just verified the change
- runner default works fine
- runner public (example) is broken. The count on the ssh key is not working. Making some small changes was able to get the sample working with running the runner on a public ip and the docker machines on a private or public one. Running both on private wasn't working.

PR #148, opened 2019-10-16
Add description forrunner_registration_token and replace underscore with dash

### LINKED ISSUE #87: Getting a error when runner_registration_token contains an underscore.
When the `aws_ssm_parameter.runner_registration_token` contains an underscore I'm getting the error:
`Error creating machine: Invalid hostname specified. Allowed hostname chars are: 0-9a-zA-Z . -`.
I solved it by recreating a runner to get a token without underscore. But maybe it makes sense to substitute underscore with a dash?

midnightrun:
Prevent tokens to contain underscore

## Description
Fixes #87 

## Migrations required
No

## Verification
Run code without any problems. Tested within `terraform console`.

## Documentation
Done.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -142,7 +142,7 @@ data "template_file" "gitlab_runner" {
     runners_token                           = var.runners_token
     secure_parameter_store_runner_token_key = local.secure_parameter_store_runner_token_key
     secure_parameter_store_region           = var.aws_region
-    gitlab_runner_registration_token        = var.gitlab_runner_registration_config["registration_token"]
+    gitlab_runner_registration_token        = replace(var.gitlab_runner_registration_config["registration_token"], "_", "-")
     giltab_runner_description               = var.gitlab_runner_registration_config["description"]
     gitlab_runner_tag_list                  = var.gitlab_runner_registration_config["tag_list"]
     gitlab_runner_locked_to_project         = var.gitlab_runner_registration_config["locked_to_project"]
--- a/variables.tf
+++ b/variables.tf
@@ -366,7 +366,7 @@ variable "runner_ami_owners" {
 }
 
 variable "gitlab_runner_registration_config" {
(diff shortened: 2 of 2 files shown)


midnightrun:
I believe using a regex would be even nicer but just wanted to prevent this behaviour just somehow. 

Thoughts?

npalm:
@midnightrun thanks will have a look at the PR next days

main.tf
```diff
@@ -142,7 +142,7 @@ data "template_file" "gitlab_runner" {
     runners_token                           = var.runners_token
     secure_parameter_store_runner_token_key = local.secure_parameter_store_runner_token_key
     secure_parameter_store_region           = var.aws_region
-    gitlab_runner_registration_token        = var.gitlab_runner_registration_config["registration_token"]
```
npalm: Is this really working? In this way the registration token is changed. It also seems the bug is solved in the gitlab runner: 
https://gitlab.com/gitlab-org/gitlab-runner/merge_requests/1144#9c5c461492d69bfd6a2ce40727e0b81b49a19e5b

midnightrun: Ah, sorry. I just tested if the substitution is working. If Gitlab really fixed this. This PR is obsolete and can be closed. Thanks!

npalm: I am going to test is today, just got a token with an underscore.

npalm: I have just created a runner with a token containing an `_` the runner is created without an error and working as normal. Seems the issue is resolved by GitLab.

npalm:
@midnightrun Thanks proposing a solution for #87 but the issue is resolved by GitLab since 11.7 (https://gitlab.com/gitlab-org/gitlab-runner/merge_requests/1144#9c5c461492d69bfd6a2ce40727e0b81b49a19e5b) So I will close you PR without merging.

PR #156, opened 2019-11-10
Add option to encrypt logs via KMS

npalm:
## Description
Add option to encrypt logs via KMS. Supported options
- no encryption
- module managed KMS key
- user provided key

## Migrations required
NO

## Verification
default

## Documentation


npalm:
@hendrixroa I was inspired by your PR #132 but took a bit different approach providing / supporting a few scenario's. What do you think of this approach?

hendrixroa:
@npalm Yes, I agree, the encryption kms should be optionally

PR #157, opened 2019-11-12
Allow log retention configuration

geota:
## Description
Destruction was failing for me due to the template trying to be generated AFTER the bucket had already been destroyed which caused it to fail. 

I also needed to support configuring log retention on the cloudwatch log group so we dont have unlimited retention. I set it to zero deays which maps to unlimited IIRC. I wanted to set it to null and let the AWS resource handle the defaulting but the autodocs script breaks with null as a default even though this is supported in TF. 

## Migrations required
NO

## Verification

## Documentation


npalm:
@geota thanks for the PR. will check it in the next days

PR #158, opened 2019-11-14
Make use of on-demand instances in docker-machine

skorfmann:
## Description
We are running into issues with certain instance types not being available on the spot instances market every now and then. In addition to this, we have scenarios where it's not acceptable that jobs could be killed due to AWS reclaming the spot instance, but where we still would benefit from having the autoscaling of docker-machine. 

That's why we'd like to make requesting a spot instance optional.

## Migrations required
NO


skorfmann:
> Thanks looks all good. For further PR's could you also generate the docs again? See instruction in the PR template. For now all fine.

Sweet, thanks for the quick response. Will keep the doc part in mind for potential future PRs

PR #159, opened 2019-11-15
Fix/tokens

rin-amezquita:
## Description
Currently, if gitlab returns a token starting with `-` aws ssm fails to push and furthermore the token fails. This PR fixes it

## Migrations required
NO


npalm:
@rin-amezquita thanks will check this week

PR #161, opened 2019-12-08
Allow assignment of Elastic IP

roock:
## Description
allow assignment of an elastic IP for the gitlab runner instance (mostly useful if it is run with a public IP

## Migrations required
NO. Feature is disabled by default

## Verification

## Documentation
Documentation has been updated.


npalm:
@roock thanks will check later this week

Issue #163, opened 2019-12-11
Runners should remove themselves on instance termination


pshuman-heb:
Right now, when a runner instance is terminated, it leaves behind a phantom runner registration in Gitlab that takes a few minutes to go unhealthy, and requires you to go into Gitlab to remove it. It should be easy enough to use the stored access token to fire off a `DELETE` request as [documented in the Gitlab API](https://docs.gitlab.com/ee/api/runners.html#remove-a-runner), and hook that script up to the instance termination alerts/scripts.

There's a number of ways to set up on-terminate scripts in EC2, and I'm not totally familiar with the pros and cons of each of them, but my understanding is that we should be able to just drop a script into `/etc/rc0.d/SCRIPTNAME` and have that fire off a request to remove the runner on system shutdowns? The other way that may be better for use with Spot instances would be a cron script that would check the Termination notices from the instance metadata service and remove when that pops up? That could also help to prevent jobs getting scheduled right before the runner is going to die.

I'll see about making a PR for this - but if anyone has some insight on the proper way to do this, it's perhaps a bit beyond my limited UNIX knowledge to say which way is best.

ado120:
I also agree that when destroying the gitlabrunners via terraform destroy, the runners themselves should be removed from GitLab Runners page

pshuman-heb:
@ado120 I think this will work as a catch-all for both that situation, and for instance termination inside of AWS, either natural reboots, ASG actions, or Spot Instance terminations.

PR #164, opened 2019-12-11
Remove runner on instance shutdown.

pshuman-heb:
## Description
This is primarily set up to resolve the issues I pointed out in #163. This creates a one-line script that sends a `DELETE` request to the proper Gitlab endpoint to remove this runner from Gitlab. I'm also removing the key from SSM at the same time, since the key is now unable to be used in the future. 

## Migrations required
NO, I don't think?

## Verification
I've been verifying this by pointing my usage of this module to my fork/branch and then trying doing apply/destroys. At this point, it seems to be working as expected, but I'd love to see someone else validate it on their own.

## Documentation
I've added what I help are appropriate comments - let me know if I should add more robust documentation elsewhere, but this doesn't seem to have a place to put it.

## Issues Outstanding

~~Currently, I'm running into problems around the use of SSM caching of the runner Token in the following situation:~~
    1. ~~Token is returned when the Runner is created the first time.~~
    2. ~~Token is put into SSM as a param.~~
    3. ~~I terminate the instance in the ASG from the AWS Console to simulate an instance failure/reboot.~~
    4. ~~The runner api is called with a `DELETE` for that token, removing the runner as expected.~~
    5. ~~A new instance boots, finds the token in SSM and presumably tries to use it like normal to register.~~
    6. ~~Seemingly, Gitlab rejects the token, since it's been removed/deleted, but there's not fallback/retry/reset built into the boot script, so the instance is up and running, but not registered with Gitlab.~~

~~I'm not totally sure how to get around this. We could probably set up some sort of retry logic around registering the Gitlab runner with tokens, where it would fall back and register from scratch if it's rejected with that token?~~

~~I also just don't generally understand the purpose of the SSM caching of something that Gitlab seems to treat as a per-Runner key. Why not just always get a new key for every instance and de-register them as the instance dies?~~

Issues seem to be resolved in my own testing at this point!

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ In this scenario *not* docker machine is used but docker to schedule the builds.
 
 Ensure you have Terraform installed the modules is based on Terraform 0.11, see `.terraform-version` for the used version. A handy tool to mange your Terraform version is [tfenv](https://github.com/kamatama41/tfenv).
 
-On macOS it is simple to install `tfenv` using brew.
+On macOS it is simple to install `tfenv` using `brew`.
 
 ``` sh
 brew install tfenv
@@ -71,7 +71,17 @@ tfenv install <version>
 
 ### AWS
 
-Ensure you have setup you AWS credentials. The module requires access to IAM, EC2, CloudWatch, S3 and SSM.
+Ensure you have setup your AWS credentials. The module requires access to IAM, EC2, CloudWatch, S3 and SSM.
+
+### JQ & AWS CLI
+
+In order to be able to destroy the module, you will need to run from a host with both `jq` and `aws` installed and accessible in the environment.
+
+On macOS it is simple to install them using `brew`.
+
+``` sh
+brew install jq awscli
+```
 
 ### Service linked roles
 
@@ -94,7 +104,7 @@ resource "aws_iam_service_linked_role" "autoscaling" {
 
 ### GitLab runner token configuration
 
-By default the runner is registered on initial deployment. In previous versions of this module this was a manual process. The manual process is still supported but will be removed in future releases. The runner token will be stored in the parameter store. See [example](examples/runner-pre-registered/) for more details.
+By default the runner is registered on initial deployment. In previous versions of this module this was a manual process. The manual process is still supported but will be removed in future releases. The runner token will be stored in the AWS parameter store. See [example](examples/runner-pre-registered/) for more details.
 
 To register the runner automatically set the variable `gitlab_runner_registration_config["token"]`. This token value can be found in your GitLab project, group, or global settings. For a generic runner you can find the token in the admin section. By default the runner will be locked to the target project, not run untagged. Below is an example of the configuration map.
 
@@ -110,7 +120,7 @@ gitlab_runner_registration_config = {
 }
 ```
 
-For migration to the new setup simply add the runner token to the parameter store. Once the runner is started it will lookup the required values via the parameter store. If the value is `null` a new runner will be created.
+For migration to the new setup simply add the runner token to the parameter store. Once the runner is started it will lookup the required values via the parameter store. If the value is `null` a new runner will be registered and a new token created/stored.
 
 ``` sh
 # set the following variables, look up the variables in your Terraform config.
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ resource "aws_iam_service_linked_role" "autoscaling" {
 
 ### GitLab runner token configuration
 
-By default the runner is registered on initial deployment. In previous versions of this module this was a manual process. The manual process is still supported but will be removed in future releases. The runner token will be stored in the AWS parameter store. See [example](examples/runner-pre-registered/) for more details.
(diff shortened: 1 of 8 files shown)


pshuman-heb:
I'm really lost here on what's going wrong with using Runlevel for cleanup. The script is in place, and if I run things manually with `/etc/rc.d/init.d/remove_gitlab_registration start`, it runs fine and works. When I terminate the instance, however, nothing gets logged, and the runner isn't removed, and it doesn't seem to be running the initial lockfile setup on startup either.

npalm:
@pshuman-heb Hopefully I have next week some time to have a look at the issue as well.

pshuman-heb:
So, it feels a bit hacky, but I think it's working right now! I don't like that I'm manually having to `touch` the lockfile during the userdata script, but that seems to be necessary since I'm not seeing the actual `start` script getting run as part of the initial boot up.

I also adjusted things some so that the Removal script is only run on shutdown, and not on reboot. In general, reboot seems like a case where we actually want to keep the key active and in SSM, so that we can pick it up on reboot. This is also necessary for now because there doesn't seem to have any logic in the scripting for dealing with having the runner fail to boot up because of a bad key.

I'll make a different issue for that matter. It seems like we should be able to recognize that the Runner failed to boot because of a bad key, and just request a new one.

It's also important to note that I don't think this script will reliably work when an instance is Terminated as opposed to Stopped. This seems to be ok when it comes to `terraform destroy`s, but could cause issues in other cases.

npalm:
@pshuman-heb thanks, will check the PR this weekend.

npalm:
@pshuman-heb I have tested several scenarios. The removal works fine when I terminate the instance. But by running a terraform destroy the runner remains in GitLab. I have no clue why the shutdown script in that case is not executed. Maybe it is related to the use of a auto scaling group. 

Another approach could be remove the runner with a local-exec resource. But in that case we should first obtain the token via SSM. 

Was terraform destroy working for you?

npalm:
I quickly tried to remove the runner via terraform. Could not got i working with a data_source to obtain the token but with a simple script is was quite ease:

I have created the following script
```shell
set -e
TOKEN=$(aws ssm get-parameters --name $3 --with-decryption --region $1 | jq -r ".Parameters | .[0] | .Value")
curl -sS --request DELETE "${2}/api/v4/runners" --form "token=${TOKEN}"
```

And added the following local-exec to the main.tf file
```hcl
resource "null_resource" "remove_runner" {
  depends_on = [aws_ssm_parameter.runner_registration_token]
  provisioner "local-exec" {
    when       = destroy
    on_failure = continue
    command    = "${path.module}/remove-runner.sh ${var.aws_region} ${var.runners_gitlab_url} ${local.secure_parameter_store_runner_token_key}"
  }
}

```

What do you think of this approach

pshuman-heb:
I've seen it work when using `terraform destroy` in my testing, but my guess is it's not necessarily going to be consistent, perhaps because it's removing the instance faster than the instance can run the script.

I like doing both solutions then, perhaps? Your script there should solve for the case of destroying via Terraform, and mine should hopefully solve for "natural" loss of the instance in AWS due to hardware issues or a reboot.

The only potential issue is if both the instance and this script both try to delete the runner at the same time, but I think that doesn't cause any real problems that would need to be solved. One or the other of the scripts would fail, but we could probably add some output logging to explain the situation if this "local" one fails.

npalm:
You mention the script solved the issue of natural lost. This is already handled by the user data script. In case the instance is destroyed it fetch the token from smm. So it will reuse the same, all ready registered runner. Are you aware of this a bit magic behaviour?

pshuman-heb:
I'm aware of the behavior, but I've found it can be a bit awkward in some situations. In the case that AWS terminates the instance and then it immediately comes back up through the ASG, or there's scaling rules that remove the runners during off-hours, it's fine to have the new instances come back up and get the value from SSM. There's a couple other use-cases where I think it could cause issues.

First, I'm unsure of how Gitlab handles potentially having multiple active instances all registered with the same token? Right now, if the ASG ever scales to 2, it seems like it could lead to Gitlab's scheduler getting confused since two Runners are reporting as the same registration. 

Second, in the case where you deploy an update to the Runner config that would cause it to need to re-register with a different Token, there's no neat way to do that right now without manual intervention, or totally destroying the module. For instance, if I rotate the key for runner registration for my Group/Project, the runners will keep thinking the old registration will work. This is also a problem if you have to move a set of runners to a different Group/Project - since they just get the key from SSM, they never pick up the new location. This is a similar issue to #173.

I'll admit that neither of these are major issues that demand a change, but personally, I also just feel like it's better practice to always de-register when the runner is going offline, and re-register as a new runner when they come up again. Registrations are essentially free, so I don't see much downside to having each instance make its own registration.

npalm:
Caching in SSM was introduced since it was hard to delete the runner, now we can delete the runner maybe I go to remove the whole SSM part. It does not make that much sence anymore.

pshuman-heb:
Yea, the SSM stuff works well, but now that it's clean and easy to create/destroy runners, that seems like it could be the better solution.

Issue #169, opened 2019-12-16
docker runner fail to upload artifacts


riccardomc:
Hi, thanks for this module.

I am running into an issue with newly created gitlab runners. I am unable to upload artifacts to gitlab.com and I suspect it is because the url is wrong. The issue is well described here:

https://gitlab.com/gitlab-org/gitlab-runner/issues/4083

npalm:
@riccardomc thanks, released the fix in 4.10.0

PR #170, opened 2019-12-16
Change default for gitlab_url to https://gitlab.com

riccardomc:
## Description
Docker runners artifact upload fails with error described in:
https://gitlab.com/gitlab-org/gitlab-runner/issues/4083

The default value of the `gitlab_url` variable should be `https://gitlab.com` and not `https://www.gitlab.com`.

See #169

## Migrations required
NO

## Verification
Tested the change using docker-runner using the value `https://gitlab.com` in `runners_gitlab_url`.

## Documentation
updated:
* examples/runner-docker/_docs/TF_MODULE.md
* examples/runner-docker/README.md


npalm:
@riccardomc  thanks, will check the PR this weekend.

PR #173, opened 2020-01-02
Forcing updates of Instances on Config change.

pshuman-heb:
## Description
Currently, it's possible to `terraform apply` an update to a number of aspects of the Runner, but not see them actually reflected in the running resources because of how AWS works. For example, if I need to change the Runner registration token, you can apply those changes, but the actual Runner resources won't update on their own to pick up that change.

This is due to the fact that in both AWS and Terraform, EC2 Launch Configurations are a separate resource from the EC2 Autoscaling groups. Many/most changes to how the Runners are configured are set as part of the Launch Config, and while Terraform will dutifully create a new Launch Config and update the ASG to use it, the ASG will only apply the new config to _new_ instances, and will not push the update to any currently-running instances in the Group.

There's a number of possible ways to get around this: 

 1. If this was Cloudformation, we could use a built-in lifecycle management tool it has to tell it to trigger a rolling update of all the instances in the ASG. This feature evidently isn't something Terraform can use naturally however, so the only way to use it would be to have Terraform make a Cloudformation resource that created the Launch Config and ASG.
 2. We could make a null resource and trigger it so that when a new Launch Config is made, it goes and Terminates the existing instance in the ASG using the AWS CLI, causing a new one to be created using the new config.
 3. We could link the ASG and Launch Config by name in Terraform so that Terraform would be forced to make a brand new ASG every time we make a new Launch Config, and the new ASG would use the new configs.

Option #1 would require some fairly major adjustment to the project, and would involve doing some templates-inside-templates patterns that I don't really like. Option #2 seems viable, but making sure it's timed correctly seems complicated, and in general I find using custom scripts to do something like this a bit awkward. Option #3 is a bit of a hack of how Terraform works, but in this specific case, happens to directly work with what we need it to do, and has the advantage that it will show up in `plan` output.

Normally, you wouldn't want to do Option 3 because you'd lose any scaling state for the ASG, which could be disastrous if you're relying on Autoscaling to handle load. Since, in this project, the ASG is generally going to only be at 0 or 1 though, we should be mostly safe.

## Migrations required
NO - this should migrate itself cleanly from old versions. The only change to the steady state is in the new names for the resources.

## Verification
Working on verification now - in general, this should be fairly straightforward though, since it's only an adjustment to how Terraform resolves the dependency graph.

## Documentation

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -256,9 +256,8 @@ data "aws_ami" "docker-machine" {
 }
 
 resource "aws_autoscaling_group" "gitlab_runner_instance" {
-  name                = "${var.environment}-as-group"
-  vpc_zone_identifier = var.subnet_ids_gitlab_runner
-
+  name                      = "${aws_launch_configuration.gitlab_runner_instance.name}-asg"
+  vpc_zone_identifier       = var.subnet_ids_gitlab_runner
   min_size                  = "1"
   max_size                  = "1"
   desired_capacity          = "1"
@@ -320,6 +319,7 @@ locals {
 }
 
 resource "aws_launch_configuration" "gitlab_runner_instance" {
+  name_prefix          = var.runners_name
   security_groups      = [aws_security_group.runner.id]
   key_name             = local.key_pair_name
   image_id             = data.aws_ami.runner.id
--- a/main.tf
+++ b/main.tf
@@ -256,7 +256,7 @@ data "aws_ami" "docker-machine" {
 }
 
 resource "aws_autoscaling_group" "gitlab_runner_instance" {
-  name                      = var.enable_forced_updates ? "${var.environment}-as-group" :"${aws_launch_configuration.gitlab_runner_instance.name}-asg"
+  name                      = var.enable_forced_updates ? "${var.environment}-as-group" : "${aws_launch_configuration.gitlab_runner_instance.name}-asg"
   vpc_zone_identifier       = var.subnet_ids_gitlab_runner
   min_size                  = "1"
   max_size                  = "1"
--- a/main.tf
+++ b/main.tf
@@ -256,7 +256,7 @@ data "aws_ami" "docker-machine" {
 }
 
 resource "aws_autoscaling_group" "gitlab_runner_instance" {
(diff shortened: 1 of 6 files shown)


npalm:
@pshuman-heb thanks for creating the PR. I agree in general to your observations. But maybe we can think about scenario 2, so we add a small cloudfront script managed by terraform that take care of rolling forward? What do you think?

The change you proposed have impact for users of the module, since the behaviour is changing. Do you have any idea how we can give current users the option not use this roll over?

pshuman-heb:
With the current way that this is being done, we could likely add a variable that would enable this behavior, and then set the name of the ASG to either static or dynamic depending on the value of that var.

My only worry with Option 2 would be when to do the trigger for it. Essentially, we'd want to make a call to [this command](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/terminate-instance-in-auto-scaling-group.html) with the proper Instance ID/IDs. If we send that command too soon though, the new Launch Config won't be available/attached yet though. I suppose you could make it trigger on creation of the new Launch Config, but then you'd need to have a case to handle initial deployment where the ASG instances may not exist yet, as well as the case where it might be scaled down to 0 at the time of the deployment already.

I'm up for doing that if you think it's an improvement - to me it just felt unnecessarily complicated and more prone to bugs than just relying on AWS to do the right thing.

main.tf
```diff
@@ -256,9 +256,8 @@ data "aws_ami" "docker-machine" {
 }
 
 resource "aws_autoscaling_group" "gitlab_runner_instance" {
-  name                = "${var.environment}-as-group"
-  vpc_zone_identifier = var.subnet_ids_gitlab_runner
-
+  name                      = var.enable_forced_updates ? "${var.environment}-as-group" :"${aws_launch_configuration.gitlab_runner_instance.name}-asg"
```
pshuman-heb: I've left the old name as-is for backwards compatibility, but it does mean that the pattern changes depending on if you're using the forced updates. I'd use the full `-as-group`, but I'm worried that the length of the name will become a problem.

Issue #175, opened 2020-01-13
Tags for EC2 instances


Glen-Moonpig:
We have a requirement to apply certain tags to EC2 instances only. The module currently only has one 'tags' input variable, which applies to all resources.
We are considering adding a new input variable for tags that will be applied to the EC2 instances only (as well as the tags being applied currently).

```
variable "runner_tags" {
  description = "Map of tags that will be added to created EC2 resources."
  type        = map(string)
  default     = {}
}
```

Is this something that you might merge in? Any other suggestions?

npalm:
@Glen-Moonpig yes a PR is welkome! Beware that there are two types of instances. Also you can have a look at the variable `overrides` this variable overwrite defaults for names of the instance

PR #179, opened 2020-01-30
Added agent- and runner-only tags

Glen-Moonpig:
### Description

Solves #175. We need to apply tags only to the EC2 resources for an application that manages resources by tag.

### Migrations required

NO

### Documentation

Readme updated

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -209,10 +209,10 @@ data "template_file" "runners" {
     docker_machine_options      = length(var.docker_machine_options) == 0 ? "" : local.docker_machine_options_string
     runners_name                = var.runners_name
     runners_tags = var.overrides["name_docker_machine_runners"] == "" ? format(
-      "%s,%s,Name,%s-docker-machine",
+      "Name,%s-docker-machine,%s,%s",
+      var.environment,
       local.tags_string,
       local.runner_tags_string,
-      var.environment,
       ) : format(
       "%s,%s,Name,%s",
       local.tags_string,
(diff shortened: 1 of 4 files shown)


npalm:
Thanks, will check in the next days

Glen-Moonpig:
Thanks Niek, I decided to add inputs rather than use the overrides because I think that was focused on only the runner machine setup and also I need to add extra tags rather than override default settings, but happy to make any changes you suggest.

main.tf
```diff
@@ -209,12 +209,14 @@ data "template_file" "runners" {
...
+      "%s,%s,Name,%s-docker-machine",
       local.tags_string,
+      local.runner_tags_string,
       var.environment,
       ) : format(
-      "%s,Name,%s",
+      "%s,%s,Name,%s",
       local.tags_string,
+      local.runner_tags_string,
       var.overrides["name_docker_machine_runners"],
```
npalm: I think we have to change the order of the tags added to the string. Such that in the default case  (case name_docker_machine_runners is empty) as follow:

First add the default Name to string, next the tags_string and finally the runner_tags_astring. In this case the runner_tags string overwrite all the previous and so on. 

The code will be than the follow:

```      "Name,%s-docker-machine,%s,%s",
      var.environment,
      local.tags_string,
      local.runner_tags_string,
      ) : format(
      "%s,%s,Name,%s",
      local.tags_string,
      local.runner_tags_string,
      var.overrides["name_docker_machine_runners"],```

Glen-Moonpig: I have changed the order, but not sure if I understand, please check the change.

npalm: thanks, will check as soon as possible.

npalm:
@Glen-Moonpig thanks!

PR #181, opened 2020-02-13
Add inputs for EBS-optimized

chrizkim:
## Description
Without specifying an EBS-optimized, AWS defaults to `false`.  This adds an EBS-optimized configuration option to both the runner instance and runners. Defaults them both to `true` (both the default GitLab runner instance type, `t3.micro`, and the default docker machine runners instance type, `m5a.large`, use EBS-optimized by default anyways, even though the instances are currently set to disabled).

## Migrations required
NO

## Verification
default example

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ terraform destroy
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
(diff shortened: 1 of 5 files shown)


npalm:
@chrizkim thanks, will check in the next days

npalm:
Thanks!

PR #182, opened 2020-02-14
Fix error create bucket false

katiatalh:
## Description
Terraform failed to create aws_iam_role_policy_attachment resource named docker_machine_cache_instance when cache_bucket["create"] is set to false. 

## Migrations required
No migrations required

## Verification
`terraform apply` and` terraform destroy`


npalm:
@chrizkim thanks, will check in the next days

npalm:
Thanks!

PR #190, opened 2020-03-02
Complete docker example with ssh keys usage

katiatalh:
## Description
In runner-docker example, there is a ssh keys generation but these keys are not used in runner module. We can't connect to ec2 instance.

## Migrations required
NO

## Verification
runner-docker example


npalm:
@katiatalh thanks will check in the next days

PR #192, opened 2020-03-09
Drop support to inject the ssh public key string, only key pair name …

npalm:
## Description
Due to security and complexity the module will not longer manage an AWS EC2 key. You can either choose to inject your own key via a variable or use SSM in the case you need to access an EC2 instance. 

## Migrations required
YES, in case the ec2 key pair was managed by the module.

## Verification
default and docker example

## Documentation
updated


npalm:
@fliphess Just wondering if you have any opinion about this change since you are quite active the last weeks. I drop the support to let the module handle the EC2 key.

fliphess:
@npalm Removing the option to add SSH key's in this module seems like a good idea to reduce complexity to me :) 

Every company prefers to deal with access and SSH keys in their own way and removing this feature does not limit that as there are still plenty of options to  achieve the same results I would say :)

There is the option for user_data for manually provisioning SSH keys or for setting up orchestrators like puppet, chef or ansible to manage it... 

Additionally, there is the option to use AWS Service Manager and the option to use the AWS metadata store to provision your SSH key.... 

I don't think removing this feature is a showstopper for users of this module and I'm pretty sure that for those who need some help to work around the removal some documentation or user_data snippets can solve all their problems.

npalm:
@fliphess thanks for the quick feedback.

PR #195, opened 2020-03-13
Add permissions boundary

mhulscher:
## Description
Add a variable to allow setting a permissions boundary policy on newly created IAM roles.

## Migrations required
No.

## Verification
Running `terraform apply` both with `permissions_boundary` set and without.

Done.


npalm:
Cool! will check in the next days!

mhulscher:
> Cool! will check in the next days!

Appreciate it!

npalm:
@mhulscher Thanks for you contribution, would you like also to share a compete example of suing the permission boundaries?

PR #205, opened 2020-03-28
Allow traffic from security group

fliphess:
**Still work in progress (havent tested this change yet, to be continued on monday)**

## Description
(I've created this a while earlier and promised to commit back, and as I remembered and need it myself, a yet not fully tested PR)

I've added the option to allow traffic from a list of security group id's rather than a list of CIDR ranges.
This is useful when your gitlab-runner instances are in a private subnet hidden behind a bastion host and you don't need exposure to the internet of your gitlab-runner and or docker-machine instances.

Additionally I've added a sanity check to prevent the cidr ranges set to allow from all while your security group id's should configure otherwise and moved the security groups to their own file as the rulebase is getting quite large this way :)

## Migrations required
NO

## Verification
Will be running this branch in my own cluster later this week.

## Documentation
Done :)

## Example
Added.

(Vriendelijke groet!)


npalm:
Nice, looking forward.

fliphess:
Hey Niek! (@npalm)

I think I'm there with my pull request. I've tested the live code and this works, but I had to remove the default "0.0.0.0/0" range from the cidr range to safely combine both methods without accidentally adding "0.0.0.0/0" when you only want to use security group IDs, I hope thats okay?

I've tried to add some validation, but it didn't really work out so if you have a better idea, let me know :)

The module is currently running live in my prod AWS account, but my lab account is such a mess that I'm not very able to test the example I added, would you perhaps be able to do this? (as you are probably testing this before merging anyway)

Let me know what you think :) I'm open to all suggestions, there's a lot of changes (sorry for that), so take your time in checking / reviewing / deciding.

fliphess:
..... Or maybe I should split it up: Offer the code first in a PR, then the example and docs in another, as this is hard to read and there's still some things to do on the example...

fliphess:
#207

PR #207, opened 2020-03-31
Allow traffic from a list of security group IDs

fliphess:
The same PR but without the examples, which need some more work on which I'm working :)

## Description
I've added the option to allow traffic from a list of security group id's rather than a list of CIDR ranges.
This is useful when your gitlab-runner instances are in a private subnet hidden behind a bastion host and you don't need exposure to the internet of your gitlab-runner and or docker-machine instances.

Additionally I've added a sanity check to prevent the cidr ranges set to allow from all while your security group id's should configure otherwise and moved the security groups to their own file as the rule base is getting quite large this way :)

## Migrations required
NO

## Breaking changes
The default cidr group of "0.0.0.0/0" is removed and the default is changed to an empty list, not configuring either `var.gitlab_runner_security_group_ids` or `var.gitlab_runner_ssh_cidr_blocks` will create a functional gitlab-runner, but one that is not accessible from ip's or security group ID's (so only for SSM)

## Verification
Running in production :) 

## Documentation
Will be added in the next PR, as part of a more readable #205 

## Example
Will be added in the next PR, as part of a more readable #205 

(Vriendelijke groet!)


npalm:
@fliphess Thanks looks great, will check the PR soon

fliphess:
Cool! :) Let me know what you think, I'm open to all suggestions and improvements :) 👍

fliphess:
I've added the suggestions to the default example

Issue #211, opened 2020-04-09
Shared cache module has to be applied manually with target before applying the runner


DavidGamba:
I ran into this error:

```
Error: Invalid count argument
                                                                                               
  on .terraform/modules/deployments_runner/main.tf line 327, in resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance":
 327:   count      = var.cache_bucket["create"] || var.cache_bucket["policy"] != "" ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

As the error indicates, I had to do:

`terraform apply -target module.cache`

That configuration section looks like this in my case:

```
  cache_bucket = {
    create = false
    policy = module.cache.policy_arn
    bucket = module.cache.bucket
  }
```

so the policy is not known in advanced.

Why does the count also depend on the policy variable?

npalm:
Thanks for creating the issue. The policy is used to attach to the runners so the runner can access the cache. In case no cache is used no policy is attache, so therefore the count.

In case you don't want to use a cache you can set the cache_bucket variable as follow.
```
  cache_bucket = {
    create = false
    policy = ""
    bucket = ""
  }
```

I agree it would be more elegant in case the property policy and bucket can be left out.

npalm:
I have made a small change, see the open PR. With the version on that branch you should be able to disable the cache by setting 
```
  cache_bucket = {
    create = "false"
  }
```
Be aware the quotes for false, they are required. This is a left over of the terrafrom 0.11 erra with less typing.. So tuples in the map have mixed types.

DavidGamba:
The fix you provided worked for me. Thanks @npalm

npalm:
Cool, thx for the feedback. I will do a quick check with my own config and merge them back asap

ilansh:
I'm not sure the issue has been solved.  
The shared cache still requires an extra step of applying with -target, before applying the runner.  
Or am I missing something?

npalm:
As far I am aware it should be fixed, can you share an example. Or even better try to reproduce on one of the module examples. The public runners are sharing a cache.

PR #212, opened 2020-04-13
Fix disabled cache

npalm:
## Description
Make it easier to disable the cache by avoiding all itmes in the map needs to be set. Cache can be disabled as follow:

```
  cache_bucket = {
    create = "false"
  }
```
## Migrations required
NO

## Verification
Default sample


DavidGamba:
Tested this branch and I didn't run into the problem I had with #211 
Thanks!

npalm:
Note for myself, update docs

npalm:
Close #211

Issue #215, opened 2020-04-14
Support custom docker machine distribution


npalm:
Docker machine is in maintenance mode, therefor GitLab is maintaining an own fork. 

Add the option to the module to set a custom docker machine download url

PR #216, opened 2020-04-14
Feature/support custom docker machine dist

npalm:
## Description
Support custom docker machine download url

## Migrations required
NO

## Verification
default example


npalm:
Close #215

Issue #235, opened 2020-07-06
Remove var.aws_zone in the next major release


kayma-hl:
The aws_zone parameter in variables.tf is already deprecated and can be removed. I have already adjusted the documention (see #232).

npalm:
Solved by #236

PR #236, opened 2020-07-06
Deprecate the var.aws_zone and read it based on the given runner subnet

### LINKED ISSUE #232: Does aws_zone has to match the subnet availability zone?
Hi,

by accident I set the aws_zone to "b" but used a subnet_id_runners from availability zone eu-central-1a. As a result no runner was created. So if both variables are related I propose to remove the aws_zone variable and read the availability zone of the subnet via

`
data aws_subnet runners {
  id = var.subnet_id_runners
}
`

Best regards
Matthias Kay
Hapag-Lloyd AG

kayma-hl:
The var.aws_zone parameter for the runner is not needed. We get the runner subnet which implies the availablity_zone.

The variable is now deprecated. It can be removed in the next major release (see #235). As module documentation is created automatically, no changes required. The examples are still valid and didn't mention the aws_zone variable.

Verification was done in the Hapag-Lloyd environment which uses this module in version 4.18.0 (docker+machine and spot instances).

Closes #232


npalm:
@kayma-hl  thanks will check the PR later this week.

PR #238, opened 2020-07-06
Merge the tags for the runner agent to remove duplicate tags

### LINKED ISSUE #233: Merge tags for agent runner EC2 instance to remove duplicates
Hi,

I set the environment variable to "abc" and use tags for the runners including the name tag.

```
module "gitlab_terraform_runner" {
  ...
  environment                = "abc"
  agent_tags                 = merge(var.tags, map("Name", "Gitlab Runner Terraform Agent"))
  runner_tags                = merge(var.tags, map("Name", "Gitlab Runner Terraform"))
}
```

The problem is, that the environment variable is always added as a name tag to the agent. In the end it contains 2 name tags.

Proposal: Add the anvironment as name tag only if the agent has no name tag.

Best regards
Matthias Kay
Hapag-Lloyd AG

kayma-hl:
The tag list for the runner agent was built with concat() which produced duplicate tag keys.

I added a short documentation to show how to set the name of the agent runner.

Verification was done in the Hapag-Lloyd environment which uses this module in version 4.18.0 (docker+machine and spot instances).

Closes #233


npalm:
@kayma-hl  thanks will check the PR later this week.

Issue #246, opened 2020-08-17
Support for multiple docker machines within a single runner instance


krzysztof-miemiec:
I'm currently doing small research in order to migrate my current GitLab runner setup (done with CloudFormation, modified and updated by hand via SSH) to Terraform. I'm currently using a setup with configuration like this:

```toml
concurrent = 20
check_interval = 0

[session_server]
  session_timeout = 3600

[[runners]]
  name = "team-a-runner"
  url = "https://gitlab.(redacted)/"
  token = "(team-a token redacted)"
  executor = "docker+machine"
  [runners.docker]
    tls_verify = false
    image = "alpine:latest"
    privileged = true
    disable_entrypoint_overwrite = false
    oom_kill_disable = false
    disable_cache = true
    volumes = ["/certs/client", "/cache"]
    shm_size = 0
  [runners.cache]
    Type = "s3"
    Shared = true
    [runners.cache.s3]
      ServerAddress = "s3.amazonaws.com"
      BucketName = "(redacted)"
      BucketLocation = "eu-central-1"
  [runners.machine]
    IdleCount = 0
    IdleTime = 1200
    MaxBuilds = 100
    MachineDriver = "amazonec2"
    MachineName = "gitlab-docker-machine-%s"
    MachineOptions = ["engine-storage-driver=overlay2", "amazonec2-instance-type=m5.large", "amazonec2-region=eu-central-1", "amazonec2-vpc-id=(redacted)", "amazonec2-security-group=(redacted)", "amazonec2-use-private-address=true", "amazonec2-iam-instance-profile=(redacted)"]
    OffPeakPeriods = ["* * 0-9,17-23 * * mon-fri *", "* * * * * sat,sun *"]
    OffPeakTimezone = "Etc/UTC"
    OffPeakIdleCount = 0
    OffPeakIdleTime = 300

[[runners]]
  name = "team-b-runner"
  url = "https://gitlab.(redacted)/"
  token = "(team-b token redacted)"
  executor = "docker+machine"
  (same as above)

[[runners]]
  name = "team-c-runner"
  url = "https://gitlab.(redacted)/"
  token = "(team-c token redacted)"
  executor = "docker+machine"
  (same as above)

[[runners]]
  name = "team-d-runner"
  url = "https://gitlab.(redacted)/"
  token = "(team-d token redacted)"
  executor = "docker+machine"
  (same as above)

[[runners]]
  name = "team-e-runner"
[...]

Issue #77, opened 2019-06-03
Allow multiple runner instance types


declension:
### Background
When adding only one instance type for a spot request (in the wizard), it becomes clear this is affecting the availability (which brings down our gitlab runners sometimes with `capacity-not-available`)
![image](https://user-images.githubusercontent.com/3322808/58792708-9d56d380-85ec-11e9-8d13-dc09b54801c7.png)
See related #76 

### Proposal
 * Deprecate `instance_type` in favour of `instance_types`, a list (e.g. `instance_types = ['m4.large', 'm5.large']`, and, err, somehow make this work downstream...

### Notes
This probably isn't possible [until Fleets / multiple instance types are supported in `docker-machine`](https://github.com/docker/machine/issues/4588), unless perhaps it could be done manually, or within Gitlab Runner itself somehow... :thinking:

npalm:
Thanks, already cracking my thoughts on this topic. Please let me know if you have any other idea. 

Options could be running multiple runners with different configs, but problem will be that unhealty runners will have stucked build. So maybe an agent as watcher or lambda to take the out of the healthy runner agent pools

npalm:
@declension the runner supports now multiple instance. You can create multiple runners and share the cache.  See the example in `examples/runner-public`.

declension:
Great - we'll try this soon. Thanks @npalm

npalm:
@declension shall I close the issue?

DavidGamba:
@npalm does this work when you have exactly the same tags for the multiple instances?
Or will GitLab be stuck with the instance it first picked until it times out?

npalm:
> @npalm does this work when you have exactly the same tags for the multiple instances?
> Or will GitLab be stuck with the instance it first picked until it times out?

Not sure, guess the build will be picked by the runner that is polling first.

Issue #76, opened 2019-06-03
Allow spot requests in multiple availability zones


declension:
### Background
At $WORK, we have a pre-defined VPC, and one subnet within that (but want more to avoid any further capacity issues) for Gitlab & runners. 

Having multiple AZs is entirely possible, as the spot request wizard has checkboxes for each:

![image](https://user-images.githubusercontent.com/3322808/58792375-f6723780-85eb-11e9-8994-deee96fd624a.png)
 

### Proposal
 * Multiple availability zones (via `aws_zone`) can be specified as parameters, *or*
 * Multiple subnets (via a new `subnet_ids_runners` perhaps?) in various AZs, with the AZ inferred automatically, so deprecating `aws_zone` perhaps?

### Notes
I don't know if this necessarily makes sense, but am definitely keen to find ways to improve our `capacity-not-available` problems (not pricing), and this seems like one major cause

declension:
Hmm, https://github.com/docker/machine/pull/4701 looks quite relevant / concerning...

npalm:
@declension you can spin now a runner in multiple zones, would this help see #77  and PR #96

declension:
Thanks @npalm, we'll try this soon. Closing for now

PR #249, opened 2020-08-20
feat: add multiple executor definitions to agent

krzysztof-miemiec:
## Description
It's my first attempt to answer [my own question](https://github.com/npalm/terraform-aws-gitlab-runner/issues/246) - if it's ok to create many runners within a single EC2 "manager" instance, then let's do this. In theory, this allows to manage something like a "spot fleet" of docker-machine instances without necessarily spinning up multiple runner instances. Instead of defining aws-gitlab-runner multiple times, we only have to pass:

```tf
  module "gitlab_runner" {
    # ...
    runners = [
      { 
        instance_type = "c4.large"
        subnet_id = "subnet-a"
      },
      {
        instance_type = "c5.large"
        subnet_id = "subnet-a"
      },
      {
        instance_type = "c5d.large"
        subnet_id = "subnet-a"
      },
      {
        instance_type = "c5a.large"
        subnet_id = "subnet-a"
      },
      { 
        instance_type = "c4.large"
        subnet_id = "subnet-b"
      },
      {
        instance_type = "c5.large"
        subnet_id = "subnet-b"
      },
      {
        instance_type = "c5d.large"
        subnet_id = "subnet-b"
      },
      {
        instance_type = "c5a.large"
        subnet_id = "subnet-b"
      },
      { 
        instance_type = "c4.large"
        subnet_id = "subnet-c"
      },
      {
        instance_type = "c5.large"
        subnet_id = "subnet-c"
      },
      {
        instance_type = "c5d.large"
        subnet_id = "subnet-c"
      },
      {
        instance_type = "c5a.large"
        subnet_id = "subnet-c"
      },
    ]
  }
```

If I get it right, we may pay only $5-10/mo for single gitlab runner instance instead of $60-120/mo for 12 instances and it may improve existing, multi-ec2/multi-module solution for:
- https://github.com/npalm/terraform-aws-gitlab-runner/issues/76
- https://github.com/npalm/terraform-aws-gitlab-runner/issues/77

## Migrations required
`NO` - It's supposed to work in backwards-compatible manner (no outputs or resources changed, yet spacing may be different in resulting config file).

## Verification
`TODO`: I'll have to test the whole module first and get back to you with results.

## Documentation
`TODO`: It's a new feature, I'd like to hear authors/maintainers opinion before I proceed further 🙇 

Please ensure you update the README in `_docs/README.md`. The README.md in the root can be updated by running the script `ci/bin/autodocs.sh`, requires `terraform-docs` version 0.8+


modules/config/main.tf
```diff
@@ -0,0 +1,69 @@
...
+    },
+    {
+      "Environment" = format("%s", var.environment)
+    },
+    var.tags,
+  )
+}
+
+resource "aws_s3_bucket" "gitlab_config" {
+  count = var.create_bucket ? 1 : 0
```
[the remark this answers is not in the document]
krzysztof-miemiec: @npalm Thanks a lot for giving the context - now I know why some bits of code look unfamiliar to me (I started with TF 0.12) 😉

@kayman-mk I've got bad news for you: for I now removed backwards compatibility, as I'd like to focus on making multi-runner work and be stable first, however I'll definitely make sure to make it backwards-compatible if possible.

krzysztof-miemiec:
I marked PR as draft, I somehow missed that function earlier... 🙈

Now, it looks like saving config to S3 bucket makes a lot of sense. Today I made a step towards shortening downtime during runner changes. First, I noticed that when runner configuration gets changed, it does not trigger EC2 reboot, so I made use of `aws_ssm_document` & created 2 shell scripts which update the config file on existing EC2 and restart `gitlab-runner` on existing EC2. Everything is controlled via Terraform (which waits until provisioner does it's job) and runner changes are up after less than 10 seconds. 🎉
I also misconfigured the runner a few times (subnet/az mismatch, probably too small NVMe volume), so that's a good opportunity to test if new config works correctly.

krzysztof-miemiec:
After a few days of testing, I think this PR is ready for initial review. I was able to apply configuration changes around 50 times and it worked every time. I also tested the case of fresh instance being started (no jq/gitlab-runner installed) and handled this case by skipping config modifications, as config is already processed after `gitlab-runner` is installed.
The only downside I see now is that configuration update requires Terraform to run local script. Potential solution: watch for [S3 events](https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html) indicating config changes, & trigger proper SSM action using lambda function.

I'm not sure how to deal with breaking changes, I see two options:
- make a breaking change (migration is as simple as `terragrunt apply`, config bucket, IAM permissions and SSM document are applied automatically, default options are retained)
- support "old" way of dealing with config in parallel with new solution (as a result it may be really hard to maintain, I'm not sure what are the benefits of that 🤔)

## Remaining tasks
- Docs/example
- I'm still trying to configure reverse proxy that would cache Docker images (and potentially other frequently accessed things, i.e. NPM packages) as per Gitlab docs: [distributed container registry mirroring](https://docs.gitlab.com/runner/configuration/autoscale.html#distributed-container-registry-mirroring), as without it, you can't really use Gitlab runners in private VPC network (that's my current setup - I need to keep all runners under single IP/network). I can imagine seeing VPC NAT Gateway data transfer costs being higher than spot instance costs, given how many Docker images & NPM packages we pull every day. I'll probably create a separate issue to cover that case. Benefits of having solution for that problem: decrease in costs 💰, ecology 🍃, faster local data transfer 🚀

kayman-mk:
@krzysztof-miemiec We do not use this docker image cache from Gitlab. We have a small EC2 instance running which serves a Sonatype Nexus for all Maven, Docker, ... artifacts. We need it for the artifacts our CI pipeline produces. So we use it as cache for all other artifacts too.

krzysztof-miemiec:
@kayman-mk this sounds great, thanks a lot! I'll have a look at OSS version, as it looks like it may replace our existing ChartMuseum *and* cache multiple Docker repos *and* NPM repos. 🎉 As with every new tool I learn - I'm hyped already! 😅

DavidGamba:
> The only downside I see now is that configuration update requires Terraform to run local script

@krzysztof-miemiec if this solution requires a local script call it might break setups like mine where we use a multi account strategy and the machine were I run the Terraform changes from is not in the same account as the runners.
That means that I need to assume a role in order to run local scripts so the change it could make the module unusable in a multi account scenario.

Keep in mind that I have not reviewed the PR and have very little context.

npalm:
With the new docker pull policies they announced last week a proxy / cache for docker images becomes really a need to. avoid you hit the pull limits.

npalm:
Just tried to run the PR in current state, having some findings see below.

1. Got the following error while running the default example:
```
Error: Error running command '../../bin/reload-runner-config.sh default-auto123-asg ReloadGitlabConfiguration': exit status 254. Output:
An error occurred (ValidationException) when calling the GetCommandInvocation operation: 2 validation errors detected: Value 'null' at 'instanceId' failed to satisfy constraint: Member must have length greater than or equal to 10.; Value 'null' at 'instanceId' failed to satisfy constraint: Member must satisfy regular expression pattern: (^i-(\w{8}|\w{17})$)|(^mi-\w{17}$).
```
Once running the apply once again, all resources seems created succesfully.

2. Next tried a runner block with several instances. but seems only my m5.large was used. Not having the time to check further.

3. Checked the assume role, which I also use in some cases. This cause several issues with the script. 
- aws configuration seems not be picked up, als the script depends on the user configured region, would suggest to pass the terraform region in to the script and pass it to regon or AWS_REGION.
- the profile is not picked-up. Here there are a few options. Passing the profile name, which requies a local profie needs to exists. This is not required by terraform. A maybe better alternative would be to run a assume_role and pass the role set in the provider (if set).

krzysztof-miemiec:
@npalm thanks for taking time to check it! After a bit of tinkering, I came up with a solution which does not require local script. I requires a lot of AWS resources though, so I extracted the whole "technical part" of config to a separate module. As a result:
- config is created on S3 bucket
- CloudTrail is notified of that change
- CloudWatch watches for events with a specific filter for S3 object
- CloudWatch triggers a "Run Command" on SSM Document
- Command is ran automatically on "manager" EC2 instance after configuration change in S3 bucket

Customizable things:
- Config S3 bucket
- Config file S3 key
- CloudTrail S3 bucket
- CloudTrail S3 prefix

krzysztof-miemiec:
@npalm 

I wasn't able to reliably start instances event with multiple runners defined, as "fake spot fleet" doesn't do it's job. GL runner doesn't retry provisioning docker-machine if single-instance-type EC2 pool is depleted. Once in a while, build failed because of that. To vastly improve that, I decided to look at alternative docker-machine drivers. I forked [docker-machine-driver-terraform](https://github.com/tintoy/docker-machine-driver-terraform), published a [slightly upgraded version of it](https://github.com/krzysztof-miemiec/docker-machine-driver-terraform/releases/tag/v0.4.2) (it now supports Terraform 0.12+) and tested it with my PR. You can see my runner configuration in `examples/runner-custom-terraform-spot-fleet`. 

To sum up, if this PR gets merged, you'll be able to create a single runner agent which can spin up different runners (spot/non-spot) and even pass enough custom scripting to provision `docker-machine`s with Terraform. This allows to create instances with spot fleets, which is a feature that couldn't land in `docker-machine`. 🤩

This solution also fixes point 2 of your remarks above. When I was testing previous version of my PR, it looked like GL runner randomly selected one of configurations, as I saw different instance types in my AWS Console. What's bad here is what I mentioned above - GL does not retry or re-check if configuration is correct and boots successfully (the fact that GL does not choose runners in any "smart" way is actually mentioned somewhere in GL docs 🤔). Job would fail if spot instance pool is depleted.

Terraform driver seems to make the whole thing nearly bulletproof (as long as your TF config is correct).
 This PR obviously became XXL in size. I hope you'll help me divide it if necessary.

npalm:
Looks good, will check later this week. Had a week holiday last week :)

npalm:
Sorry took a while, just ran a quick check. Agent is started, docker machines are created but not connected to the agent. So jobs are not started. In the example I missed the providers so used them from the default sample. I also had an issue with the VPC. This line was not correct `azs             = data.aws_availability_zones.available.names` brackers should be removed since   the vpc module returns a list. Also the ubuntu ami could not be selected since there was not a unique selection. Added `most_recent = true` to force the selection of one AMI. Not sure if this is correct.

So for now I don't have a clue why my job is not started on the docker machine. Seems the docker machine is not connected, on the console in the ec2 agent instance I found out that the docker-machien was timed out

```
NAME                                         ACTIVE   DRIVER      STATE     URL   SWARM   DOCKER   ERRORS
runner-8zfd4nrd-runner-1602623114-c3d5d0a4            terraform   Timeout

```

@krzysztof-miemiec any clue what I am missing? Could it be that something is wrong in the security groups?

krzysztof-miemiec:
@npalm quick (organizational) update from my side - I had quite busy time recently and didn't have time to go through an example I provided and make sure it runs correctly.

### Some general thoughts (I probably wrote that already 🙈):
👉 Spot instances don't work well when they're defined as `runners` array, but this feature can be used to setup various runner configurations with different tags (i.e. we can define it as `[ { huge spot instance pool done with terraform inside }, { small, but  stable instance pool for deployments }, { an instance pool with GPUs }]`)
👉 `docker-machine-terraform-driver`, after a few patches, does it's job quite well in spinning up multiple - I included a sample Terraform config already, however I have to check it one more time to ensure it matches what I used internally in my company. I tested that with default (just one element) `runners` definition.

### ... And some answers:
> Question: Is the cloudtrail config via SSM still used, seems not. Or do I miss something?

I left the whole S3-to-CloudTrail-to-CloudWatch-to-SSM-to-EC2 config untouched (besides fixes), but let me check that one more time. CloudTrail is indirectly used by CloudWatch events (we have to set up CloudTrail to track S3 bucket actions as CW events).

>Question: can you share config how to spin up multiple agents? Should this work? I was not able to spin up multiple agents

I had some issues with private-only IPs - quite big data transfer bill when EC2s download stuff(~100s of GBs/day in various artifacts/Docker containers) through VPC NAT Gateway, so EC2s in my fleet usually had public IPs (with no incoming traffic allowed in Security Groups). However private IPs are the right way for `docker-machine` communication 🤔 Are you using Terraform driver or a regular one?

### TODOs for me:
- [ ] Resolve conflicts
- [ ] Make sure that sample config corresponds to config I tested internally
- [ ] Ensure that there are no leftovers in new `config` submodule
- [ ] Apply fixes to AZs and AMI selection in examples
- [ ] Check what's wrong with private IPs & why there are timeouts

Fuuzetsu:
@krzysztof-miemiec I'd just like to check the status: do you plan to work on updating this PR and getting it merged in? Have you decided to not pursue this?

krzysztof-miemiec:
@Fuuzetsu I definitely would like to get this merged, however I had very little spare time to do this recently. I'll probably come with new commit(s) around the end of December. In the meantime, I can confirm that my fork works well when combined with adjusted terraform driver, creating spot instances using various machine types and AZs (as in example).

template/runner-config-runners.tpl
```diff
@@ -0,0 +1,63 @@
...
+      "amazonec2-tags=${tags}",
+      "amazonec2-use-ebs-optimized-instance=${ebs_optimized}",
+      "amazonec2-monitoring=${monitoring}",
+      "amazonec2-iam-instance-profile=%{ if iam_instance_profile_name != "" }${iam_instance_profile_name}%{ else }${instance_profile}%{ endif ~}",
+      "amazonec2-root-size=${root_size}",
+      "amazonec2-ami=${ami}"%{ if docker_machine_options != "" },%{ endif ~}
+%{ endif }
+${docker_machine_options}
+    ]
+    OffPeakTimezone = "${off_peak_timezone}"
```
Fuuzetsu, line 60: These OffPeak parameters are deprecated in favour of `runners.machine.autoscaling`: https://docs.gitlab.com/runner/configuration/autoscale.html#off-peak-time-mode-configuration-deprecated

Fuuzetsu:
I've been trying this out today. It seems quite nice so far though I've only just started. I should say that I never used `terraform-aws-gitlab-runner` before and am starting from this PR.

There is a feature request that I would like to add to this PR as it seems to fit the general agenda: allow some of the things from `gitlab_runner_registration_config` to be part of the `runners` block. This can mean that each of the registered runners will have its own token.

My main motivation for this is to be able to assign different tags per runner. A simple use-case is creating runners with instance types like `t3.micro` for small jobs but `m5.large` for something larger. We can create and register such runners but there's no way to pick in `.gitlab-ci.yml`. You are allowed to specify `tags` key to filter but we only have one GitLab runner with multiple executors. Instead it'd be cool if each executor was registered by itself.

Fuuzetsu:
A bug report: currently we have

```terraform
data "aws_subnet" "runners" {
  id = var.subnet_id_runners
}

data "aws_availability_zone" "runners" {
  name = data.aws_subnet.runners.availability_zone
}
…
    aws_zone                  = data.aws_availability_zone.runners.name_suffix
…
      "amazonec2-zone=${aws_zone}",
```

This makes it impossible to use multiple AZs!

A fix is to simply allow `aws_zone` be set per runner:
```
data "aws_availability_zone" "subnets" {
  for_each = aws_subnet.foo
  name     = each.value.availability_zone
}
…
  runners = [
    for k, v in aws_subnet.codebuild : {
      name        = "${var.name}-${data.aws_availability_zone.subnets[k].name}-${var.env}"
      instance_id = "m5.large"
      aws_zone    = data.aws_availability_zone.subnets[k].name_suffix
      # off_peak_idle_count = v.id == aws_subnet.codebuild[0].id ? 1 : 0
    }
  ]
```

The `data` bit could live inside code to completely hide it from the user and it should Just Work™.

Fuuzetsu:
Another report: applying today I got `Error: Error creating CloudTrail: InsufficientS3BucketPolicyException: Incorrect S3 bucket policy is detected for bucket: <bucket-name>` so there seems to be some race condition as re-running the apply gave:

```terraform
  + resource "aws_cloudtrail" "cloudtrail" {
```
as the only resource left and it applied well.

Fuuzetsu:
w.r.t https://github.com/npalm/terraform-aws-gitlab-runner/pull/249#issuecomment-741572105

>A fix is to simply allow aws_zone be set per runner:

Please ignore this. You can do exactly that (set `aws_zone` for the runner) and it works properly. I think I messed something else up at the time.

Fuuzetsu:
Another thing: The whole CloudWatch event machinery to pull in a new runner configuration does not work unless SSM is enabled (`enable_runner_ssm_access        = true`). Without this, the instance is never considered as "Managed instance" in SSM terms and so the RunCommand is not able to target it. This should probably be set to `true` by default and/or there should be some guard that doesn't make the resources related to config updates if it's not enabled.

Fuuzetsu:
>This can mean that each of the registered runners will have its own token.

Not only does this allow to tag each executor in a different way, there is an additional advantage: we're also able to register multiple _specific_ runners. Right now the choice is either single project, group or shared runner. If we want to have a specific runner for each project, this currently means we need a gitlab runner bastion for each.

So it seems that parametrising over the block per executor would be a huge improvement on multiple fronts while still only using a single bastion.

Fuuzetsu:
Another report: changing the `gitlab_runner_registration_token` we have deployed once before has no effect. This is because when we initially registered with this value, in user_data we put the result in SSM:

```bash
token=$(aws ssm get-parameters --names "${secure_parameter_store_runner_token_key}" --with-decryption --region "${secure_parameter_store_region}" | jq -r ".Parameters | .[0] | .Value")
if [[ `echo ${runners_token}` == "__REPLACED_BY_USER_DATA__" && `echo $token` == "null" ]]
then
  token=$(curl --request POST -L "${runners_gitlab_url}/api/v4/runners" \
    --form "token=${gitlab_runner_registration_token}" \
    --form "tag_list=${gitlab_runner_tag_list}" \
    --form "description=${giltab_runner_description}" \
    --form "locked=${gitlab_runner_locked_to_project}" \
    --form "run_untagged=${gitlab_runner_run_untagged}" \
    --form "maximum_timeout=${gitlab_runner_maximum_timeout}" \
    --form "access_level=${gitlab_runner_access_level}" \
    | jq -r .token)
  aws ssm put-parameter --overwrite --type SecureString  --name "${secure_parameter_store_runner_token_key}" --value="$token" --region "${secure_parameter_store_region}"
fi
```

What this means that changing any of the parameters inside the `if` block has no useful effect (it user_data changes so it will remake the launch configuration and therefore the ASG and we'll get a new instance: but it's not going to do anything for us due to persisted global token state in SSM).

This can probably be addressed by storing what parameters we're using in the `if` block at token generation time and if these do not match when we retrieve the token (with metadata) then we should register again.

There does seem to be an attempt to not re-use the token when the gitlab runner instance gets cycled as evident here:
```bash
# Remove all running docker-machine instances
# Overwrite token in SSM with null and remove runner from Gitlab
stop() {
    logger "Removing Docker Machine Instances"
    for runner in $(docker-machine ls | awk '{print $1}' | grep runner); do docker-machine rm -y "$runner"; done
    logger "Removing Gitlab Runner Token"
    aws ssm put-parameter --overwrite --type SecureString  --name "${secure_parameter_store_runner_token_key}" --region "${secure_parameter_store_region}" --value="null" 2>&1 | logger &
    curl -sS --request DELETE "${runners_gitlab_url}/api/v4/runners" --form "token=$token" 2>&1 | logger &
    retval=\$?
    rm -f \$lockfile
    return \$retval
}
```

but this doesn't run when the instance is recreated, at least I don't see anything in CloudWatch. Perhaps it should be made into a systemd service that stops after the gitlab-runner service does? Still, we can't trust a clean shutdown so we should have a check at the start that data is all up to date too.

Fuuzetsu:
Another issue:
```terraform
  # module.gitlab_runner.aws_cloudwatch_log_group.environment[0] will be created
  + resource "aws_cloudwatch_log_group" "environment" {
      + arn               = (known after apply)
      + id                = (known after apply)
      + name              = "…"
      + retention_in_days = 0
      + tags              = {
          + "Environment" = "…"
          + "Name"        = "…"
        }
    }
```
```
Error: Creating CloudWatch Log Group failed: ResourceAlreadyExistsException: The specified log group already exists:  The CloudWatch Log Group '…' already exists.
```

So if instance starts first, it starts logging and creates the log group itself. Then terraform tries to do the same and fails. The group should be made before the ASG or something.

Fuuzetsu:
Today I realised there is very easy way to simplify the whole thing and have multiple executors. The main problem is about synchronising config against some source of truth. really, it's only the parameters used for registering executors and everything else is just part of the config in S3 that we send verbatim. We could put everything in SSM like we do with token but it means some complicated handling of changing all the parameters, making sure we properly register everything, deregister old tokens when changing params etc: generally a mess.

There's a much simpler way though: we already have a central source of truth, it's terraform itself (and its state file).

So simply we can register (and deregister) runners with curl for every runner and insert the tokens into config. This means the config we send is always complete and all the tokens are tracked properly and remade on parameter changes (and only if necessary: no unregistering every token because one param changed).

local-exec based resource to get tokens will just work. I may post a simple wrapper provider for it on Monday.

Fuuzetsu:
I've made a start here https://github.com/tsurucapital/terraform-provider-gitlab/tree/gitlab-runner-register

This allows to register each executor/runner via terraform separately and get tokens for each which was probably the biggest hurdle. With that, we don't have to register on the actual machine and mess with SSM etc. I will polish it up tomorrow I hope and ask for inclusion in official GitLab provider. Still WIP.

modules/config/main.tf
```diff
@@ -0,0 +1,304 @@
...
+  cloudtrail_bucket_arn  = "arn:aws:s3:::${local.cloudtrail_bucket}"
+
+  config_key         = var.config_key == "" ? "config.toml" : var.config_key
+  config_bucket      = length(var.config_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : "${var.name}-config"
+  config_bucket_name = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : aws_s3_bucket.config[0].bucket
+  config_bucket_arn  = "arn:aws:s3:::${local.config_bucket}"
+
+  extra_files_prefix = trim(var.extra_files_prefix == "" ? "/extra-files/" : var.extra_files_prefix, "/")
+
+  config_uri      = "s3://${aws_s3_bucket_object.config.bucket}/${aws_s3_bucket_object.config.key}"
```
Fuuzetsu, line 16: I understand wanting to use the exact values from the `aws_s3_bucket_object` but sadly terraform does not co-operate nicely here: those values are considered dynamic so for every single config change, we also end up generating noise in the terraform plan for any values that depend on `config_uri` and `extra_files_uri`, notably the SSM policy. Instead, if you apply the below diff, even if _content_ of the config file changes, it doesn't produce any extra noise in the plan to do with SSM policies.

```diff
diff --git a/modules/config/main.tf b/modules/config/main.tf
index 6d8f0ad..a1b03aa 100644
--- a/modules/config/main.tf
+++ b/modules/config/main.tf
@@ -13,8 +13,8 @@ locals {
 
   extra_files_prefix = trim(var.extra_files_prefix == "" ? "/extra-files/" : var.extra_files_prefix, "/")
 
-  config_uri      = "s3://${aws_s3_bucket_object.config.bucket}/${aws_s3_bucket_object.config.key}"
-  extra_files_uri = "s3://${aws_s3_bucket_object.config.bucket}/${local.extra_files_prefix}"
+  config_uri      = "s3://${local.config_bucket_name}/${local.config_key}"
+  extra_files_uri = "s3://${local.config_bucket_name}/${local.extra_files_prefix}"
 
   post_reload_script = length(var.post_reload_script) > 0 ? "echo \"Executing post_reload_config script...\"\n${var.post_reload_script}" : ""
 }
```

template/runner-config-runners.tpl
```diff
@@ -0,0 +1,63 @@
...
+  post_build_script = "${post_build_script}"
+  pre_clone_script = "${pre_clone_script}"
+  request_concurrency = ${request_concurrency}
+  output_limit = ${output_limit}
+  limit = ${limit}
+  [runners.docker]
+    tls_verify = false
+    image = "${image}"
+    privileged = ${privileged}
+    disable_cache = false
```
Fuuzetsu, line 17: The auto-scaling docs specifically set `disable_cache = true` so maybe we should too: https://docs.gitlab.com/runner/configuration/runner_autoscale_aws/#the-runnersdocker-section

template/runner-config-runners.tpl
Fuuzetsu: Something from my branch that replaces these and makes it more flexible: https://github.com/tsurucapital/terraform-aws-gitlab-runner/commit/f4b0fd4fef83273e37ffc1465ff7834555cc7b77

Fuuzetsu:
> I've made a start here https://github.com/tsurucapital/terraform-provider-gitlab/tree/gitlab-runner-register
> 
> This allows to register each executor/runner via terraform separately and get tokens for each which was probably the biggest hurdle. With that, we don't have to register on the actual machine and mess with SSM etc. I will polish it up tomorrow I hope and ask for inclusion in official GitLab provider. Still WIP.

Until I have time to finish that up, properly, anyone that wants to try it out now can: https://registry.terraform.io/providers/tsurucapital/gitlab/latest/docs/resources/runner_register

Fuuzetsu:
> There's a much simpler way though: we already have a central source of truth, it's terraform itself (and its state file).

That can be found in https://github.com/tsurucapital/terraform-aws-gitlab-runner/tree/multiple-runners-static-config , it seems to work quite well. Still work in progress.

modules/config/main.tf
```diff
@@ -0,0 +1,304 @@
+locals {
+  cloudtrail_name = "${var.name}-config-trail"
+
+  cloudtrail_prefix      = var.cloudtrail_prefix == "" ? "gitlab-runner-config" : var.cloudtrail_prefix
+  cloudtrail_bucket      = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.cloudtrail[0].bucket : "${var.name}-cloudtrail"
+  cloudtrail_bucket_name = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.cloudtrail[0].bucket : aws_s3_bucket.cloudtrail[0].bucket
+  cloudtrail_bucket_arn  = "arn:aws:s3:::${local.cloudtrail_bucket}"
```
Fuuzetsu, line 7: Similarly to comment on `config_uri`, creating this name by hand means that we don't have a dependency visible to terraform. In case we leave `var.config_bucket` empty, we end up using `${var.name}-cloudtrail` as the value here and we end up with static string. Now terraform tries to make the `resource "aws_cloudtrail" "cloudtrail"` using this `local.config_bucket_arn` value as part of the policy. We _do_ create the bucket but it tends to take longer to do than making the cloudtrail policy. So AWS sees the ARN doesn't refer to a real bucket and we end up with:
```
Error: Error creating CloudTrail: InsufficientS3BucketPolicyException: Incorrect S3 bucket policy is detected for bucket: …-cloudtrail
```

What has to happen is we take this ARN from an actual terraform resource to enforce proper ordering.

So something like

```hcl
cloudtrail_bucket_arn  = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.cloudtrail[0].arn : aws_s3_bucket.cloudtrail[0].arn
```

This then means that we have a cycle as the bucket policy mentions the ARN which we get from the bucket... so we have to resolve that via `aws_s3_bucket_policy` and attach the policy after bucket creation. This gets us back to the original problem: we create the cloudtrail before some resource is ready (before it was the bucket, now it's the policy). So we also need a `depends_on` on the `aws_s3_bucket_policy` in the cloudtrail.

With all these pieces in place, `terraform apply` just works again.

You can see all the changes at https://github.com/tsurucapital/terraform-aws-gitlab-runner/commit/751420121ade306cdebbb9d6eb186822579bf301

modules/config/main.tf
```diff
@@ -0,0 +1,304 @@
...
+
+  config_key         = var.config_key == "" ? "config.toml" : var.config_key
+  config_bucket      = length(var.config_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : "${var.name}-config"
+  config_bucket_name = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : aws_s3_bucket.config[0].bucket
+  config_bucket_arn  = "arn:aws:s3:::${local.config_bucket}"
```
Fuuzetsu, line 12: Exactly the same issue as for cloudtrail_bucket_arn. This should probably say:

```hcl
config_bucket_arn  = length(var.config_bucket) > 0 ? data.aws_s3_bucket.config[0].arn : aws_s3_bucket.config[0].arn
```

modules/config/main.tf
```diff
@@ -0,0 +1,304 @@
...
+  cloudtrail_name = "${var.name}-config-trail"
+
+  cloudtrail_prefix      = var.cloudtrail_prefix == "" ? "gitlab-runner-config" : var.cloudtrail_prefix
+  cloudtrail_bucket      = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.cloudtrail[0].bucket : "${var.name}-cloudtrail"
+  cloudtrail_bucket_name = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.cloudtrail[0].bucket : aws_s3_bucket.cloudtrail[0].bucket
+  cloudtrail_bucket_arn  = "arn:aws:s3:::${local.cloudtrail_bucket}"
+
+  config_key         = var.config_key == "" ? "config.toml" : var.config_key
+  config_bucket      = length(var.config_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : "${var.name}-config"
+  config_bucket_name = length(var.cloudtrail_bucket) > 0 ? data.aws_s3_bucket.config[0].bucket : aws_s3_bucket.config[0].bucket
```
Fuuzetsu, line 11: This looks wrong, probably a copy+paste error? Surely the condition should be `length(var.config_bucket) > 0`, _not_ `cloudtrail_bucket`.

Fuuzetsu:
I am finished making changes on https://github.com/tsurucapital/terraform-aws-gitlab-runner/tree/multiple-runners-static-config ; I have started using the `terraform` provider and keeping compatibility with both is annoying so I am forking the project to be terraform driver specific for our use-case. Please feel free to take anything you need from `multiple-runners-static-config` branch!

template/user-data.tpl
```diff
@@ -12,11 +12,27 @@ EOL
...
 ${eip}
 
+echo ""
+echo "Installing AWS CLI..."
+AWSCLI_LOG=/var/log/aws-cli-install.log
+touch $AWSCLI_LOG
+
+for i in {1..7}; do
+  echo "Attempt: ---- " $i
+  curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${aws_cli_version}.zip" -o "awscliv2.zip" >>$AWSCLI_LOG 2>&1 && break || sleep 60
```
Fuuzetsu, line 22: This line means we can't use ARM instances to run gitlab-runner instance. These are usually cheaper and worth using.

`arch` command is available at least on Amazon AMIs, it may be worth probing. Perhaps something like

```diff
diff --git a/template/user-data.tpl b/template/user-data.tpl
index 6eac5b3..2775564 100644
--- a/template/user-data.tpl
+++ b/template/user-data.tpl
@@ -17,9 +17,15 @@ echo "Installing AWS CLI..."
 AWSCLI_LOG=/var/log/aws-cli-install.log
 touch $AWSCLI_LOG
 
+ARCH=$(arch)
+if [ $? -ne 0 ]; then
+  echo "Failed to detected machine architecture for AWS CLI."
+  exit 1
+fi
+
 for i in {1..7}; do
   echo "Attempt: ---- " $i
-  curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${aws_cli_version}.zip" -o "awscliv2.zip" >>$AWSCLI_LOG 2>&1 && break || sleep 60
+  curl "https://awscli.amazonaws.com/awscli-exe-linux-$${ARCH}-${aws_cli_version}.zip" -o "awscliv2.zip" >>$AWSCLI_LOG 2>&1 && break || sleep 60
 done
 unzip awscliv2.zip >>$AWSCLI_LOG 2>&1
 sudo ./aws/install >>$AWSCLI_LOG 2>&1
```

examples/runner-custom-terraform-spot-fleet/main.tf
```diff
@@ -0,0 +1,265 @@
...
+      spot_fleet_tagging_role_arn = aws_iam_role.spot_fleet_tagging.arn
+      tags = {
+        "tf-aws-gitlab-runner:instancelifecycle" = "spot:yes"
+      }
+    })
+  }
+
+  # Let's run Terraform & docker-machine-driver-terraform installation script on boot
+  userdata_pre_install = <<SCRIPT
+    chmod +x /extra-files/docker-machine-terraform.sh
```
Fuuzetsu, line 171: This example suffers a bit. If we change the `docker-machine-terraform.sh` in the `extra_files`, we will indeed get the new version of the files on the machine (via SSM command). But the new version will not be installed.

We want to taint the machine somehow if the `docker-machine-terraform.sh` content changes  so that it's re-made with the new content.

A simple example could be to include a hash of the realised `docker-machine-terraform.sh` file in the `userdata`, perhaps as a comment or something. This will taint the ASG and force the machine replacement.

modules/config/main.tf
```diff
@@ -0,0 +1,304 @@
...
+  description = "Watch configuration changes for Gitlab runner."
+
+  event_pattern = jsonencode({
+    source = ["aws.s3"]
+    detail = {
+      eventSource = ["s3.amazonaws.com"]
+      eventName   = ["PutObject", "CompleteMultipartUpload"]
+      requestParameters = {
+        bucketName = [aws_s3_bucket_object.config.bucket]
+        key = concat([aws_s3_bucket_object.config.key], [
```
Fuuzetsu, line 283: These keys are known up-front rather than being dynamic. This means that if we add a new extra file, there's a race condition between the cloudwatch rule being updated and the file in question being uploaded.

Perhaps the `aws_s3_bucket_object` should depend on this `aws_cloudwatch_event_rule`? I think we want to update the rule first and upload second: this allows blocks like

```hcl
post_reload_config = "(cd /extra-files/runners-config; zip -r ../runners-config.zip *)"
```

to actually work.

examples/runner-custom-terraform-spot-fleet/main.tf
```diff
@@ -0,0 +1,265 @@
...
+    /extra-files/docker-machine-terraform.sh
+    # We have to pass TF config as zip file, because docker-machine-driver-terraform symlinks a folder if passed directly
+    (cd /extra-files/runners-config; zip -r ../runners-config.zip *)
+  SCRIPT
+
+  # After each new configuration change that happens on S3, don't forget to reload runners-config
+  # extra_files we passed earlier, are pulled to `/extra-files` directory on Gitlab Agent instance
+  # Let's zip the contents of runner config after each config update
+  post_reload_config = "(cd /extra-files/runners-config; zip -r ../runners-config.zip *)"
```
Fuuzetsu, line 180: The command here probably should be `zip -r ../runners-config.zip .`. If we use the wildcard `*`, this will not pick up any hidden files. An example of a useful hidden file we might have is `.terraform.lock.hcl` file which is used from terraform 0.13. Specifying it allows minimise surprises so it's quite nice to include.

trallnag:
> Instead it'd be cool if each executor was registered by itself.

Is that possible with docker+machine? I assumed everything goes through the runner agent And this is why we need a separate runner agent for every set of executors.

Fuuzetsu:
> > Instead it'd be cool if each executor was registered by itself.
> 
> Is that possible with docker+machine? I assumed everything goes through the runner agent And this is why we need a separate runner agent for every set of executors.

You don't need a separate agent for different executors. You can use multiple `[[runners]]` section in single `config.toml`, each with their separate tokens etc. You can see something like this is the repo/branch I mentioned somewhere earlier. Whether it's `docker+machine` or some other executor doesn't really matter.

We actually run different executor per GitLab repository we want CI for and even more, we have separate executors for different EC2 instance types. They all get their own token etc and a single machine deals with all this.

vgropp:
we are interested as well, is here any help needed to get this working @krzysztof-miemiec, @npalm?  For me it looks like @Fuuzetsu took the work further.

npalm:
Sorry I am not working on the PR right now, but I am open to community support!

krzysztof-miemiec:
I'm really ashamed I didn't continue work on this PR 😞 I had very little time recently and didn't want to engage in extra OSS  activities in because of that.

I updated the example in this repo with my findings from around September last year and did my best to rebase my fork on top of latest master. I hope it'll help you to continue towards the right direction. Check out `README.md` in example - this was pretty much my setup which worked nicely with spot instances.

Fuuzetsu:
While I probably don't plan to further evolve my branch/changes and instead took the path of "strip the stuff we don't need for our use-case" (sorry), I would like to share the finding about the multiple-runners in one machine approach: this turns out that it can very quickly overwhelm the machine if we use the `terraform`-backed docker-machine executor as in this branch. `terraform` turns out to be pretty resource-hungry and if you have a CI pipeline that has say 10 parallel jobs, it's going to run 10 instances of `terraform` at once and probably grind your small gitlab runner instance to a halt.

This resulted in us using a bigger (more expensive) EC2 instance type and limiting number of jobs to like 20: which is silly because we only wanted to protect against usage spikes.

I am planning to work on something (again, for our use-case but it may be made public for others to modify) that instead of using terraform to provision the resources, will instead simply ask AWS to apply a CloudFormation stack: this lets AWS do all the heavy work and we just make simple HTTP request, completely skipping any need for terraform and such.

Hopefully this insight saves some head-ache for anyone that was planning to use this with terraform as-is with many jobs.

Fuuzetsu:
> I am planning to work on something (again, for our use-case but it may be made public for others to modify) that instead of using terraform to provision the resources, will instead simply ask AWS to apply a CloudFormation stack: this lets AWS do all the heavy work and we just make simple HTTP request, completely skipping any need for terraform and such.

I started this at https://github.com/tsurucapital/docker-machine-driver-cloudformation in case anyone finds this useful.

In hindsight I think it's actually usable without changes to this project/my branch of this project: it's just the user data scripts that change and the executor type which is great.

rgarrigue:
Just a word for cheering up on getting this PR through :-)

VasundharaKumar2020:
@npalm I am using this module and has been very helpful so thank you but it would be great to get this CR out like everyone else I landed up into this issue too.

veshant:
This would really be useful.  Happy to help, but don't know where this was left off.

krzysztof-miemiec:
I'm no longer contributing to this PR, but my old fork/branch should be still functional. 🤔 I managed to get it up and running in my previous project, with one caveat of setting up a cleanup cron job for whenever something randomly fails to clean up instances automatically.

If you're willing to help, the first step would be to resolve conflicts and then verify if it still works. 🙌🏻

kayman-mk:
Conflicts resolved and pushed to https://github.com/Hapag-Lloyd/terraform-aws-gitlab-runner/tree/feat/multiple-runners

kayman-mk:
It seems, that this PR also add some other changes which do not match the PR title:
- setting the docker-machine driver
- having a S3 bucket for the configuration file + CloudTrail stuff

kayman-mk:
Everything is rebased on current `main` but not tested.

kayman-mk:
Especially the config module introducing the ssm_document looks good to me. But the initial reason to introduce it, was, that the 16k user data limit was reached. This has been fixed in the meantime by gzipping the content. As many new resources are introduced, I tend to open a separate issue and push a copy of the branch to the main repository to save the work.

PR #250, opened 2020-08-31
Update version add dependabot

npalm:
## Description
- upgrade terraform modules in examples
- bump default runner to 13.3

## Migrations required
No

## Verification
default exaple


kayman-mk:
Good point. Dependabot is a nice tool to update the used modules to the newest version frequently.

PR #255, opened 2020-10-05
Add a kms alias to kms key

Michenux:
## Description
Add a kms alias to the kms key (for log group).

## Migrations required
NO

## Verification
One runner instance with spot runners


npalm (commented):
Code looks good, wil verify the change later this week.

npalm (approved):
@Michenux thanks!

PR #257, opened 2020-10-05
feat: allow setting gitlab-runner instance egress rules

mhulscher:
## Description
Currently, the module allows all egress traffic from the gitlab-runner instance. From a security point of view, I might want to limit the egress network connectivity from the gitlab-runner instance. This PR introduces a new variable called `gitlab_runner_egress_rules` that allows the user of this module to configure additional egress rules.

## Migrations required
NO, however the new default egress rule for the gitlab-runner now also allows egress traffic to all IPv6 address.

## Documentation
:heavy_check_mark:


npalm (approved):
@mhulscher thanks

PR #258, opened 2020-10-06
Add option to attach additional policies to the runner agent instance

wojukasz:
## Description
Ability to attach additional policies is quite helfpuf, for example if one would like to use gitlabrunner to build golden AMIs.

## Verification
I have tested proposed changes and was able to deploy and attach additional policy successfully.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -581,6 +581,6 @@ variable "log_group_name" {
 
 variable "runner_iam_policy_arns" {
   type        = list(string)
-  description = "List of policy ARNs to be added to the instance profile of the runners."
+  description = "List of policy ARNs to be added to the instance profile of the gitlab runner agent ec2 instance."
   default     = []
 }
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -578,3 +578,9 @@ variable "log_group_name" {
   default     = null
   type        = string
 }
+
+variable "runner_iam_policy_arns" {
+  type        = list(string)
+  description = "List of policy ARNs to be added to the instance profile of the runners."
```
npalm, line 584:
```suggestion
  description = "List of policy ARNs to be added to the instance profile of the gitlab runner agent ec2 instance."
```

npalm (changes requested):
@wojukasz thanks, just a small comment. Can you fix the description?

variables.tf
wojukasz: Thanks, corrected.

npalm (approved):
@wojukasz thank for you contribution

PR #269, opened 2020-11-12
Allow custom policies to be attached to the docker machine profile

bliles:
## Description
Provide support for user defined IAM policies to be attached to the docker machine runners. This allows pipelines that need to access IAM resources to have their permissions defined by the docker machine profile.

## Migrations required
NO

## Verification
Ran the ci verification steps. Also tested with my desired use case, attached a custom policy to the docker machine instance role.

## Documentation
Documentation was updated.


npalm (commented):
looks goodl will run a quick check asap

npalm (approved):

PR #273, opened 2020-12-10
Adding docker runtime

thomaskelm:
## Description
I was in need of the nvidia docker runtime in order for my gitlab jobs to be able to access the GPUs on the instances I was using and figured that someone else using this module in the future may also like this option.

I found where to add the value in this GitLab ticket https://gitlab.com/gitlab-org/gitlab-runner/-/issues/2465#note_65167962

## Migrations required
NO

## Verification
I have deployed using this version from my github and was able to see the GPU now.

Prior to change:
image

After this change and passing in "nvidia" to the runner_docker_runtime variable:
image

## Documentation
Can't tell if I'm crazy or not, but I can't seem to find the __doc/README.md file, but I updated the one in root


npalm:
@thomaskelm will check later this week, looks good. About the docs. The repo uses pre-commit hooks, those hooks will update via terraform-docs the readme file. The `_doc` folder was an old approach and should be removed completely. I will update the docs, thanks for mentioning.

npalm:
merged in develop instead of master

npalm:
@thomaskelm thanks!

PR #278, opened 2021-01-22
Feat: Support Security Group custom description

pandarouxbsd:
## Description
Just a small update to be able to customize the security groups description for both gitlab runner and docker machine instances.

## Migrations required
 NO

## Verification
Verified on runner-docker example

## Documentation
Documentation updated in runner-docker example

Please ensure you update the README in `_docs/README.md`. The README.md in the root can be updated by running the script `ci/bin/autodocs.sh`, requires `terraform-docs` version 0.8+

### DIFF
--- a/examples/runner-docker/variables.tf
+++ b/examples/runner-docker/variables.tf
@@ -31,5 +31,6 @@ variable "gitlab_url" {
 }
 
 variable "registration_token" {
+  default = "UkCzuTzyM8rBzs68Hpsu"
 }
 
--- a/examples/runner-docker/variables.tf
+++ b/examples/runner-docker/variables.tf
@@ -31,6 +31,5 @@ variable "gitlab_url" {
 }
 
 variable "registration_token" {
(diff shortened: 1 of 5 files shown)


examples/runner-docker/variables.tf
```diff
@@ -31,5 +31,6 @@ variable "gitlab_url" {
 }
 
 variable "registration_token" {
+  default = "UkCzuTzyM8rBzs68Hpsu"
```
npalm, line 34: Please can you remove the token, in case a real token would suggest to reset the token.

npalm (changes requested):

examples/runner-docker/variables.tf
pandarouxbsd: Hi,

It's done,

Thank you very much for your warning.
It was on a fake project that I previously deleted.

npalm (approved):

npalm:
@pandarouxbsd thanks for your contribution 👍

PR #293, opened 2021-02-11
Add option to customize helper image

stefan-kolb:
## Description

We need an option to customize the helper image used by the executor. I therefore propose to add the variable `runners_helper_image` to allow for this option. We would be happy if you can include this into a subsequent release. Thanks for your great work!

## Migrations required

NO

## Verification

Local deployment.

## Documentation

Added to README.


npalm (commented):
:+1 LGTM, will run a check asap.

npalm (approved):
@stefan-kolb thanks!

PR #295, opened 2021-02-13
Restrict public access and public objects for cache bucket

stefan-kolb:
## Description

This PR entirely restricts the public access to the cache bucket. Currently, the bucket is private via ` acl    = "private"` but objects might be set to public via an owner (not likely but AWS shows a arning, so this is merely a security/warning fix). This is shown as a warning in the AWS UI (see images - sory for german UI). The PR resolves these warnings. Functionality should not be affected as all runners have an access policy for the bucket.

image

image

## Migrations required

NO

## Verification

Used in my deployment.

## Documentation

No documentation needed, buckets are already private as default and its not changeable.

### DIFF
--- a/modules/cache/main.tf
+++ b/modules/cache/main.tf
@@ -61,6 +61,16 @@ resource "aws_s3_bucket" "build_cache" {
   }
 }
 
+# block public access to S3 cache bucket
+resource "aws_s3_bucket_public_access_block" "build_cache_policy" {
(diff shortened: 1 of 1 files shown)


npalm (commented):
LGTM to mee, will run asap a verification test.

npalm (approved):
@stefan-kolb thanks!

PR #296, opened 2021-02-16
fix: respect create_cache_bucket variable and avoid concurrent change…

stefan-kolb:
## Description

I missed two problems with my previous PR (https://github.com/npalm/terraform-aws-gitlab-runner/pull/295):
- respect `create_cache_bucket` variable 
- avoid concurrent changes to cache bucket (apply policy only after bucket is created)

Sorry for my mistakes!

## Migrations required

NO

## Verification

I have tested this with a staging deployment.

PR #306, opened 2021-02-22
Add count statement to build cache policy

ainskeep:
If create_cache_bucket is set to false, we don't want to build the S3 policy, this line should stop that from happening


stefan-kolb:
Duplicate of https://github.com/npalm/terraform-aws-gitlab-runner/pull/296

npalm:
@stefan-kolb thanks for make the note of the duplicate, still need  to check that one

npalm:
Thanks for thaking the time of the contribution, already solved via #296

Issue #271, opened 2020-12-09
Install the docker-ecr-credential-helper


Fuuzetsu:
If we want to use docker-based executors and pull docker images from private ECR, we are likely to use https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#using-credential-helpers with DOCKER_AUTH_CONFIG set. To make use of this, the appropriate credential helper needs to be installed: https://github.com/awslabs/amazon-ecr-credential-helper

With current EC2 setup, this is just `yum install -y amazon-ecr-credential-helper`. The user can work-around this by using `userdata_post_install` variable but it'd be cool if there was a proper setting. It could even be on by default, there's not much reason (as far as I can tell) to not just install this by default.

npalm:
Thanks, great suggestion!

npalm:
solved via #311

PR #308, opened 2021-02-27
feat: amazon-ecr-credential-helper

anasinnyk:
depends on npalm/terraform-aws-gitlab-runner#271

## Description

It's will be awesome if we can use ECR credentials helper in runner

## Migrations required

NO

### DIFF
--- a/template/gitlab-runner.tpl
+++ b/template/gitlab-runner.tpl
@@ -48,6 +48,9 @@ docker-machine rm -y dummy-machine
 unset HOME
 unset USER
 
+# Install amazon-ecr-credential-helper
+yum install amazon-ecr-credential-helper -y
+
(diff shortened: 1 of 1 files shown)


npalm (changes requested):
Thanks for submitting the PR. The change could be ok. But in this way it force all the users to always have the ecr credential helper installed. You can allready instal the the credentialhelpe by using the pre_instaal and push the line tyou have added.

Adding the option would be fine for me, but can then I prefer to have the additonal fully configurable. So the user of the module can make the choice. The default shoud be in that case not installed. you can introduce a variable as `install_.....` Please have a look on other configuration options in the templates

npalm:
Thanks for the PR, feature added via  #311

PR #324, opened 2021-05-20
Make disable_cache configurable

pierlucg-xs:
## Description

Runners are running out of space rapidly because they do not discard local docker images

## Migrations required

NO

## Verification

Tested with my own TF setup (runners_default)

## Documentation

Updated!

------
I've left the default to `false` because this is how this project was currently set-up, however I believe `disable_cache` should be set to `true` for most use-cases when using auto-scaling, as described in [Gitlab's documentation](https://docs.gitlab.com/runner/configuration/runner_autoscale_aws/#the-runnersdocker-section)


npalm (approved):
@pierlucg-xs thanks for your contribution. looks good.

PR #331, opened 2021-06-21
fix: add tags for instance profile

akli1512:
the resource "aws_iam_instance_profile" "instance"  has not tags, we need the tags for policy verification purposes


MeNsaaH:
Can you extend the tags to root_block_device volumes? https://github.com/npalm/terraform-aws-gitlab-runner/blob/develop/main.tf#L218

npalm:

npalm (approved):

PR #334, opened 2021-07-03
Remove the null_data_source (#332)

kayman-mk:
## Description

As described in #332 `null_data_source` can be replaced by locals to avoid the warning message during plan/apply.

## Migrations required

NO

## Verification

Checking the tags output in the plan:
```
      + tags                      = [
          + {
              + "key"                 = "Environment"
              + "propagate_at_launch" = "true"
              + "value"               = "prod-gitlab-ci-runner-terraform"
            },
          + {
              + "key"                 = "Name"
              + "propagate_at_launch" = "true"
              + "value"               = "Gitlab Runner Terraform Agent"
            },
          + {
              + "key"                 = "allowReadPlatformCredentials"
              + "propagate_at_launch" = "true"
              + "value"               = "gitlab-ci-runner"
            },
          + {
              + "key"                 = "app"
              + "propagate_at_launch" = "true"
              + "value"               = "runner"
            },
          + {
              + "key"                 = "environment"
              + "propagate_at_launch" = "true"
              + "value"               = "prod"
            },
          + {
              + "key"                 = "service"
              + "propagate_at_launch" = "true"
              + "value"               = "gitlab-ci"
            },
          + {
              + "key"                 = "team"
              + "propagate_at_launch" = "true"
              + "value"               = "sales"
            },
        ]
```

PR #336, opened 2021-07-06
fix: replace deprecated null_data_source with locals

Michenux:
## Description

null_data_source should not be used anymore.
Terraform locals should be used instead.

A warning is displayed with terraform 1.0.1.

## Migrations required

 NO


kayman-mk:
See also #334

npalm:

npalm (approved):

PR #337, opened 2021-07-06
feat: replace launch configuration with launch template

Michenux:
## Description

AWS recommend that we should use launch templates instead of launch configurations.

With the launch template, we can now use a CMK for storage encryption. Globally, launch template allows
to configure more things than the launch configuration.

## Migrations required

NO

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -162,9 +162,22 @@ resource "aws_autoscaling_group" "gitlab_runner_instance" {
   max_size                  = "1"
   desired_capacity          = "1"
   health_check_grace_period = 0
-  launch_configuration      = aws_launch_template.gitlab_runner_instance.name
   enabled_metrics           = var.metrics_autoscaling
   tags                      = local.agent_tags_propagated
+
+
+  launch_template {
+    id      = aws_launch_template.gitlab_runner_instance.id
+    version = aws_launch_template.gitlab_runner_instance.latest_version
+  }
+
+  instance_refresh {
+    strategy = "Rolling"
+    preferences {
+      min_healthy_percentage = 0
+    }
+    triggers = ["tag"]
+  }
 
   timeouts {
     delete = var.asg_delete_timeout
@@ -211,12 +224,13 @@ resource "aws_launch_template" "gitlab_runner_instance" {
   image_id      = data.aws_ami.runner.id
   user_data     = local.template_user_data
   instance_type = var.instance_type
+  update_default_version = true
   ebs_optimized = var.runner_instance_ebs_optimized
   monitoring {
     enabled = var.runner_instance_enable_monitoring
   }
   dynamic instance_market_options {
-    for_each = length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
+    for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
     content {
       market_type = instance_market_options.value
       spot_options {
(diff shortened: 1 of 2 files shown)


npalm:

npalm (changes requested):
Need a bit more time to check the PR. I ran a quick test with the default example. The plan is failing for this example. In this example `runner_instance_spot_price` is not set. In this case the agent will run on a NON spot instance. Which is the default. This will causse the following error:


```
Error: Invalid function argument

  on ../../main.tf line 219, in resource "aws_launch_template" "gitlab_runner_instance":
 219:     for_each = length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
 
 ```
 
 Please can you also rebase with develop. I have fixed the CI jobs

**Comment by Michenux** [comment]
I have fixed the issues. Note that i have added an instance_refresh parameter on the runner autoscaling group to recreate the instance runner when a config is changed.

**Comment by npalm** [comment]
sorry for the late response but will check the PR

**Comment by npalm** [comment]
@Michenux great work, sorry for the late response. But I still got the following error while running the default example, see also review comment:

```bash

Error: Invalid function argument

  on ../../main.tf line 233, in resource "aws_launch_template" "gitlab_runner_instance":
 233:     for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
    |----------------
    | var.runner_instance_spot_price is null

Invalid value for "value" parameter: argument must not be null.


Error: Invalid function argument

  on ../../main.tf line 271, in resource "aws_launch_template" "gitlab_runner_instance":
 271:     for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
    |----------------
    | var.runner_instance_spot_price is null

Invalid value for "value" parameter: argument must not be null.
```

Did you fixed this issue? See only commits about formatting after my review.

main.tf
```diff
@@ -205,30 +218,63 @@ data "aws_ami" "runner" {
...
-  ebs_optimized        = var.runner_instance_ebs_optimized
-  enable_monitoring    = var.runner_instance_enable_monitoring
-  spot_price           = var.runner_instance_spot_price
-  iam_instance_profile = aws_iam_instance_profile.instance.name
-  dynamic "root_block_device" {
+resource "aws_launch_template" "gitlab_runner_instance" {
+  name_prefix            = var.runners_name
+  key_name               = var.ssh_key_pair
+  image_id               = data.aws_ami.runner.id
+  user_data              = local.template_user_data
```
npalm, line 225:
```suggestion
  user_data              = base64encode(local.template_user_data)
```

main.tf
```diff
@@ -205,30 +218,63 @@ data "aws_ami" "runner" {
...
+  instance_type          = var.instance_type
+  update_default_version = true
+  ebs_optimized          = var.runner_instance_ebs_optimized
+  monitoring {
+    enabled = var.runner_instance_enable_monitoring
+  }
+  dynamic "instance_market_options" {
+    for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
```
npalm, line 233:
```suggestion
    for_each = var.runner_instance_spot_price == null ? [] : ["spot"]
```

main.tf
```diff
@@ -205,30 +218,63 @@ data "aws_ami" "runner" {
...
+  tag_specifications {
+    resource_type = "instance"
+    tags          = local.tags
+  }
+  tag_specifications {
+    resource_type = "volume"
+    tags          = local.tags
+  }
+  dynamic "tag_specifications" {
+    for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
```
npalm, line 271:
```suggestion
    for_each = var.runner_instance_spot_price == null ? [] : ["spot"]
```

npalm (changes requested):
got it working locally, need to test a bit more, sse suggestion for my changes

main.tf
Michenux: I would prefer to keep the test with the empty string.

Should work : 
```suggestion
    for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? ["spot"] : []
```

npalm:
Please can you also rebase develop, change in PR #348 requires a small update in the launch template.

Michenux:
I have rebased from develop and applied another solution for var.runner_instance_spot_price condition.

main.tf
npalm: I still get an error on this line, aslo the for_each for the tag_specification looks not correct. Shoul the loop not only run onse the agent instance / runner agent instance is configured as spot by setting `runner_instance_spot_price`. The code in the PR leads to a raise condition on the default example. I have changed the for_each line to:

```
for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? [] : ["spot"]
```
After this change and the change to tag_specification it looks working like a charm. Tested for both spot and non spot.

main.tf
```diff
@@ -205,34 +218,69 @@ data "aws_ami" "runner" {
...
+  tag_specifications {
+    resource_type = "instance"
+    tags          = local.tags
+  }
+  tag_specifications {
+    resource_type = "volume"
+    tags          = local.tags
+  }
+  dynamic "tag_specifications" {
+    for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? ["spot"] : []
```
npalm, line 271:
```suggestion
    for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? [] : ["spot"]
```

npalm (changes requested):
@Michenux thanks for all the hard work, sorry for the delay (once again). The current code is not working, once update the for_each on 2 places to `for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? [] : ["spot"]` the default example works. Tested for spot and non spot.

main.tf
```diff
@@ -205,34 +218,69 @@ data "aws_ami" "runner" {
...
+  image_id               = data.aws_ami.runner.id
+  user_data              = base64encode(local.template_user_data)
+  instance_type          = var.instance_type
+  update_default_version = true
+  ebs_optimized          = var.runner_instance_ebs_optimized
+  monitoring {
+    enabled = var.runner_instance_enable_monitoring
+  }
+  dynamic "instance_market_options" {
+    for_each = var.runner_instance_spot_price != null && length(var.runner_instance_spot_price) == 0 ? [] : ["spot"]
```
npalm, line 233: The array should be set with spot once the runner_instance_spot_price is set, same as for tag_specification
```suggestion
    for_each = var.runner_instance_spot_price == null || var.runner_instance_spot_price == "" ? [] : ["spot"]
```

npalm (changes requested):
Looks like you missed on configuration for the istance market option, have made a suggestion with the fix. 

Did you test one ot the examples?

main.tf
npalm: Tested with this changes and it looks all good!

npalm (approved):

PR #352, opened 2021-07-29
Support for setting Sentry DSN

radimsuckr:
## Description

[GitLab runners support reporting errors to Sentry](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-global-section) which makes troubleshooting a bit easier than going through system logs.

## Migrations required

NO

## Verification

Tested in our live GitLab runners. Did not run the examples.

## Documentation

On first apply it sets the Sentry DSN to empty string which pretty much is the default behaviour. It also creates the SSM parameter in which you should set the **legacy version** of your Sentry DSN and on the next startup it will use that DSN.

Btw when I ran `pre-commit run` it gave me following output which I find kind of suspicious.

```
Terraform fmt........................................(no files to check)Skipped
Terraform validate with tflint.......................(no files to check)Skipped
Terraform docs.......................................(no files to check)Skipped
Check for merge conflicts............................(no files to check)Skipped
```

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -9,7 +9,8 @@ locals {
   runners_max_builds_string = var.runners_max_builds == 0 ? "" : format("MaxBuilds = %d", var.runners_max_builds)
 
   // Define key for runner token for SSM
-  secure_parameter_store_runner_token_key = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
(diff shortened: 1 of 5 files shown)


npalm:
Will review in the weekend

npalm (approved):
@radimsuckr thanks!

PR #356, opened 2021-08-24
[Proposal] Use better ressources names

aperigault:
## Description

Launch template is named based on `runners_name`variable. It seems to be an error because this launch template is used to deploy the agent runner not docker machines. Tell me if i'm wrong.
Moreover security groups have tag `Name` with `local.name_sg`but the `name_prefix` do not use this local. Maybe it is an oversight ?

## Migrations required

NO, default values not change

## Verification

Minor changes


npalm (approved):
@aperigault Thanks, LGTM

PR #358, opened 2021-08-25
[Proposal] Add IAM objects name override option

aperigault:
## Description

I added a new key `name_iam_objects` in `overrides` map to rename all IAM objects (instance profile, role and policies).

## Migrations required

NO, default values not change

## Documentation

Ok


npalm (changes requested):
@aperigault Still planning to update the module to a later TF version. But for now the easiest way to fix the format errors in CI. Is as follow. Install latest terraform 0.3.x (via tfenv). With tfenv you can easy switch between terraform verions.

Please can you fix the format errors?

Change lookds good to me

aperigault:
CI is now ok but my terraform-docs do not generate same format. Do you use specific configuration or simply an older version ?

npalm (approved):

PR #369, opened 2021-09-17
fix: avoid creating some resources if runners_executor is not docker+machine

halilkaya:
## Description

No need to create the resources below since variable `runners_executor` does not have value `docker+machine`:

- `resource "aws_iam_role" "docker_machine"`
- `resource "aws_iam_instance_profile" "docker_machine"`
- `resource "aws_security_group" "docker_machine"`

## Migrations required

NO - If yes please describe the migration.

## Verification

## Documentation


halilkaya:
Hi @npalm , could you please review this MR? I'd like to hear your feedback.

npalm:
> Hi @npalm , could you please review this MR? I'd like to hear your feedback.

Thx for the reminder, will check over the weekend. Was on holiday 🌴 last month

npalm (changes requested):
@halilkaya thanks for taking the time to improve the module. You changes to avoid creation of not required resources are introducing a count on several reousrces. A count requires tha refering to resources refer to the resource in the list or loop over the list. Please see the failing CI on the examples. 

Some example error output
```

For example, to correlate with indices of a referring resource, use:
    aws_iam_role.docker_machine[count.index]


Error: Missing resource instance key

  on ../../outputs.tf line 33, in output "runner_role_name":
  33:   value       = aws_iam_role.docker_machine.name

Because aws_iam_role.docker_machine has "count" set, its attributes must be
accessed on specific instances.

For example, to correlate with indices of a referring resource, use:
    aws_iam_role.docker_machine[count.index]

```

halilkaya:
@npalm It was my bad. I pushed a new commit that contains fix for that error and some `terraform fmt` changes as well. `terraform validate` should no longer throw an error relevant to these changes.

halilkaya:
I think it's failed because of formatting it with the newer version of Terraform. I re-ran the `terraform fmt` command with Terraform 0.12 and pushed it. `Verify module` step should pass now.

outputs.tf
```diff
@@ -25,12 +25,12 @@ output "runner_agent_role_name" {
 
 output "runner_role_arn" {
   description = "ARN of the role used for the docker machine runners."
-  value       = aws_iam_role.docker_machine.arn
+  value       = aws_iam_role.docker_machine[0].arn
```
npalm, line 28: This will not work for the case you try to run the example for docker, e.g. a non docker-machine execturor. Did you check your change (actual deployment with a single instance)? The problem is that in that case the list is null, and you refer here in the output to the elment 0. 

Simple trick is to use a splat expression, add a default element (if null) and take the first element. 

An example looks like:

```hcl

output "runner_role_arn" {
  description = "ARN of the role used for the docker machine runners."
  value = element(concat(aws_iam_role.docker_machine.*.arn, [""]), 0)
}

```

outputs.tf
```diff
@@ -25,12 +25,12 @@ output "runner_agent_role_name" {
...
 }
 
 output "runner_role_name" {
   description = "Name of the role used for the docker machine runners."
-  value       = aws_iam_role.docker_machine.name
+  value       = aws_iam_role.docker_machine[0].name
```
npalm, line 33: please see previous comment

outputs.tf
```diff
@@ -40,7 +40,7 @@ output "runner_agent_sg_id" {
 
 output "runner_sg_id" {
   description = "ID of the security group attached to the docker machine runners."
-  value       = aws_security_group.docker_machine.id
+  value       = aws_security_group.docker_machine[0].id
```
npalm, line 43: please see previous comment

npalm (dismissed):
I checked the example with the default and docker example. Default works as expected no changes for an upgrade. Docker example fails, see comments for the output. Also 1 CI job is failing.

npalm:
I have updated several versions, including terraform. The modules can now be developed with the latest Terraform version. For usages it still supports older versions. Please can you rebase to the latest commit on develop? `git rebase -i develop`

kayman-mk (approved):

PR #374, opened 2021-10-04
feat: Add ability to define throughput for root block device on runner

IvanDechovsky:
## Description

Add ability to define throughput for root block device on runner


npalm (approved):
@IvanDechovsky thanks!

PR #381, opened 2021-10-08
Separate runner agent private address

tiborhercz:
## Description

Allow separate control of the private addresses for the runners
and the Gitlab agent.

You might want to have separate auto assign ip address settings
for the Gitlab agent and runners.

As an example when you which to have the Gitlab agent with a
public ip address so it can register itself to the Gitlab
cloud. But if you want the runners to be in a private subnet
and have the external ip of the NAT gateway instead of their
own ip.

This is needed when you have a CICD account which will
run deployments in a separate account. The CICD account will
host the custom Gitlab runner.
In the separate account you which to deploy to you want to whitelist 
the CICD ip addresses so it can perform updates to the infrastructure.
These ip addresses will be always the same, because the ip's are 
used from the NAT gateway which have their own eip's.

## Migrations required

YES, Users who are using public ip's addresses (`runners_use_private_address` is set to `false`) for their runners they have to set variable `runner_agent_uses_private_address` to `false` too.

## Verification

Tested it in my AWS account.


tiborhercz:
I have added the documentation and updated the description.
Thank you for your quick reply!

Issue #186, opened 2020-02-25
Remove the support for register the runner outside the module


npalm:
Due to backwards compatibly the module still supports the scenario to register a runner manual and pass the token. To reduce the complexity and number of scenarios this option will be removed.

trallnag:
A major version bump that throws out a bunch of deprecated stuff would be great

npalm:
fully agree, feel free to create a PR

kayman-mk:
@npalm As we always have to store the token somewhere: Which option should be removed (you are referring in your first post)?

dsalaza4:
We're currently having issues with the parameter store after upgrading from 6.3.1 to 6.5.2. 

This only happens when we run a `terraform plan` with a read-only role. A `terraform deploy` works fine.

Logs say:

```
│ Error: reading SSM Parameter (ci-common-large-x86-0-sentry-dsn): InvalidKeyId: arn:aws:kms:us-east-1:205810638802:key/b1bd2030-d46d-41b7-89ef-3ed07a9181fb is pending deletion. (Service: AWSKMS; Status Code: 400; Error Code: KMSInvalidStateException; Request ID: 81ba4e7e-f4b0-4f76-b1c3-9f6df3eac93f; Proxy: null)
│ 
│   with module.runners["common-large-x86-0"].aws_ssm_parameter.runner_sentry_dsn,
│   on .terraform/modules/runners/main.tf line 27, in resource "aws_ssm_parameter" "runner_sentry_dsn":
│   27: resource "aws_ssm_parameter" "runner_sentry_dsn" {
```

When inspecting the parameter store I see this:

<img width="939" alt="image" src="https://github.com/cattle-ops/terraform-aws-gitlab-runner/assets/11205987/286307c2-0812-4214-bb34-0fe9967f1e21">

I will try removing all parameters and letting the module recreate them to see what happens.

dsalaza4:
> We're currently having issues with the parameter store after upgrading from 6.3.1 to 6.5.2.
> 
> This only happens when we run a `terraform plan` with a read-only role. A `terraform deploy` works fine.
> 
> Logs say:
> 
> ```
[...]

PR #387, opened 2021-10-14
Remove deprecated resources

kayman-mk:
## Description

Remove all resources markes as deprecated (see #186)

## Migrations required

* Remove `aws_zone` variable. It's no longer in use. Delete it.
* Replace `runners_off_peak_*` variables with `runners_machine_autoscaling`.
* Remove `enable_forced_updates` variable. Use `enable_asg_recreation` instead.

## Verification

Deployed the new version of the module into our production environment and ran some build jobs. Worked fine.

## Documentation


npalm:
We have no issues with dnd, so just would like to learn what the issue is with dind. See here an example workflow that just work fines:

```yaml
build:
  image: docker:latest
  services:
    - docker:dind
  script:
    - docker version
    - docker build --pull -t blurk .
    - docker run blurk
```

PR #389, opened 2021-10-14
refactor!: Remove SSH connections from the outside

kayman-mk:
## Description

To connect to the EC2 instance SSM should be used as good practice. This PR drops the SSH support. So no SSH connections from the outside to the runner and/or agent are possible. For debugging purposes use the built in SSM support.

## Migrations required

If you use SSH to connect to your instances, switch to [SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html) which is safer and does not require to open ports to the outside.

```hcl
module "gitlab_runner" {
  enable_runner_ssm_access         = true # for the agent
  enable_docker_machine_ssm_access = true # for the spot instances
}
```

## Verification

Deployed the new version of the module into our production environment and ran some build jobs. Worked fine.

## Documentation


npalm:
@kayman-mk sorry for the late response, will check the PR asap. I agree SSM is a better practice but not sure I I prefer to drop the supprt fully.

kayman-mk:
Yeap, not an easy decision. I like SSH too. But you could still stay with it and tunnel it through an SSM connection. See https://www.techbeatly.com/2020/10/how-to-ssh-tunnel-through-aws-systems-manager-to-access-private-vpcresources.html

So on the developer machine SSH can be used as always but it is streamed through an SSM connection.

npalm:
Do my best to check the PR early next week

npalm:
I agree, dropping SSH is fine by me. Better to preach good practices. Please can you do me a favour and also update the README. The SSH keys option are mentioned several places. Please can you also refer to the example you mentioned: `you could still stay with it and tunnel it through an SSM connection. See https://www.techbeatly.com/2020/10/how-to-ssh-tunnel-through-aws-systems-manager-to-access-private-vpcresources.html` 

Tested and LGTM! Great work!

kayman-mk:
Oh yes, sure. I must have overlooked that.

kayman-mk:
@npalm Just started to amend the documentation and mentioned the tunnel through SSM feature. Well, if you can't set the SSH keys you won't be able to connect to the instance, right? So it makes no sense to reference it.

npalm (approved):

npalm (changes requested):
@kayman-mk please can you remove the file: https://github.com/kayman-mk/terraform-aws-gitlab-runner/blob/kayma/remove-ssh/examples/runner-public/key.tf This cause the failure of the CI

kayman-mk:
@npalm File is gone, pipeline green.

npalm (approved):

PR #392, opened 2021-10-15
Add ASG lifecycle management Lambda function

joshbeard:
## Description

This introduces an Auto Scaling Group instance termination lifecycle
hook using Lambda and related resources. The Lambda function is a Python
script that is triggered when the persistent runner instance in the ASG
is terminated. The function receives the instance ID of the "parent"
runner and queries for spawned instances that it launched to terminate.
Additionally, it will check for other "orphaned" instances that have a
`gitlab-runner-parent-id` tag that doesn't match an existing instance. This
resolves the issue where spawned instances could be orphaned when their
parent runner is terminated.

This feature is disabled by default.

The user data script is updated to provide the 'parent' instance ID as a
tag named 'gitlab-runner-parent-id' on spawned instances.

A new sub-module is provided called "terminate-workers". It is optional
to use this feature, and the input variable
`asg_terminate_lifecycle_hook_create` can be toggled `true` or `false`
for this behavior.

This partially addresses concerns discussed in issue #214

## Migrations required

NO

## Verification

I've tested existing standard configuration with this addition _disabled_ and _enabled_ with expected results.

* On a default setup without enabling this feature, there are 2 changes - the user data is updated to provide the `gitlab-runner` and `gitlab-runner-parent` tags to spawned worker instances.

    ```
    Plan: 0 to add, 2 to change, 0 to destroy.
    ````

* When enabled with its default configuration, there are 10 resource additions and the two changes mentioned previously.

    ```
    Plan: 10 to add, 2 to change, 0 to destroy.
    ```

## Documentation

Documentation was updated, tflint and pre-commit hooks ran.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -111,7 +111,7 @@ locals {
       runners_additional_volumes  = local.runners_additional_volumes
       docker_machine_options      = length(local.docker_machine_options_string) == 1 ? "" : local.docker_machine_options_string
       runners_name                = var.runners_name
-      runners_tags = replace(var.overrides["name_docker_machine_runners"] == "" ? format(
+      runners_tags = replace(replace(var.overrides["name_docker_machine_runners"] == "" ? format(
         "Name,%s-docker-machine,%s,%s",
         var.environment,
         local.tags_string,
@@ -121,7 +121,7 @@ locals {
         local.tags_string,
         local.runner_tags_string,
         var.overrides["name_docker_machine_runners"],
-      ), ",,", ",")
+      ), ",,", ","), "/,$/", "")
       runners_token                     = var.runners_token
       runners_executor                  = var.runners_executor
       runners_limit                     = var.runners_limit
@@ -503,4 +503,25 @@ resource "aws_iam_role_policy_attachment" "eip" {
 
   role       = aws_iam_role.instance.name
   policy_arn = aws_iam_policy.eip[0].arn
+}
+
+################################################################################
+### Lambda function for ASG instance termination lifecycle hook
+################################################################################
+module "terminate_instances_lifecycle_function" {
+  source = "./modules/terminate-instances"
+
+  count = var.asg_terminate_lifecycle_hook_create ? 1 : 0
+
+  name                                 = var.asg_terminate_lifecycle_hook_name == null ? "terminate-instances" : var.asg_terminate_lifecycle_hook_name
+  environment                          = var.environment
+  asg_arn                              = aws_autoscaling_group.gitlab_runner_instance.arn
+  asg_name                             = aws_autoscaling_group.gitlab_runner_instance.name
+  cloudwatch_logging_retention_in_days = var.cloudwatch_logging_retention_in_days
+  lambda_memory_size                   = var.asg_terminate_lifecycle_lambda_memory_size
+  lifecycle_heartbeat_timeout          = var.asg_terminate_lifecycle_hook_heartbeat_timeout
+  name_iam_objects                     = local.name_iam_objects
+  role_permissions_boundary            = var.permissions_boundary == "" ? null : "${var.arn_format}:iam::${data.aws_caller_identity.current.account_id}:policy/${var.permissions_boundary}"
+  lambda_timeout                       = var.asg_terminate_lifecycle_lambda_timeout
+  tags                                 = local.tags
 }
--- a/modules/terminate-instances/cloudwatch.tf
+++ b/modules/terminate-instances/cloudwatch.tf
@@ -1,0 +1,33 @@
+# ----------------------------------------------------------------------------
+# Terminate Instances - CloudWatch/EventBridge Resources
+#
+# This deploys an event rule and target for triggering the provided Lambda
+# function from the ASG lifecycle hook.
+# ----------------------------------------------------------------------------
+resource "aws_cloudwatch_event_rule" "terminate_instances" {
+  name        = "${var.environment}-${var.name}"
+  description = "Trigger GitLab runner instance lifecycle hook on termination."
+
+  event_pattern = <<EOF
+{
+  "source": ["aws.autoscaling"],
+  "detail-type": ["EC2 Instance-terminate Lifecycle Action"],
+  "detail": {
+    "AutoScalingGroupName": ["${var.asg_name}"]
+  }
+}
+EOF
+}
+
+resource "aws_cloudwatch_event_target" "terminate_instances" {
+  rule      = aws_cloudwatch_event_rule.terminate_instances.name
+  target_id = "${var.environment}-TriggerTerminateLambda"
+  arn       = aws_lambda_function.terminate_runner_instances.arn
+}
+
+resource "aws_cloudwatch_log_group" "lambda" {
+  name              = "/aws/lambda/${var.environment}-${var.name}"
+  retention_in_days = var.cloudwatch_logging_retention_in_days
+
+  tags = var.tags
+}
--- a/variables.tf
+++ b/variables.tf
@@ -734,3 +734,33 @@ variable "subnet_ids_gitlab_runner" {
   type        = list(string)
   default     = []
 }
+
+variable "asg_terminate_lifecycle_hook_name" {
+  description = "Specifies a custom name for the ASG terminate lifecycle hook and related resources."
+  type        = string
+  default     = null
+}
+
+variable "asg_terminate_lifecycle_hook_create" {
+  description = "Boolean toggling the creation of the ASG instance terminate lifecycle hook."
+  type        = bool
+  default     = true
+}
+
+variable "asg_terminate_lifecycle_hook_heartbeat_timeout" {
+  description = "The amount of time, in seconds, for the instances to remain in wait state."
+  type        = number
+  default     = 90
+}
+
+variable "asg_terminate_lifecycle_lambda_memory_size" {
+  description = "The memory size in MB to allocate to the terminate-instances Lambda function."
+  type        = number
+  default     = 128
+}
+
+variable "asg_terminate_lifecycle_lambda_timeout" {
+  description = "Amount of time the terminate-instances Lambda Function has to run in seconds."
+  default     = 30
+  type        = number
+}
--- a/README.md
+++ b/README.md
@@ -143,6 +143,19 @@ Once you have created the parameter, you must remove the variable `runners_token
 
 Finally, the runner still supports the manual runner creation. No changes are required. Please keep in mind that this setup will be removed in future releases.
 
+### Auto Scaling Group Instance Termination
+
+The Auto Scaling Group may be configured with a
+[lifecycle hook](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html)
+that executes a provided Lambda function when the runner is terminated to
+terminate additional instances that were spawned.
+
+The use of the termination lifecycle can be toggled using the
+`asg_termination_lifecycle_hook_create` variable.
+
+When using this feature, a `builds/` directory relative to the root module will
+persist that contains the packaged Lambda function.
+
 ### Access runner instance
 
 A few option are provided to access the runner instance:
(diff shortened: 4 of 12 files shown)


joshbeard:
Thanks for the review, @kayman-mk! I'll work on fixes and adjustments from your feedback. The Python requirement comes from the use of an external module (https://registry.terraform.io/modules/terraform-aws-modules/lambda/) for the Lambda function and is mentioned as a prerequisite when using this feature in the documentation updates in this pull request. However, I think we can get rid of this and handle this simple use case in this module. I'll do some refactoring for this.

On the issue of it not actually doing anything - I'm curious if you can check the CloudWatch event rule to see if it triggered an invocation and if it failed? There's also a CloudWatch log stream for the function itself. On the worker instance, do you see the `gitlab-runner-parent` tag with a value of the parent instance ID?

joshbeard:
@npalm rebased

joshbeard:
> @npalm Thank you so much for the thorough review and consideration! I will resolve the issues you pointed out, rebase, and push within the next couple of days.

Couple of days = couple of weeks ;) I've been away for the past couple of weeks.

main.tf
```diff
@@ -503,3 +503,24 @@ resource "aws_iam_role_policy_attachment" "eip" {
   role       = aws_iam_role.instance.name
   policy_arn = aws_iam_policy.eip[0].arn
 }
+
+################################################################################
+### Lambda function for ASG instance termination lifecycle hook
+################################################################################
+module "terminate_workers_lifecycle_function" {
+  source = "./modules/terminate-workers"
```
kayman-mk, line 511: question: Should we rename this to `terminate-runner-instances`? I checked the the Gitlab documentation and they have a Runner (the agent or coordinator running 24/7 and requesting new jobs) and Runner Instances (the machines doing the work). I am also not very precise with the wording in my PRs, but I think it would be better to do so. This should be done for all other occurrences as well.

main.tf
```diff
@@ -503,3 +503,24 @@ resource "aws_iam_role_policy_attachment" "eip" {
...
+
+  count = var.asg_terminate_lifecycle_hook_create ? 1 : 0
```
kayman-mk, line 513: question: Should we really make this configurable? As far as I understand this PR (and also from my daily work with this module) it fixes a bug. Thus it should always be enabled to make sure that there are no dangling instances. @npalm

modules/terminate-workers/variables.tf
```diff
@@ -0,0 +1,71 @@
...
+  description = "The name of the Auto Scaling Group to attach to."
+  type        = string
+}
+
+variable "asg_arn" {
+  description = "The ARN of the Auto Scaling Group to attach to."
+  type        = string
+}
+
+variable "create" {
```
kayman-mk, line 19: question: I am not 100% sure how to use this variable. What is the use case not to create the resources? This results in a module which does not the work it was written for, right? Wouldn't I get the same result in not creating the whole module as every resource has a `count = var.create ? 1 :0`?

modules/terminate-workers/iam.tf
```diff
@@ -0,0 +1,103 @@
...
+      identifiers = ["lambda.amazonaws.com"]
+      type        = "Service"
+    }
+  }
+}
+
+resource "aws_iam_role" "lambda" {
+  count = var.create ? 1 : 0
+
+  name                  = var.name_iam_objects == "" ? var.name : var.name_iam_objects
```
kayman-mk, line 24: question: According to the variable description `name_iam_objects` it sounds to me that the name of this and all other IAM objects should be something like `name = "${var.name_iam_objects]-suffix"`

modules/terminate-workers/lambda/lambda_function.py
```diff
@@ -0,0 +1,120 @@
...
+
+    print(json.dumps({
+        "Level": "info",
+        "Message": f"Searching for children of GitLab runner instance {args['parent']}"
+    }))
+
+    ec2_instances = client.describe_instances(Filters=[
+            {
+                "Name": "instance-state-name",
+                "Values": ['running'],
```
kayman-mk, line 25: question: Is this sufficient? What if a new instance is being created? We should kill it too.

modules/terminate-workers/lambda/lambda_function.py
```diff
@@ -0,0 +1,120 @@
...
+                            "Level": "info",
+                            "InstanceId": instance['InstanceId'],
+                            "Name": _name,
+                            "LaunchTime": str(instance['LaunchTime']),
+                            "Message": f"{instance['InstanceId']} will be terminated because its parent is terminating."
+                        }))
+                        _terminate_list.append(instance['InstanceId'])
+                    else:
+                        try:
+                            # Handle other workers without a parent that are still running.
```
kayman-mk, line 55: praise: Good job to handle these instances too!

modules/terminate-workers/iam.tf
```diff
@@ -0,0 +1,103 @@
...
+
+  # Permit the function to terminate instances with the 'gitlab-runner-parent-id'
+  # tag.
+  statement {
+    sid = "GitLabRunnerLifecycleTerminateWorkers"
+    actions = [
+      "ec2:TerminateInstances"
+    ]
+    resources = ["*"]
+    condition {
```
kayman-mk, line 59: praise: Showing the concept of least privileges. Nice!

modules/terminate-workers/versions.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_version = ">= 0.13"
+  required_providers {
+    aws   = ">= 2.46"
+    local = ">= 1.4"
```
kayman-mk, line 5: question: Is this provider still in use?

modules/terminate-workers/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+This example shows using the parent module with the lifecycle hook enable.
+
+Note the `asg_terminate_lifecycle_hook_*` variables:
+
+```terraform
+module "runner" {
+  source  = "npalm/gitlab-runner/aws"
+
+  aws_region                    = "us-west-2"
```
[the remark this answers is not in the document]
joshbeard: Resolved: https://github.com/npalm/terraform-aws-gitlab-runner/pull/392/commits/28a7488c63089fdd888fba61fd08036886a3e973#diff-93dca4429ba0587d39551de9d4aa0822dfd70511e62075c5949c6c5d5e07cf68R52

modules/terminate-workers/variables.tf
```diff
@@ -0,0 +1,71 @@
+# ----------------------------------------------------------------------------
+# Terminate Workers - Input Variables
+# ----------------------------------------------------------------------------
+variable "name" {
```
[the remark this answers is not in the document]
joshbeard: I've added this in 28a7488c63089fdd888fba61fd08036886a3e973

modules/terminate-workers/cloudwatch.tf
```diff
@@ -0,0 +1,39 @@
+# ----------------------------------------------------------------------------
+# Terminate Workers - CloudWatch/EventBridge Resources
+#
+# This deploys an event rule and target for triggering the provided Lambda
+# function from the ASG lifecycle hook.
+# ----------------------------------------------------------------------------
+resource "aws_cloudwatch_event_rule" "terminate_instances" {
+  count = var.create ? 1 : 0
+
+  name        = var.name
```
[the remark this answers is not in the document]
joshbeard: Added in 28a7488c63089fdd888fba61fd08036886a3e973

modules/terminate-workers/main.tf
```diff
@@ -0,0 +1,68 @@
...
+resource "aws_lambda_function" "terminate_runner_workers" {
+  count = var.create ? 1 : 0
+
+  architectures    = ["x86_64"]
+  description      = "Lifecycle hook for terminating GitLab runner worker instances"
+  filename         = data.archive_file.terminate_runner_workers_lambda.output_path
+  source_code_hash = data.archive_file.terminate_runner_workers_lambda.output_base64sha256
+  function_name    = var.name
+  handler          = "lambda_function.handler"
+  memory_size      = 128
```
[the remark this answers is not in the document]
joshbeard: I added a variable for this in https://github.com/npalm/terraform-aws-gitlab-runner/pull/392/commits/a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/variables.tf
```diff
@@ -0,0 +1,71 @@
...
+  default     = true
+}
+
+variable "tags" {
+  description = "Map of tags to apply to resources."
+  type        = map(any)
+  default     = {}
+}
+
+variable "local_existing_package" {
```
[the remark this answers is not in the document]
joshbeard: Removed in a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/variables.tf
```diff
@@ -0,0 +1,71 @@
...
+  description = "If a path is specified, this will use an existing ZIP file built externally for the Lambda function rather than attempt to build and package it in the module."
+  type        = string
+  default     = null
+}
+
+variable "source_path" {
```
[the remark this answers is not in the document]
joshbeard: Removed in a290efa4934c4a43c3f14143be6f21d5ba1ef689

main.tf
joshbeard: I replaced 'workers' with 'instances' in a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/variables.tf
joshbeard: You're right - this is useless in the submodule. I've removed it in a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/iam.tf
joshbeard: Fixed the IAM resource names to be consistent and use the module's "environment" variable in a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/lambda/lambda_function.py
joshbeard: I added "pending" to this list in a290efa4934c4a43c3f14143be6f21d5ba1ef689

modules/terminate-workers/versions.tf
joshbeard: Nope, removed in a290efa4934c4a43c3f14143be6f21d5ba1ef689

joshbeard:
@npalm @kayman-mk I've finally gotten back around to this and have made updates based on your most recent feedback in commit a290efa4934c4a43c3f14143be6f21d5ba1ef689.

I noticed the `README` is updated with some unrelated changes from the pre-commit hook. Not sure why. I left it as a separate commit for now to make it easier to adjust if you want me to remove the unrelated changes (which still seem to be accurate?) That's visible specifically in ab8a2b79af35a0dff4741b0c2d2ef2cb5afc5cc3

I'll rebase this after follow-up reviews and potential tweaks.

npalm:
Great work, added to my backlog.

npalm:
Will check in the weekend

kayman-mk:
@joshbeard  @npalm I did a quick check in my environment.
- Deploying the module: Success
- Killing the runner manually: The executor was killed by the Lambda function. Success
- Didn't find any suspicious log entries in the Lambda log.
- Executors have a new tag: `gitlab-runner-parent-id`

What I noticed:
- Lambda timeout is 3s. This might be too short as the duration of the call was 2s.
- I have to enable the new function via `asg_terminate_lifecycle_hook_create`. This doesn't feel handy. I think it is a base functionality which should always be enabled to be sure not to waste money.
- Log messages are JSON formatted and have a structured field "Level". Really appreciated. Makes log analysis easy.

joshbeard:
> @kayman-mk I also will do a check. Would suggest to make the lambda memory as well the time out configurable with a reasonable default

@npalm I've made the lambda memory size and the timeout configurable: https://github.com/npalm/terraform-aws-gitlab-runner/pull/392/files#diff-05b5a57c136b6ff596500bcbfdcff145ef6cddea2a0e86d184d9daa9a65a288e

main.tf
npalm: Yes I would suggest to make it configurable, but with default enabled. We don't know who is using the module but some orgs are very restrictive for lambda, for example enforcing using lambda's in a custom VPC. By making the the feature optiona (opt-out) we do not block anyone to use the module

joshbeard: Pushed an update to enable this by default

joshbeard:
Rebased against "develop"

GertVil:
Really looking forward to having this merged, because this was one of the "issues" I documented before we could look into using this module at work. Do you have any idea when this can be merged @npalm? (Thank you for the awesome work!)

npalm:
> Really looking forward to having this merged, because this was one of the "issues" I documented before we could look into using this module at work. Do you have any idea when this can be merged @npalm? (Thank you for the awesome work!)

@GertVil I will try to check the PR tonight. Can you provide a short direction how to test the lambda?

GertVil:
> > Really looking forward to having this merged, because this was one of the "issues" I documented before we could look into using this module at work. Do you have any idea when this can be merged @npalm? (Thank you for the awesome work!)
> 
> @GertVil I will try to check the PR tonight. Can you provide a short direction how to test the lambda?

I'm afraid that @joshbeard will probably be in a better position to provide you with that information since he created all this so he probably has a scenario that he has used before.

variables.tf
```diff
@@ -734,3 +734,33 @@ variable "subnet_ids_gitlab_runner" {
...
+
+variable "asg_terminate_lifecycle_lambda_memory_size" {
+  description = "The memory size in MB to allocate to the terminate-instances Lambda function."
+  type        = number
+  default     = 128
+}
+
+variable "asg_terminate_lifecycle_lambda_timeout" {
+  description = "Amount of time the terminate-instances Lambda Function has to run in seconds."
+  default     = 10
```
npalm, line 764: Would it not be better to set the time out a bit higher? 30 for example? The lambda was running for me 2 to 3 seconds, but only 1 instance to kill.

modules/terminate-instances/cloudwatch.tf
```diff
@@ -0,0 +1,33 @@
...
+  "detail": {
+    "AutoScalingGroupName": ["${var.asg_name}"]
+  }
+}
+EOF
+}
+
+resource "aws_cloudwatch_event_target" "terminate_instances" {
+  rule      = aws_cloudwatch_event_rule.terminate_instances.name
+  target_id = "TriggerTerminateLambda"
```
npalm, line 24: According the docs this should be a unique identifier, if not set a random will be created. So I suggest either to remove it or add the environment name to make it unique.
```suggestion
  target_id = "${var.environment}-TriggerTerminateLambda"
```

npalm (changes requested):
@joshbeard great job! works like a charm! Only one small change request

variables.tf
joshbeard: I increased the default to 30

joshbeard:
> @joshbeard great job! works like a charm! Only one small change request, see inline. Can you also run a rebase please.

@npalm Adjustments made and rebased against 'develop'

npalm:
Sorry, missed the update. Will merge the PR tomorrow. If you have time for a rebase, please. Othewrise I will fix it locally!

joshbeard:
Rebased again. The README changed a bit since my last rebase, as I'm sure you're aware. I added the resources to the "resources" section, but that may get overwritten with the current pre-commit config.

npalm:
@joshbeard thx, checking. Just a qucik tip for a next time. Fore reviewing it helps if you are not squashing, only rebase. Now I have to go over the full PR again. No worries we get this merged!

npalm (approved):
@joshbeard All good, thanks!!!

joshbeard:
> @joshbeard thx, checking. Just a qucik tip for a next time. Fore reviewing it helps if you are not squashing, only rebase. Now I have to go over the full PR again. No worries we get this merged! 

I'm sorry about that. Thanks for the review.

joshbeard:
> @joshbeard All good, thanks!!!

Great! Thanks for your work on this, the review, feedback, and including this in the project!

Issue #393, opened 2021-10-17
Backport features from forks


kayman-mk:
Just a list of forks which implemented some features which were not backported (but check this carefully. I had a rough look only.)

Create issues/PR out of the commits from the following forks and close this ticket then.

https://github.com/mbluemer/terraform-aws-gitlab-runner 

- [x] https://github.com/libri-gmbh/terraform-aws-gitlab-runner
- [x] https://github.com/illagrenan/terraform-aws-gitlab-runner
- [x] https://github.com/erback/terraform-aws-gitlab-runner
- [x] https://github.com/DNXLabs/terraform-aws-gitlab-runner

kayman-mk:
[engine-registry-mirror](https://github.com/illagrenan/terraform-aws-gitlab-runner/commit/b36e500b35ec0e49ae07ab00e1359ad96433bff9)

[shared cache](https://github.com/erback/terraform-aws-gitlab-runner/commit/3100e3735731d77880375126c540cb87f109a0a7)

[mount /builds](https://github.com/DNXLabs/terraform-aws-gitlab-runner/commit/7bf10797e32a32bd553bb31258c4553c21d28081)

kayman-mk:
Checked all forks on Oct 15, 2021

PR #400, opened 2021-10-30
feat!: Add variable for Docker registry mirror

### LINKED ISSUE #270: Support for Docker Hub registry mirror for rate-limited pulls
While running some automation jobs in bulk on my Gitlab projects, I got hit today by the infamous new [Docker Hub Rate Limiting](https://www.docker.com/increase-rate-limits)...

```
Waiting for services to be up and running...
Pulling docker image docker:latest ...
ERROR: Preparation failed: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit (docker.go:142:0s)
``` 

There are some solutions to this problem:
* **Authenticate to docker hub**. Anonymously you get 100 container image requests per 6 hours, with an authenticated account you get 200 container image requests per 6 hours. 
* **Pay for Docker Hub**. There's a [Pro account](https://www.docker.com/pricing) already for only $5/month. Requires you to authenticate. For a company, this is the easiest solution and makes the most sense to pay for such a critical service. In my case, it's a personal project for fun so worth looking for other (cheaper) ways. 
* **Cache on GitLab container registry**: tag and push the docker hub image you use (e.g. `php:8`) to your Gitlab repository container registry to "cache", and simply pull from that one to decrease pulls from docker hub. Bit tedious... 
* Use the free [**Gitlab Dependency Proxy**](https://docs.gitlab.com/ee/user/packages/dependency_proxy/) 
  * Requires you to add a `docker login` statement and to prefix every docker image with `$CI_DEPENDENCY_PROXY_GROUP_IMAGE_PREFIX`. Currently, I tried this but get a 500 error back on `docker pull`...
* **Self-Hosted Docker Registry As Pull-Through Cache**:  [Docker Hub Registry Mirror](https://docs.gitlab.com/runner/configuration/speed_up_job_execution.html#docker-hub-registry-mirror) and let the runner use this as a registry as a pull-through cache. A similar thing can be done with JFrog Artifactory/Container Registry. Because distribution uses HEAD manifest requests to check if cached content has changed, it will not count towards the rate limit. Note that initial requests for Hub images that are not yet in cache will be counted.
* Google and AWS have both launched a solution: https://cloud.google.com/container-registry/docs/pulling-cached-images and https://aws.amazon.com/blogs/containers/advice-for-customers-dealing-with-docker-hub-rate-limits-and-a-coming-soon-announcement/ and https://aws.amazon.com/blogs/aws/amazon-ecr-public-a-new-public-container-registry/

Does anyone have experience with any of these solutions? Did anyone try the self-hosted Docker registry mirror? Would it make sense to spin it up in this repo, or provide config options in this repo to add it?

kayman-mk:
## Description

As [Docker](https://docs.docker.com/docker-hub/download-rate-limit/) introduced rate limiting we should set a Docker registry mirror. This was possible via `docker_machine_options`. I think it makes sense to make this explicit using a dedicated variable.

Solves a part of #393 
Closes #270 

## Migrations required

Yes, if the option `engine-registry-mirror` is used within the `docker_machine_options`. If so move the mirror URL to the new variable `runners_docker_registry_mirror`

## Verification

Checked all combinations of `runners_docker_registry_mirror` and `docker_machine_options`.

## Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,7 @@ terraform destroy
 | <a name="input_runners_gitlab_url"></a> [runners\_gitlab\_url](#input\_runners\_gitlab\_url) | URL of the GitLab instance to connect to. | `string` | n/a | yes |
 | <a name="input_runners_helper_image"></a> [runners\_helper\_image](#input\_runners\_helper\_image) | Overrides the default helper image used to clone repos and upload artifacts, will be used in the runner config.toml | `string` | `""` | no |
 | <a name="input_runners_iam_instance_profile_name"></a> [runners\_iam\_instance\_profile\_name](#input\_runners\_iam\_instance\_profile\_name) | IAM instance profile name of the runners, will be used in the runner config.toml | `string` | `""` | no |
+| <a name="input_runners_docker_registry_mirror"></a> [runners\_docker\_registry\_mirror](#input\_runners\_docker\_registry\_mirror) | The docker registry mirror to use to avoid rate limiting by hub.docker.com | `string` | `""` | no |
 | <a name="input_runners_idle_count"></a> [runners\_idle\_count](#input\_runners\_idle\_count) | Idle count of the runners, will be used in the runner config.toml. | `number` | `0` | no |
 | <a name="input_runners_idle_time"></a> [runners\_idle\_time](#input\_runners\_idle\_time) | Idle time of the runners, will be used in the runner config.toml. | `number` | `600` | no |
 | <a name="input_runners_image"></a> [runners\_image](#input\_runners\_image) | Image to run builds, will be used in the runner config.toml | `string` | `"docker:18.03.1-ce"` | no |
--- a/locals.tf
+++ b/locals.tf
@@ -2,10 +2,10 @@ locals {
   // Convert list to a string separated and prepend by a comma
   docker_machine_options_string = format(
     ",%s",
-    join(",", formatlist("%q", concat(var.docker_machine_options, [local.runners_docker_registry_mirror_option]))),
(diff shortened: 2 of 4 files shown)


npalm:
@kayman-mk is this PR still work in progress? If yes can you set it in the draft mode?

kayman-mk:
@npalm Verification still missing. It shows Draft here?!

npalm (approved):
LGTM, still in draft?

kayman-mk:
Yes, unfortunately. The code looks good to me, but is not working. Leaving the registry mirror empty also removes the all docker machine options from the `config.toml`. And I have no idea, why?!

npalm:
OK will try to check also, do you have an example config?

kayman-mk:
Sure. I use one of my other runners and manipulate the config as follows:
* `docker_machine_options = ["a=b", "b=c]`
* `runners_docker_registry_mirror` remains unset

The docker_machine_options will not appear in the config file.

kayman-mk:
Just tested with 
* `docker_machine_options` unset
* `runners_docker_registry_mirror = "https://xxx/"`

No docker_machine_options in config.toml

kayman-mk:
@npalm Found it. Checked all 4 settings for the 2 variables.

npalm (approved):
LGTM,

npalm:
Need only to run a test asap

npalm:
Tested

- [ ] default example
- [ ] default example with mirror to gcr

npalm:
> Tested
> 
> * [x]  default example
> * [x]  default example with mirror to gcr

Checked on the agent instance, config looks proper. Not good way to test the mirror was used at all.

PR #401, opened 2021-10-31
Remove runner agents if provider assumed a role

### LINKED ISSUE #371: Unable to remove runner when assuming role
Currently, when using an [assume role configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume-role) on the AWS provider, this module cannot unregister the runner due to the remove-runner.sh script not assuming the same role assumed by the AWS provider. This leads to the AWS CLI call using the wrong role to try to pull the token value from the SSM parameter store.

I think a potential solution would be to use the [ssm parameter data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssm_parameter) to read the value using the role assumed by the provider, and then pass the token to the script. This would be done instead of having the script use the AWS CLI to pull the value from the SSM parameter store itself.

Since the script uses the AWS CLI, the current workaround is to configure a shared credentials file in ~/.aws/credentials and have an ~/.aws/config file with a configuration set to assume the role using the credentials from the shared credentials file like so:

```
# ~/.aws/credentials
[default]
aws_access_key_id=<your key ID>
aws_secret_access_key=<your secret key>

# ~/.aws/config
[default]
role_arn=<ARN of role assumed by the AWS provider>
source_profile=default
role_session_name=<session name you want to use>
region=<same AWS region the AWS provider is configured to use>
```

kayman-mk:
## Description

If the AWS provider assumes a role, it is not possible to use the AWS CLI to interact with the cloud resources as the CLI does not assume the role automatically. The removal of the runner agents from Gitlab fails.

This PR moves the whole script to remove the runner agent into the Terraform code and does not use the AWS CLI.

Closes #371 

## Migrations required

No

## Verification

Still pending.

## Documentation

### DIFF
--- a/bin/remove-runner.sh
+++ b/bin/remove-runner.sh
@@ -1,5 +1,0 @@
-#!/usr/bin/env sh
-set -e
-
-TOKEN=$(aws ssm get-parameters --name $3 --with-decryption --region $1 | jq -r ".Parameters | .[0] | .Value")
-curl -sS --request DELETE "${2}/api/v4/runners" --form "token=${TOKEN}"
--- a/main.tf
+++ b/main.tf
@@ -21,13 +21,20 @@ resource "aws_ssm_parameter" "runner_registration_token" {
   }
 }
 
+# to read the current token for the null_resource. aws_ssm_parameter.runner_registration_token.value is never updated!
+data "aws_ssm_parameter" "current_runner_registration_token" {
+  depends_on = [aws_ssm_parameter.runner_registration_token]
+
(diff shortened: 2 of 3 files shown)


npalm (commented):
LGTM, still in draft?

npalm (approved):

kayman-mk:
This fixes the issue mentioned in #371, but the `null_resource.triggers.runner_registration_token` contains `null` as the `aws_ssm_parameter.runner_registration_token` is set by the runner during the registration process.

npalm (approved):

Issue #409, opened 2021-11-23
Cache policy attached to runner instead of docker machine instance


hntrmrrs:
https://github.com/npalm/terraform-aws-gitlab-runner/blob/3ea16b04df7129c813f46cdcbf42aba3da90e65a/main.tf#L406

If docker machine runners are used, I believe this should be attached to the docker machine instance role instead.

kayman-mk:
@hntrmrrs I think this issue is done as explained in #410. Could you please close it?

PR #410, opened 2021-11-23
Attach cache bucket policy to docker machine instance role

hntrmrrs:
## Description

The cache bucket policy is attached to the runner instance role by default but should be attached to the docker machine instance role instead.

## Migrations required

NO

## Verification

Verified that this works with a simple docker+machine runner configuration and cache creation.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -403,7 +403,7 @@ resource "aws_iam_role_policy_attachment" "user_defined_policies" {
 ################################################################################
 resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance" {
   count      = var.cache_bucket["create"] || lookup(var.cache_bucket, "policy", "") != "" ? 1 : 0
-  role       = aws_iam_role.instance.name
+  role       = aws_iam_role.docker_machine.name
   policy_arn = local.bucket_policy
 }
 


hntrmrrs:
Fixes #409

npalm:
Sorry, had not time to check the PR's early. Add this to my todo list. Thanks for your contribution!!!

npalm (commented):
LGTM, need to test

npalm:
Sorry for the late response. Just tested and the cache is not working anymore. This workflow is one of the basic flows I use for testing. 


```yaml
cache:
  key: "$CI_BUILD_REF"
  untracked: true

job:
  stage: build

  image: npalm/cowsay

  script:
    - cowsay  -f ghostbusters building  "$CI_BUILD_NAME" @ stage "$CI_BUILD_STAGE" > cow.txt

  tags:
     - docker_spot_runner

verify:
  stage: verify

  image: npalm/cowsay

  script:
    - cat cow.txt
```

See here also the build log
test


The change looks logic, and I remember from years ago that I have tried the same. But for some reason that is not clear to me it seems the agent needs to have access to cache.

npalm (changes requested):
Did you had time to check my comment?

kayman-mk:
I checked the documentation: "If the S3 cache adapter is configured to use an IAM instance profile, the adapter uses the profile attached to the GitLab Runner machine." (see https://docs.gitlab.com/runner/configuration/advanced-configuration.html).

So from my point of view everything is ok and no fix is needed.

kayman-mk:
I suppose we should place a comment in the code to ducument this.

hntrmrrs:
Ok thanks for checking the documentation. I'll leave this alone then since it should work as-is.

npalm:
@hntrmrrs @kayman-mk I checked the docs as well. And I really have no clue why I am not getting it working. 

1. I deplolyed the module (development branch). And run a test with s3 cach, push and pull. Works fine. See example above
2. Upgrade to this PR and got a forbidden once I try to write the file to cache in the workflow. 

In develop the policy is attached the agent instance, in this branch to the docker machine instance. Which sounds logical and correct according the docs. Any clue what I miss?

kayman-mk:
@npalm I think it is said in the documentation. Yes, it sounds strange but matches what you have discovered. Using the S3 cache in the Gitlab Runner config.toml and use an instance profile --> attach the cache policy to the Gitlab Runner and not the executors.

Sounds really weired, but I guess the cache is transferred from the executor to the Runner and then stored in S3. I found this in the documentation mentioned above:

> The cache mechanism uses pre-signed URLs to upload and download cache. URLs are signed by GitLab Runner on its own instance. It does not matter if the job’s script (including the cache upload/download script) are executed on local or external machines. For example, shell or docker executors run their scripts on the same machine where the GitLab Runner process is running.

npalm:
Thanks, but that explains why the role should be attached the runner agent. Thanks for reaching out and taking the time for making the PR.

Issue #411, opened 2021-11-24
Can not assign subnets during execution "terraform plan"


kostyanius:
Error: Variables not allowed
│
│   on <value for var.subnet_ids_gitlab_runner> line 1:
│   (source code not available)
│
│ Variables may not be used here.
╵
╷
│ Error: No value for required variable
│
│   on variables.tf line 27:
│   27: variable "subnet_ids_gitlab_runner" {
│
│ The root module input variable "subnet_ids_gitlab_runner" is not set, and has no default value. Use a -var or
│ -var-file command line argument to provide a value for this variable.

if to hardcode the value in the variables.tf than error changes to the following:

│ Error: Variables not allowed
│
│   on <value for var.subnet_ids_gitlab_runner> line 1:
│   (source code not available)
│
│ Variables may not be used here.
╵
╷
│ Error: No value for required variable
│
│   on variables.tf line 28:
│   28: variable "subnet_ids_gitlab_runner" {
│
│ The root module input variable "subnet_ids_gitlab_runner" is not set, and has no default value. Use a -var or
│ -var-file command line argument to provide a value for this variable.
╵

npalm:
Are you using one of the examples?

kostyanius:
i tried to do it from scratch and also checked examples. both cases do not work fro me. so looks like there is an issue with subnets.

kayman-mk:
Do you have step by step instructions to reproduce the problem? Like: `git clone something && run something`? Hard to say what's going wrong without seeing your console.

npalm:
@kostyanius still having the issues. Our deployment relies also on predefined subnets and we having no issues.

kostyanius:
Hello folks. 
Thanks for your updates. I will record the video tomorrow for you all to show the issue i am facing with during this setup. Sorry for delay in my response. Was busy.

kostyanius:
![image](https://user-images.githubusercontent.com/9063908/145838110-e9bcab72-230d-41ae-a12e-710fd022da56.png)

kostyanius:
Hey
Any updates on this?
Thanks.

kayman-mk:
@kostyanius Variable names can be improved, yes.

```hcl
variable "subnet_id_runners" {
  description = "List of subnets used for hosting the gitlab-runners."
  type        = string
}

variable "subnet_ids_gitlab_runner" {
  description = "Subnet used for hosting the GitLab runner."
  type        = list(string)
}
```

[...]

PR #420, opened 2021-12-28
refactor: Harmonize subnet variables

### LINKED ISSUE #419: Harmonize the subnet_id* variables
#411 raised the question of renaming the `subnet_ids_gitlab_runner` and `subnet_id_runners`. Very hard to understand what is configured here.

Proposal:
Introduce a new variable `subnet_id`. Use this subnet for the runner and the executors. I do not see any advantage of having the runner placed in another subnet as the executors. Both should be in the same subnet and the module should be deployed per availability zone to get the resilience needed.

Mark the other variables deprecated. But still use them if `subnet_id` is not filled.

```hcl
variable "subnet_id" {
  description = "Subnet used for the runner and the executors."
  type        = string
}

variable "subnet_id_runners" {
  description = "Deprecated. Use subnet_id instead. List of subnets used for hosting the gitlab-runners."
  type        = string
}

variable "subnet_ids_gitlab_runner" {
  description = "Deprecated. Use subnet_id instead. Subnet used for hosting the GitLab runner."
  type        = list(string)
}
```

kayman-mk:
## Description

Deprecates `subnet_ids_gitlab_runner` and `subnet_id_runners` and introduces `subnet_id` to make the module more readable as explained in #411 and #419. Thus it it no longer possible to place the runner and the executors in different subnets. For a resilient configuration deploy the module once per availability zone.

Closes #419

## Migrations required

No, but migration should be done.

Replace `subnet_ids_gitlab_runner` and `subnet_id_runners` by `subnet_id`. Please notice that it is no longer possible to place the runner and executors in different subnet. For a resilient configuration, I suggest to deploy the module once per availability zone.

## Verification

Checked the following:
- config.toml checked for subnet value using `subnet_id`. Success. Correct subnet used.
- Check that the Runner is created in correct subnet using `subnet_id`. Success.
- Remove `subnet_id` and use the old variables `subnet_ids_gitlab_runner` and `subnet_id_runners` again. `terraform plan` does not show any differences. Success.

## Documentation

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Niek Palm
+Copyright (c) 2018 The terraform-docs Authors.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 The terraform-docs Authors.
+Copyright (c) 2017 Niek Palm
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/variables.tf
+++ b/variables.tf
@@ -22,6 +22,7 @@ variable "vpc_id" {
 variable "subnet_id" {
   description = "Subnet used for the runner and executors."
   type        = string
+  default     = ""
 }
 
 variable "subnet_id_runners" {
--- a/variables.tf
+++ b/variables.tf
@@ -25,16 +25,6 @@ variable "subnet_id" {
   default     = ""
 }
 
-variable "subnet_id_runners" {
-  description = "Deprecated! Use subnet_id instead. List of subnets used for hosting the gitlab-runners."
-  type        = string
-}
-
-variable "subnet_ids_gitlab_runner" {
-  description = "Deprecated! Use subnet_id instead. Subnet used for hosting the GitLab runner."
-  type        = list(string)
-}
-
 variable "extra_security_group_ids_runner_agent" {
   description = "Optional IDs of extra security groups to apply to the runner agent. This will not apply to the runners spun up when using the docker+machine executor, which is the default."
   type        = list(string)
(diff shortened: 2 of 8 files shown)


kayman-mk:
Missing:
- verification
- rewrite examples

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Niek Palm
+Copyright (c) 2018 The terraform-docs Authors.
```
npalm, line 3: Can you revert this change, was it changed by terraform-docs?

variables.tf
```diff
@@ -19,13 +19,19 @@ variable "vpc_id" {
   type        = string
 }
 
+variable "subnet_id" {
```
npalm, line 22: I agree to you that the current names of varialbes can be conflicting.

- `subnet_id_runners` is intended as the subnet in which the docker machine instancesa are created
- `subnet_ids_gitlab_runner` is the subnet to host the instance with gitlab agent.

As you mention in the PR this drops the option to deploy to different subnets.  I have no clue if some is doing that. But don't see really a case for that.

variables.tf
```diff
@@ -19,13 +19,19 @@ variable "vpc_id" {
...
+  description = "Subnet used for the runner and executors."
+  type        = string
+  default     = ""
```
npalm, line 25: Setting the default value to `""` won't work with the conditon you use like `var.subnet_id ? .....` The conditon should be evaluate to a boolean expression.

npalm (changes requested):
@kayman-mk thanks for making the module simpler and better. I have added some reviews comment. 

- First, I was not able to run an upgrade of the default example. The condition is not valid.
- As you see I have alwasy some doubt about dropping flexibility. Since it cn impact the current ussers. But I agree there is no use of having 2 subnets.

npalm (changes requested):
@kayman-mk thanks for making the module simpler and better. I have added some reviews comment. 

- First, I was not able to run an upgrade of the default example. The condition is not valid.
- As you see I have alwasy some doubt about dropping flexibility. Since it cn impact the current ussers. But I agree there is no use of having 2 subnets.

kayman-mk:
Just some simple changes to the code how it could work and to discuss it. I will fix everything else to get a running version.

LICENSE
kayman-mk: Oh yes, of course. I didn't do that manually.

kayman-mk:
Verification still open and to be done as soon as I am back to work. ;-)

kayman-mk:
Verified

npalm:
Great will check!

npalm (approved):
@kayman-mk thanks for making the module simpler, removing complexity one of the habits we should practice more!

PR #425, opened 2022-01-08
feat: Request new Runner token if invalid

### LINKED ISSUE #230: gitlab-runner.tpl does not check the validity of secure_parameter_store_runner_token
I've run into some issues where the token stored in SSM is invalid in testing. This seems to mostly arise from the runner not triggering remove_gitlab_registration on termination of the instance from the AWS Console. This has been tested with the "Runner - Docker runner" setup 
https://github.com/npalm/terraform-aws-gitlab-runner/tree/develop/examples/runner-docker 

Instead of assuming the token in valid check if the token in valid on startup using the following endpoint
https://docs.gitlab.com/ee/api/runners.html#verify-authentication-for-a-registered-runner

I'd purpose something like 
```
token=$(aws ssm get-parameters --names "${secure_parameter_store_runner_token_key}" --with-decryption --region "${secure_parameter_store_region}" | jq -r ".Parameters | .[0] | .Value")

valid_token=true
if [[ `echo $token` != "null" ]]
then
  valid_token_response=$(curl -s -o /dev/null -w "%{http_code}" --request POST -L "${runners_gitlab_url}/api/v4/runners/verify" \
      --form "token=${token}" ) \
  if [[ `echo $valid_token_response` != "200" ]]
  then
       valid_token=false
  fi
fi

if [[ `echo ${runners_token}` == "__REPLACED_BY_USER_DATA__" && `echo $token` == "null" ]] || [`echo $valid_token` == "false"]
then
  token=$(curl --request POST -L "${runners_gitlab_url}/api/v4/runners" \
    --form "token=${gitlab_runner_registration_token}" \
    --form "tag_list=${gitlab_runner_tag_list}" \
    --form "description=${giltab_runner_description}" \
    --form "locked=${gitlab_runner_locked_to_project}" \
    --form "run_untagged=${gitlab_runner_run_untagged}" \
    --form "maximum_timeout=${gitlab_runner_maximum_timeout}" \
    --form "access_level=${gitlab_runner_access_level}" \
    | jq -r .token)
  aws ssm put-parameter --overwrite --type SecureString  --name "${secure_parameter_store_runner_token_key}" --value="$token" --region "${secure_parameter_store_region}"
fi
```

kayman-mk:
## Description

This PR is based on the work of patramsey as described in #230.

Gitlab Runner checks the validity of the token stored in SSM. In case the token is invalid a new one is requested.

Closes #230 

## Migrations required

No

## Verification

The following verifications were done:
- Change the registration token manually, kill the Runner manually. I expect to get a new Runner registered at Gitlab. Success.
- Delete the Gitlab Runner from Gitlab and kill the Runner EC2. Expect to get a newly registered Runner. Success.
- Check Cloudwatch logs to see the result of the verification call: 403. Success.

## Documentation

### DIFF
--- a/template/gitlab-runner.tpl
+++ b/template/gitlab-runner.tpl
@@ -64,7 +64,7 @@ token=$(aws ssm get-parameters --names "${secure_parameter_store_runner_token_ke
 valid_token=true
 if [[ `echo $token` != "null" ]]
 then
-  valid_token_response=$(curl -s -o /dev/null -w "%{response_code}" --request POST -L "${runners_gitlab_url}/api/v4/runners/verify" --form "token=${token}" )
+  valid_token_response=$(curl -s -o /dev/null -w "%%{response_code}" --request POST -L "${runners_gitlab_url}/api/v4/runners/verify" --form "token=${token}" )
   [[ `echo $valid_token_response` != "200" ]] && valid_token=false
 fi
 
--- a/template/gitlab-runner.tpl
+++ b/template/gitlab-runner.tpl
@@ -64,7 +64,7 @@ token=$(aws ssm get-parameters --names "${secure_parameter_store_runner_token_ke
 valid_token=true
 if [[ `echo $token` != "null" ]]
 then
-  valid_token_response=$(curl -s -o /dev/null -w "%%{response_code}" --request POST -L "${runners_gitlab_url}/api/v4/runners/verify" --form "token=${token}" )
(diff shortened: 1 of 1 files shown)


kayman-mk:
Verification done

npalm (approved):
@kayman-mk LGTM tested as well. Strnage I never had an issue with this. Furthermore, by checking this part we should also move the registration token to SSM as well. Once checking the code, I also asked myself why we not just acking a registratin token eacht time and only store the registratrion token in ssm. The only reason we could have at that time is to avoid new runners got registred.

PR #441, opened 2022-02-11
fix: remove the null resource

### LINKED ISSUE #426: Upgrading the module causes problem with null_resource (Missing map element)
After updating to 4.37 and applying the plan, an error ocurred. Details below.

As far as I can see, it has nothing to do with the 4.37 upgrade as there were no changes to the null_resource. I think it is due to the fact that I didn't upgrade to 4.36 (and maybe the upgrade to 4.35 was also missing) and the problem was introduced in an earlier version.

However I do not fully understand what's going on here. Seems that the old resource had different keys in the trigger map (yes, I remember the change). And in consequence the null_resource is destroyed and executes the code which is not working as the new parameter is not in the map.
Seems that changing the null_resource ends with destroy/create this resource and then the trigger is fired. This was not intended.

In this case I fixed it with deleting the objects from the state file.

Plan:
```
  # module.gitlab_runner.module.gitlab_terraform_runner["subnet-12345"].null_resource.remove_runner must be replaced
-/+ resource "null_resource" "remove_runner" {
      ~ id       = "596161117442277782" -> (known after apply)
      ~ triggers = { # forces replacement
          + "runner_registration_token"               = (sensitive)
          - "script"                                  = ".terraform/modules/gitlab_runner.gitlab_terraform_runner/bin/remove-runner.sh" -> null
          - "secure_parameter_store_runner_token_key" = "prod-gitlab-ci-c-token" -> null
            # (2 unchanged elements hidden)
        }
    }
```

State:
```
$ terraform state show module.gitlab_runner.module.gitlab_terraform_runner[\"subnet-12345\"].null_resource.remove_runner
# module.gitlab_runner.module.gitlab_terraform_runner["subnet-12345"].null_resource.remove_runner:
resource "null_resource" "remove_runner" {
    id       = "596161117442277782"
    triggers = {
        "aws_region"                              = "eu-central-1"
        "runners_gitlab_url"                      = "https://gitlab.xxx.cloud"
        "script"                                  = ".terraform/modules/gitlab_runner.gitlab_terraform_runner/bin/remove-runner.sh"
        "secure_parameter_store_runner_token_key" = "prod-gitlab-ci-c-token"
    }
}
```

Result:
```
│ Error: Missing map element
│
│   on .terraform\modules\gitlab_runner.gitlab_terraform_runner\main.tf line 43, in resource "null_resource" "remove_runner":
│   43:     command    = "curl -sS --request DELETE \"${self.triggers.runners_gitlab_url}/api/v4/runners\" --form \"token=${self.triggers.runner_registration_token}\""
│     ├────────────────
│     │ self.triggers is map of string with 4 elements
│
│ This map does not have an element with the key "runner_registration_token".
```

kayman-mk:
## Description

Removes the `null_resource` to remove the Gitlab runner from the Gitlab instance. Sometimes the `null_resource` was triggered due to changes in Terraform code and your runner was removed from Gitlab causing outages of your pipelines. Unfortunately there is no Terraform resource available indicating that a module was completely removed.

In case you want to get rid of this module, don't forget to remove the runner from your Gitlab instance manually.

Closes #426 

## Migrations required

The `null_resource` is completely removed. Please make sure to delete it from the statefile manually to avoid the runners being removed from the Gitlab instance.

- `terraform state list` will show you everything in the state, you need to look for the "remove_runner" null_resource.
- `terraform state rm module.<runners-module-name>.null_resource.remove_runner`

## Verification

No verifications done as the resource was removed.

## Documentation


npalm (approved):

npalm:
@kayman-mk looks good! can you rebase, and resolve the conflicts please?

kayman-mk:
rebased

npalm (approved):

PR #446, opened 2022-02-23
feat: Skip runner download and install if already installed

Swsystem-ATuin:
## Description

Currently the gitlab runner user data will try to download/run install script then install. This change will allow alternative installation via userdata_pre_install

## Migrations required

NO

## Verification

Tested without issue using default install and when already installed. 

## Documentation

pre-commit run fine, no change to terraform inputs or docs required.

### DIFF
--- a/template/gitlab-runner.tpl
+++ b/template/gitlab-runner.tpl
@@ -28,8 +28,11 @@ then
   yum install amazon-ecr-credential-helper -y
 fi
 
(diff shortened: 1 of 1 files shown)


npalm:
@Swsystem-ATuin thanks aadded to my queue for PR's this weekend

npalm (approved):
@Swsystem-ATuin thx!

Issue #457, opened 2022-03-02
Docker-machine download URL appears to be inaccessible


JulianCBC:
It appears that GitLab has made the "generic" `https://gitlab-docker-machine-downloads.s3.amazonaws.com/VERSION/docker-machine` URL inaccessible, so only the "fully qualified" URLs with a system type and architecture work.

```
$ curl https://gitlab-docker-machine-downloads.s3.amazonaws.com/v0.16.2-gitlab.12/docker-machine
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>ENYDP0R4QVSFP0QC</RequestId><HostId>iJ7zVyel7aLWz4qEE0lMPpwXbItiAmiud1LkRzwbzts36AqLG1rvINDixGl4Qs2SvrJ1NL7cIYc=</HostId></Error>
```

```
$ curl https://gitlab-docker-machine-downloads.s3.amazonaws.com/v0.16.2-gitlab.12/docker-machine-Linux-x86_64 | file -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 59.8M  100 59.8M    0     0  1957k      0  0:00:31  0:00:31 --:--:-- 2982k
/dev/stdin: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=H8fu0C_uLjkoi-o2ao1v/Greq4F4710ptsLoOLyoB/6Hmp0aU-D8xXDMQeVDPz/wbFjEQ0zwCgSg31oPilA, not stripped
```

I think the simplest solution here would be to just switch the default over to specifying a version instead of a full URL as that appears to do the right thing, and will follow the architecture used on the coordinator EC2 instance without any further configuration.

ryancausey:
Can confirm this issue prevents the runner from starting properly, so I've had to lock to version 4.40.0 for now, as well as manually setting `gitlab_runner_version` to something other than the default of "14.8.0" due to #456.

Issue #456, opened 2022-03-02
Runner release 14.8.0 has been removed by GitLab due to regression


william00179:
Hi!

GitLab has removed the 14.8.0 release due to a regression.

https://gitlab.com/gitlab-org/gitlab-runner/-/releases/v14.8.0

This seems to be the default version that the module will try to install. I suggest this could be bumped to 14.8.1 that includes the fix for the regression, or 14.8.2 which is the latest.

PR #458, opened 2022-03-03
Update / fix dependency versions

JulianCBC:
## Description

Both the current default GitLab Runner version as well as the default docker-machine download URL are broken:

* Version 14.8.0 of GitLab Runner had a major regression and was removed from all download sources #456
* GitLab has stopped making the "generic" download URL (ending in `/docker-machine`) available #457

Fix these by:

1. Updating the `docker-machine` download URL generation to explicitly use GitLab's version as the official Docker version is dead and buried
2. Switch from an explicit `docker-machine` download URL to specifying the latest GitLab fork's version
3. Upgrade to GitLab Runner version 14.8.2

## Migrations required

* If users are explicitly setting the download URL or GitLab runner version, this won't change anything
* If users are using the defaults, this will upgrade these components which shouldn't cause issues
* **However** if users are expecting to use the official version of `docker-machine` by specifying a _version_ and blanking the download URL, this will migrate them to the GitLab fork and probably break as the version number scheme has changed.

## Verification

This has been verified on my production environment, which is a slightly tweaked version of the "runner-docker" example.

## Documentation

Documentation has been updated using `terraform-docs`.


npalm (approved):
@JulianCBC All good! thanks a lot!

Issue #464, opened 2022-04-01
An argument named "architectures" is not expected here


hiago-miguel:
Hey,

With aws provider version `3.59.0`, and version 4.41.1 of this module, I got the following error:

```» terraform validate                                                                                                                                                                  ╷
│ Error: Unsupported argument
│ 
│   on .terraform/modules/runner/modules/terminate-instances/main.tf line 20, in resource "aws_lambda_function" "terminate_runner_instances":
│   20:   architectures    = ["x86_64"]
│ 
│ An argument named "architectures" is not expected here.
╵
--------------------------------------
```

The argument `architectures` in the aws_lambda_function resource was released in version 3.61.0:
- https://github.com/hashicorp/terraform-provider-aws/releases/tag/v3.61.0

Since the submodule `terminate-instances` uses this argument, 

```
resource "aws_lambda_function" "terminate_runner_instances" {
  architectures    = ["x86_64"]
  description      = "Lifecycle hook for terminating GitLab runner instances"
...
```
The module should require aws provider ~> 3.61.

Or maybe, it could just remove the `architecures` argument:

> [architectures](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#architectures) - (Optional) Instruction set architecture for your Lambda function. Valid values are ["x86_64"] and ["arm64"]. Default is ["x86_64"]. Removing this attribute, function's architecture stay the same.

---

After updating my provider version to 3.61.0 the issue was fixed.

npalm:
Thanks, will check it

npalm:
Justed tested the module with the 4.0 provder (#460) with the default example. Not able to reproduce the rror you mention.

hiago-miguel:
Yes, the issue only happens when using aws provider < 3.61.
And I have opened this issue because the module's minimum requirement for the aws provider is 3.35.
- https://github.com/npalm/terraform-aws-gitlab-runner/tree/4.41.1#requirements

cmavromichalis:
I also ran into this using main branch, tag 4.41.1 (https://github.com/npalm/terraform-aws-gitlab-runner/releases/tag/4.41.1). Bumping to hashicorp/aws 3.75.1 fixed it.

npalm:
Merged solution #468

PR #460, opened 2022-03-07
feat: Support AWS provider 4

npalm:
See also #44

PR #468, opened 2022-04-12
fix: terminate lambda: don't specify architecture

joshbeard:
## Description

https://github.com/npalm/terraform-aws-gitlab-runner/issues/464 reports that the `architectures` argument was added in the AWS provider 3.61.0 and this module specifies 3.35 as a minimum.

The `architectures` argument is optional and defaults to 'x86_64' anyway.

This removes it so that it's compatible with the older AWS provider.)

## Migrations required

NO

## Verification

Verified using the _runner-default_ example

## Documentation

N/A

### DIFF
--- a/modules/terminate-instances/main.tf
+++ b/modules/terminate-instances/main.tf
@@ -17,7 +17,6 @@ data "archive_file" "terminate_runner_instances_lambda" {
 }
 
 resource "aws_lambda_function" "terminate_runner_instances" {
-  architectures    = ["x86_64"]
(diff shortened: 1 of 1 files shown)


joshbeard:
This may not be wanted/needed, considering https://github.com/npalm/terraform-aws-gitlab-runner/pull/460 and the future AWS provider 4 support in this module.

npalm (approved):
@joshbeard thanks, sorry for the delay

joshbeard:
> @joshbeard thanks, sorry for the delay

No problem!

Issue #471, opened 2022-04-25
Add AuthenticationType to S3 runner cache configuration


gravitybacklight:
The upcoming 15.0 release of the gitlab runner will expect the AuthenticationType setting to be defined.
https://about.gitlab.com/blog/2022/04/18/gitlab-releases-15-breaking-changes/
https://gitlab.com/gitlab-org/gitlab-runner/-/issues/28171

It looks as though this would always be set to IAM if used in this module.
The setting was introduced in 14.5.

gravitybacklight:
one method to achieve this in https://github.com/npalm/terraform-aws-gitlab-runner/pull/472
could alternatively just add AuthenticationType = "IAM" to the config, but I am unaware of the impact on earlier runner versions.

npalm:
Combining work of #742 and #483 to solve the issue.

PR #472, opened 2022-04-25
471 add AuthenticationType

gravitybacklight:
## Description

One method to support adding AuthenticationType = "IAM" whilst ensuring no changes to configuration for earlier versions.
Only adds it in if the major version of the runner is 15 or higher.

## Migrations required

NO

## Verification

Tested with versions 14 & 15 to confirm the user-data only sees this change if version 15.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -131,6 +131,7 @@ locals {
       bucket_name                       = local.bucket_name
       shared_cache                      = var.cache_shared
       sentry_dsn                        = var.sentry_dsn
+      gitlab_runner_major_version       = tonumber(split(".", var.gitlab_runner_version)[0])
     }
   )
 }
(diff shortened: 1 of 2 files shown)


kayman-mk:
Not 100% clear what problem is solved in this PR. What is the problem with the default value?

According to the docs
>   In GitLab 14.4 and later, set to iam or access-key. Default is access-key if ServerAddress, AccessKey, and SecretKey are all provided. Defaults to iam if ServerAddress, AccessKey, or SecretKey are missing.

kayman-mk:
Got it, explained in #471 It becomes a mandatory parameter.

gravitybacklight:
@kayman-mk yeah, I was unsure if there was any use case of an access key for this module, but if not, it can just be hardcoded as `AuthenticationType = "IAM"`

kayman-mk:
I think the Runners and Executors usw instance profiles only. So hardcoding the value looks good to me.

kayman-mk:
You are right. It's the better solution.

npalm (approved):
@gravitybacklight thanks for your work, sorry for the far too late reply. Will merge the change in a temporay branch and apply some extra fixes..

PR #483, opened 2022-05-16
Add AuthenticationType for [runners.cache.s3]

zooomey:
## Description
This PR adds the AuthenticationType for [runners.cache.s3]. Not specifying an AuthenticationType is a [deprecated](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/28171) feature and breaking change for GitLab 15.0.

The AuthenticationType can either be `iam` or `credentials`, and this PR allows the user to specify which AuthenticationType they use as a variable. Previously, this would default to iam credentials, so I maintained the default authentication method in `variables.tf`

## Migrations required

No

## Verification

When the runner was manually updated, the change only impacts the actual launch template of the child runners. All future spawns will be using the correct auth type.

### DIFF
--- a/template/runner-config.tpl
+++ b/template/runner-config.tpl
@@ -33,6 +33,7 @@ log_format = "json"
     Type = "s3"
     Shared = ${shared_cache}
     [runners.cache.s3]
+      AuthenticationType = "${auth_type}"
       ServerAddress = "s3.amazonaws.com"
       BucketName = "${bucket_name}"
       BucketLocation = "${aws_region}"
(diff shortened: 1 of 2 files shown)


kayman-mk:
Please check #472 I guess this change is already in the pipeline.

kayman-mk:
#471 is the issue belonging to #472

zooomey:
@kayman-mk I saw #471 and #472, however,`"IAM"` should be `"iam"`. When my team did this locally no child runners came up until the `AuthenticationType` was changed to `iam`. Since we are not on GitLab 15.0 this is backwards compatible.

Also, this does not hardcode any AuthType into our template, however, it will default to `iam` if the developer chooses to not specify.

zooomey:
Apologies, I fatfingered the `auth_type` type and wrote "strings" instead of "string". Should be fixed now

PR #497, opened 2022-05-25
fix: use IMDSv2 url when setting PARENT_INSTANCE_ID

### LINKED ISSUE #496: config.toml has empty "gitlab-runner-parent-id"
Hi, 

I've found that, for what ever reason, the user-data configuration doesn't seem to get the instance ID of the gitlab runner agent.

This variable
https://github.com/npalm/terraform-aws-gitlab-runner/blob/9d39138b7f526ebe7132c5d0bb1e4a50cea50626/template/gitlab-runner.tpl#L3

contains `gitlab-runner-parent-id,`.

This is also what the log says: 

```
++ echo gitlab-runner-parent-id,
+ sed -i.bak s/__PARENT_TAG__/gitlab-runner-parent-id,/g /etc/gitlab-runner/config.toml
```

The curl call in the log to the instance-id endpoint show this:

```
++ curl http://169.254.169.254/latest/meta-data/instance-id
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

+ PARENT_INSTANCE_ID=
+ PARENT_TAG=gitlab-runner-parent-id,
```

All other calls to metadata or getting the accesstokens work. It's just this one.

Funnily enough, a few lines above in the log, there is a call to `http://169.254.169.254/latest/dynamic/instance-identity/document` which return the instance-id. So maybe the metadata isn't available at this point in the EC2 creation process?

I can reproduce this issue with every deployment.

snoord:
## Description

This PR updates the URL used for the `PARENT_INSTANCE_ID` variable so that it now points to the AWS IMDSv2 endpoint. 

Using the old URL caused an empty `gitlab-runner-parent-id` tag to be added to all runner instances created by agents, which broke the Lambda function for terminating stray runner instances.

Closes #496 

## Migrations required

NO

## Verification

Successfully tested in private AWS account.

## Documentation

### DIFF
--- a/template/gitlab-runner.tpl
+++ b/template/gitlab-runner.tpl
@@ -1,5 +1,10 @@
+# Install jq if not exists
+if ! [ -x "$(command -v jq)" ]; then
+  yum install jq -y
+fi
+
 # Provide the parent instance id in the spawned runner tags
-PARENT_INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id);
+PARENT_INSTANCE_ID=$(curl -s -H "X-aws-ec2-metadata-token: $token" http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .instanceId)
 PARENT_TAG="gitlab-runner-parent-id,$${PARENT_INSTANCE_ID}"
 
 mkdir -p /etc/gitlab-runner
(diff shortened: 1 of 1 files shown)


npalm (approved):
LGTM

AlexEndris:
@snoord  Thanks!

Issue #322, opened 2021-05-17
Spot instances over multiple AWS Availability Zones


guillemtp:
Hi!

Is there any option to configure the same GitLab Runner agent to use multiple AWS AZ on spot instances?

Thanks

npalm:
Yes, just provide a subnet for each zone, instances should be started (guess random) over your zones.

guillemtp:
In which module attribute should be defined?

The attribute `subnet_id_runners` is only for one subnet.

pierlucg-xs:
I don't think this is possible at the moment: https://docs.gitlab.com/runner/configuration/runner_autoscale_aws/#:~:text=If%20not%20specified%2C%20the%20availability%20zone%20is%20a%2C%20it%20needs%20to%20be%20set%20to%20the%20same%20availability%20zone%20as%20the%20specified%20subnet%2C%20for%20example%20when%20the%20zone%20is%20eu-west-1b%20it%20has%20to%20be%20amazonec2-zone%3Db

kayman-mk:
As a workaround you could spin up agents in different AZs. It's almost the same.

npalm:
Also having 2 deployments

PR #500, opened 2022-05-27
support executors in multiple AZ

martin-paulus:
## Description

Support multiple profiles in the gitlab runner config (multiple [[runners]] sections), based on the subnets provided.

We required this because we saw our pipelines stalling due to unavailability of our preferred EC2 Spot machine type in our default availability zone. After extending the running confguration with more [[runners]] sections the launch failures were followed by successful launches in another availability zone.

## Migrations required

YES - The value of var.subnet_id needs to be input (as a list) at var.executor_subnets to maintain backwards compatibilty.

## Verification

(verification was based on a modified 4.18.0 release of the module, this exact PR is yet to be verified)

## Documentation

Terraform inputs in the documentation were updated via [terraform-docs](https://github.com/terraform-docs/terraform-docs).


variables.tf
```diff
@@ -846,18 +850,6 @@ variable "docker_machine_egress_rules" {
...
-  type        = string
-  default     = ""
-}
-
-variable "subnet_ids_gitlab_runner" {
-  description = "Deprecated! Use subnet_id instead. Subnet used for hosting the GitLab runner."
-  type        = list(string)
-  default     = []
-}
-
```
kayman-mk, line 860: issue: In general a good idea to remove theses outdated variables. But this creates extra work for all users to update their module call. With the next major release we will restructure and clean the inputs. Please add them again.

kayman-mk (changes requested):

kayman-mk:
@martin-paulus Sorry, have to open a discussion here:

@npalm Is it really useful that an agent in AZ A handles executors in different AZ? Due to stability you deploy the module at least twice, choosing different AZs for the agent. So the problem described in this PR would be solved and no change is needed.

Having multiple `[[runners]]` sections is indeed useful to spawn different machine types. See #249 

There was also a short discussion in #322 

I tend to close this PR. It seems to be better to deploy the module in different AZs for a better stability. The queue time of the pipeline will also drop and it reacts faster.

kayman-mk:
@npalm What do you think about my comment from Jan 19th?

PR #503, opened 2022-06-08
Make statement IDs unique

JulianCBC:
## Description

Somehow the lambda permissions in the "terminate-instances" module have identical statement IDs and somehow this was fine until I tried to migrate my runner infrastructure to a different Terraform configuration.

## Migrations required

No migrations required.

## Verification

This has been verified on my production environment, which is a slightly tweaked version of the "runner-docker" example.


npalm (approved):
@JulianCBC thx LGTM

PR #510, opened 2022-06-19
fix: always add policy to maintain SSM parameters

### LINKED ISSUE #509: enable_manage_gitlab_token = false does not store runner token
Hello,

when starting the runner with `enable_manage_gitlab_token = false` the script is supposed to set the token in SSM. However there is an error due to a missing permission.

A workaround is to put the token value manually into SSM.

This is the error from the EC2 instance user-data:
```
aws ssm put-parameter --overwrite --type SecureString --name gitlab-runner-manager-runner-token --value=XXX --region eu-central-1

An error occurred (AccessDeniedException) when calling the PutParameter operation: User: arn:aws:sts::XXX:assumed-role/gitlab-runner-manager-instance/i-XXX is not authorized to perform: ssm:PutParameter on resource: arn:aws:ssm:eu-central-1:XXX:parameter/gitlab-runner-manager-runner-token because no identity-based policy allows the ssm:PutParameter action
```

kayman-mk:
## Description

According to #509 the Runner always maintains a SSM parameter. Thus it needs the policy to be able to to so.

This MR removes the possibility to steer whether or not to add the policy to allow the Runner to `ssm:GetParameter` and `ssm:PutParameter`. The variable `enable_manage_gitlab_token` is marked as deprecated and no longer evaluated as the last usage was removed.

Fixes #509 

## Migrations required

No

## Verification

Deployed this version and executed a job successfully.


JulianCBC:
Looks good to me, however we should probably make it explicit that the variable now does nothing, not just "deprecated".

Maybe something like:

```
Ignored and deprecated: This variable used to disable management of the SSM property containing the runner token, meaning that the user-data script still created it, but it wasn't managed by terraform. The runner token is now always stored in a terraform managed SSM property.
```

kayman-mk:
Good point. Fixed that.

npalm (approved):
@kayman-mk  looks good! Thx

rewb0rn:
As discussed in #509 this MR should probably not be merged. Setting `enable_manage_gitlab_token` to false allows to reuse the previous runner registration when `terraform apply` is executed (contrary to registering a new runner in the gitlab server when the value is set to true).

kayman-mk:
@npalm Can we decide for one way here? Either a) remove the whole functionality as it is not working at the moment and according to your comment in #509 this belongs to an old scenario which is superfluous now or b) fix it.

I propose to remove this function too as the automatic registration process is less error prone.

npalm:
Not close to a computer this weekend. The automatic registration is the default, right? There where we request gitlab for a token. The other option is legacy for day zero as far I remember. 

So suggest we remove the function. I can have a quick look Monday evening

kayman-mk:
Automatic registration is the default, true.

npalm:
So we can close this one, since we will remove automatic registration. OK?

kayman-mk:
Not 100% sure. This is a quick one solving the problem. The mentioned correct solution could take some time to be completed and could also be a breaking change.

But closing this one would be ok from my side.

npalm:
@kayman-mk should I test this PR? Can do this next week? Feel free to merge the PR.

npalm (changes requested):
Testing the PR at the moment, have some troubles. So for that reason mark it as `request for changes`

npalm:
For some reason the agent is not staring after applying changes. So blocking the merge for the momet. Sorry

kayman-mk:
Branch updated

kayman-mk:
Let me test this as well. Should be running as nothing has been removed from the policies.

kayman-mk:
@npalm This branch is running in my environment. Deployed a runner and executed one job successfully.

npalm:
> @npalm This branch is running in my environment. Deployed a runner and executed one job successfully.

Will run a test asap. Assume it is now fine. Thanks!

kayman-mk:
@npalm Seems to be ready now.

main.tf
```diff
@@ -513,7 +511,7 @@ resource "aws_iam_role_policy_attachment" "ssm" {
   count = var.enable_manage_gitlab_token ? 1 : 0
```
npalm, line 511: Here the variable is still used. Evoluation will now cause an error, since the value is always null.

npalm (approved):

PR #511, opened 2022-06-19
feat!: allow to set all docker options for the Executor

kayman-mk:
## Description

Adds a new variable `runners_docker_options` which holds all values for the `[runners.docker]` section and makes the single variables
- `runners_image`
- `runners_privileged`
- `runners_disable_cache`
- `runners_additional_volumes`
- `runners_shm_size`
- `runners_docker_runtime`
- `runners_helper_image`
- `runners_pull_policy`

obsolete.

## Migrations required

Yes, as the minimum Terraform version is 1.3.0 to support optional block variables with defaults.

A migration script is provided to restructure the variables. See `/migrations/migrate-to-7-0-0.sh`. Attention Mac users: The script will not work out of the box as the `sed` implementation is different. Use a Docker container with Alpine or Ubuntu to run the script.

```hcl
module "gitlab_ci_runner" {
  ...
  runners_docker_options {
    # set whatever is necessary
  }
```

## Verification

- [x] Use current configuration and ensure that the `config.toml` remains unchanged
- [x] Set all new block variables and ensure that the `config.toml` is valid (use `gitlab-runner verify)
- [x] Check that the default settings with Terraform 1.3 work as expected
- [x] Verify all docker settings against the documentation to ensure correct names

The runner starts in both cases and is available in Gitlab. No example tested but used our active configuration at Hapag-Lloyd.


JulianCBC:
Could we get rid of the `runners_enable_docker_options` variable and maintain backwards compatibility by making the default just use the old variables?

We'd have to note on the old variables that if `runners_docker_options` is set, then they're ignored, but that'd simplify a lot of the code here.

kayman-mk:
Sure. But setting a default for the new block is problematic. I have to know whether the new block is used or the old variables.

We could set a default `null` and then decide: `if var.runners_docker_options == null then use old variables else use new block variables`.

As soon as the old variables are removed we could set the default for the block.

JulianCBC:
Other than enabling the experiment which I can't see actually being used right now, this looks good to me!

I can't help but wonder if you could use the template for both the original and new variable sets, and do all the formatting of the volumes string in the template, but those are pretty minor quibbles and the code I'm complaining about _is_ deprecated.

kayman-mk:
Yeah, the experiment is not my favorite too. But it made it into Terraform 1.3.0 (released soon). Default values will also be possible.

kayman-mk:
Good point though. Reworked that section. Looks much better now.

npalm:
Would be great when we can move to Terrfaform 1.3. The optiional in object will made the code a lot cleaner on my places. Just checked, there are two alphaversions out. So it is still in an early state but at least we see the progress towards 1.3

kayman-mk:
Still waiting for the Terraform 1.3 release. No idea when this will happen.

kayman-mk:
Terraform 1.3 was released some days ago.

npalm:
@kayman-mk please can you rebase the PR with develop?

npalm:
Did a quick check with default example, but got the error `ERROR: Preparation failed: unsupported pull_policy config: ""`

kayman-mk:
@npalm Any idea why the Terraform workflows do not run in this branch? I merged the current `main` already.

kayman-mk:
Tested with my converted configuration. Jobs are still processed.

versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1"
+  required_version = ">= 1.3"
```
npalm, line 2: Should we mark this a breaking?

npalm: Anser to myself, yes will be a breaking change. Also variables renamed

migrations/migrate-to-7-0-0.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
npalm, line 1: 🚀

npalm (commented):
@kayman-mk nice one, but hard to chck if it is complete. Really a lot of variables. Will run some test.

npalm (commented):
Ran a basic test, works as expected. Maybe an idea we create a test repo with wome workflows.

migrations/migrate-to-7-0-0.sh
kayman-mk: - I will remove the `pipefail` to support more shells
- Mac was not in scope. Do you see a chance to support the script on a Mac as well? Otherwise Mac users have to start a Docker container with Alpine, ... and run the script there.

kayman-mk:
As soon as we are satisfied, we should release it with #723, I think. This saves one major release and it fits perfectly into that PR.

kayman-mk:
I changed the PR description and mention the problem on Macs. Possible solution is to run the script in a Docker container.

Let me do a test on my machine again to make sure that everything is running fine. I already doublechecked the variable names with the handbook. They look ok.

kayman-mk:
I verified the config.toml by adding all variables manually according to the template file. Names and types were correct. I am good to go.

tmeijn:
Hi, I just became aware of this Pull Request and I got a couple of remarks/suggestions.

While this makes the interface for the user a lot simpler and gives more capabilities/settings to the user, I'm a bit worried about the complexity this introduces into the module itself.

Instead of parsing each variable from the object separately we can pass the key+value pair as is to the `runners_docker_options.tfpl` since we control both the key and value type with the object definition. Within the template itself we have to accommodate for only three variable types:

* Arrays/lists
* numbers
* strings

I tried to do this in my own PR, but unfortunately I didn't get any feedback from you all yet. But it seems you are trying to do the same. See my draft PR #711 and specifically the logic in https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/711/files#diff-fd9e1a870839d390694bebe28698eddedcb6a97b768004316ce5c27b78811181

Something like this should probably work (but not tested):

```template
%{~ for key, value in config ~}
      %{ if contains([LIST_OF_VALUES_TO_CONVERT_TO_TOML_ARRAY], key) }${key} = [${replace(format("\"%s\"", join("\",\"", value)), "/\"{2,}/", "\"")}]%{ else }
      ${key} = ${ can(tonumber(value)) ? value : format("\"%s\"", value)}
      %{~ endif ~}
%{~ endfor ~}
```

This should drastically reduce the parsing that's happening now in `locals.tf` and should also remove most if not all `if` statements for each key in the template file.

kayman-mk:
Sounds good to me. I saw this problem during development but didn't find a solution. I will try yours. It should make the configuration a lot easier.

tmeijn:
@kayman-mk actual we can and should even keep the logic out of the template. We can use `for ... in ...` to convert the values to TOML compatible values. I've found a generic "function" we could use to basically map HCL values to TOML values. See below for a quick and dirty experiment.

"function":

```hcl
runners_machine_autoscaling = templatefile("${path.module}/template/runners_machine_autoscaling.tpl", {
    runners_machine_autoscaling = [for entry in var.runners_machine_autoscaling :
      { for option, value in entry : option => (
        can(distinct(value)) ?                                                     # distinct can only be used on lists. In combination with 'can' we determine whether the current value is a list or other.
        "[${replace(format("\"%s\"", join("\",\"", value)), "/\"{2,}/", "\"")}]" : # Wrap and convert to TOML array
        can(tonumber(value)) ? value : format("\"%s\"", value)                     # Leftover value is either a number or string. If string, we wrap with qoutes.
        )

    }]
    }
  )
```

template:

```template
%{~ for config in runners_machine_autoscaling ~}
    [[runners.machine.autoscaling]]
    %{~ for key, value in config ~}
      ${key} = ${value}
      %{~ endfor ~}
%{~ endfor ~}

```

input:

```hcl
  runners_machine_autoscaling = [
    {
      an_array_with_str = ["one", "two", "three"]
      a_number = 1
      a_string = "hello, World!"
    }
  ]
```

output:

image

kayman-mk:
@npalm @tmeijn Looks much better now, right? Thanks for the hint.

template/runners_docker_options.tftpl
```diff
@@ -0,0 +1,4 @@
+  [runners.docker]
+%{ for key, value in options ~}
+    ${key} = ${jsonencode(value)}
```
tmeijn, line 3: @kayman-mk hahaha! 😅 man, look at me doing all fancy for in and you solving it with one single function 🤦🏾

kayman-mk: Saw it somewhere else. Quite neat!

template/runner-config.tpl
```diff
@@ -17,22 +17,17 @@ listen_address = "${prometheus_listen_address}"
...
-    disable_cache = ${runners_disable_cache}
-    volumes = ["/cache"${runners_additional_volumes}]
-    extra_hosts = ${jsonencode(runners_extra_hosts)}
-    shm_size = ${runners_shm_size}
-    pull_policy = ${runners_pull_policies}
-    runtime = "${runners_docker_runtime}"
-    helper_image = "${runners_helper_image}"
-    ${runners_docker_services}
+
+  ${runners_docker_options}
```
tmeijn, line 21: **suggestion:** since the template is now only four lines, WDYT about just putting it inline here? Then we can remove the template, the local variable and directly feed it as a variable to the `runner-config.tpl`.

JulianCBC: It's probably good practice to keep it as a template in case it grows again.

tmeijn:
@kayman-mk did something go wrong here? This PR shows it's now empty and merging `refactor-variables` in #711 shows changes that I think were in this PR.

kayman-mk:
Oh no, what happened here?! I should have a local copy.

kayman-mk:
Ok, back on track.

template/runner-config.tpl
kayman-mk: I stay with the template for now.

npalm (approved):

tmeijn:
FYI: `services_security_opt` made it into GitLab Runner 15.10 today... https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/3760. It's not documented yet though...

kayman-mk:
@npalm Conflicts solved and merged into the upcoming major release branch.

PR #514, opened 2022-06-21
fix: improve cache settings

kayman-mk:
## Description

A caching period of one day seems to be too short as default. This has been adjusted to 7 days to allow some caching.

## Migrations required

No

## Verification

No verification done. No relevant parts changed.

### DIFF
--- a/modules/cache/variables.tf
+++ b/modules/cache/variables.tf
@@ -54,7 +54,7 @@ variable "cache_lifecycle_clear" {
 variable "cache_lifecycle_prefix" {
   description = "Object key prefix identifying one or more objects to which the clean up rule applies."
   type        = string
-  default     = "project/"
+  default     = "runner/"
 }
 
 variable "arn_format" {
--- a/modules/cache/variables.tf
+++ b/modules/cache/variables.tf
@@ -30,7 +30,7 @@ variable "cache_bucket_versioning" {
 variable "cache_expiration_days" {
   description = "Number of days before cache objects expires."
   type        = number
-  default     = 1
+  default     = 7
 }
 
 variable "tags" {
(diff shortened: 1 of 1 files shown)


kayman-mk:
Shall the caching period reduced to 5 days may be? Usually you work with versioned artifacts so you could cache 100 days but it doesn't make sense to have all old (and hopefully unused) versions in the cache.

modules/cache/variables.tf
```diff
@@ -54,7 +54,7 @@ variable "cache_lifecycle_clear" {
 variable "cache_lifecycle_prefix" {
   description = "Object key prefix identifying one or more objects to which the clean up rule applies."
   type        = string
-  default     = "runner/"
+  default     = "project/"
```
npalm, line 57: The root directory is runner in the bucket, for that reason I had set the default to runner. What is the issue here?

npalm: Checking my current deployment, to see if it is not cleared. Will update in a few days.

kayman-mk: Strange. The root directory is `project/` in my case and I am using the default values only.

The lifecycle rule says `runner/` thus no clean up is done here.

I will also doublecheck my code.

npalm: Just doubled checked, deployed the default example (examples dir) and cache files are stored in runners/<id>/project. So the root dir in my case is runner

kayman-mk: Strange. I will revert this change.

kayman-mk:
@npalm What do you think about increasing the default caching period? PR to be closed or to be merged?

npalm:
I personally not a fan of long caching, since it can introduce a false sens of things keep working due to caching. For that reason I have set the caching short. Of course it the writer of the pipeline to think about the impact of caching. So setting the default higher is fine by me as well. In the end it is configurable

kayman-mk:
As it is configurable, I will close this one.

PR #516, opened 2022-07-09
feat: add `clone_url` to `config.toml`

kayman-mk:
## Description

Adds the missing parameter ` clone_url` to the `config.toml` of the runner. In case the default clone URL from the GitLab project is not accessible from the runner's infrastructure (e.g. private DNS), the `clone_url` can be used to specify a different one (e.g. replace the private DNS entry by a public IP address).

Check the [configuration docs for the runner](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runners-section).

## Migrations required

No.

## Verification

- [x] Deploy infrastructure without `clone_url`. Check that the runner can register and `config.toml` shows an empty `clone_url`. Ensure that jobs are still processed as we didn't have a `clone_url` in the config before.
- [x] Deploy infrastructure with `clone_url`. Check that the runner can register and `config.toml` shows the mentioned `clone_url`.

I check this within the Hapag-Lloyd GitLab Runner infrastructure.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -75,7 +75,7 @@ locals {
     {
       aws_region                  = var.aws_region
       gitlab_url                  = var.runners_gitlab_url
-      clone_url                   = var.runners_clone_url
+      gitlab_clone_url            = var.runners_clone_url
       runners_vpc_id              = var.vpc_id
       runners_subnet_id           = length(var.subnet_id) > 0 ? var.subnet_id : var.subnet_id_runners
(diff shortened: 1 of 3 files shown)


npalm (approved):
LGTM, will run later today a quick test. Catching up with the PRs

veshant:
Using this method has an issue with ssl certificates that don't have the clone url.  Is there a way to have the dns resolve to a different ip address instead?  I have tried to append records to the hosts file, but it only does in the container and not the runner machine.

PR #517, opened 2022-07-12
fix: user data tracing flag

ravenolf:
## Description

This commit fixes the tracing flag for bash in the user data. 

Currently if you set the [enable_runner_user_data_trace_log](https://github.com/npalm/terraform-aws-gitlab-runner#input_enable_runner_user_data_trace_log) flag to `false` expecting to disable the tracing in bash it will actually trace the user data script.

The reason is that the flag will be translated to `user_data_trace_log == false`. Once the condition is evaluated in the script you get `if [[ $(echo ${user_data_trace_log}) == false ]]` -> `false == false` which is `true` and ends up enabling `set -x`.

This small change should fix it.

## Migrations required

NO

## Verification

Tested by running the script locally and setting the value in the current version to `true` and saw the opposite result


kayman-mk:
Sounds reasonable to me.

kayman-mk:
@ravenolf I think it would be good to mention this change in the `Migrations required` section. As the guys using this flag have to invert it in order to get the same results.

npalm (commented):
@ravenolf thx, looks good. Would suggest also tw swap the default value https://github.com/npalm/terraform-aws-gitlab-runner/blob/d3912b0666d2fb0eb97bd53c66702153543261d4/variables.tf#L616 to true, so the behavior remains the same

kayman-mk (dismissed):
Added the change @npalm requested.

kayman-mk (approved):

kayman-mk:
#604 fixes the TF Docs problem. Can be ignored.

PR #522, opened 2022-08-01
feat: do not add a the Name tag for docker+machine >= 0.16.2

### LINKED ISSUE #521: Duplicate tag key 'Name' error with newer docker-machine version
When setting `docker_machine_version` to >=0.16.2-gitlab.16, I'm seeing the following error when the runner tries to create docker-machine ec2 instances:
```
{
    "driver": "amazonec2",
    "level": "error",
    "msg": "Error creating machine: Error in driver during machine creation: Error launching instance: InvalidParameterValue: Duplicate tag key 'Name' specified.",
    "name": "runner-zeq53ex4-runner-1658964971-a77712f5",
    "operation": "create",
    "time": "2022-07-27T23:36:15Z"
}
```

This looks to be cause by [this change](https://gitlab.com/gitlab-org/ci-cd/docker-machine/-/merge_requests/68) in gitlab's docker-machine, which is present in version 0.16.2-gitlab.16 and above. Both the [runner config](https://github.com/npalm/terraform-aws-gitlab-runner/blob/develop/main.tf#L92) and [docker-machine](https://gitlab.com/gitlab-org/ci-cd/docker-machine/-/blob/v0.16.2-gitlab.18/drivers/amazonec2/amazonec2.go#L1106) adds a `Name` tag for the instance, which will cause a conflict in the newer docker-machine versions since tags are now being added at instance creation time instead of after instance creation.

I think removing the `Name` tag from https://github.com/npalm/terraform-aws-gitlab-runner/blob/develop/main.tf#L92 and letting docker-machine create the `Name` tag from the MachineName could be a solution.

kayman-mk:
## Description

This PR checks the version of the docker machine and adds the Name tag if it is not done by the driver itself. It also removes the second `Name` tag from the `amazonec2-tags` in the `config.toml`. In consequence it is no longer possible to change the Name of the Runners as the `Name` tag is now added by the docker-machine driver (for versions >= 0.16.2).

`docker_machine_version` >= 0.16.2 automatically adds the Name tag to the EC2 instances. Adding another Name tag here causes the EC2 creation process to fail.

To be able to configure the name, `var.overrides["name_docker_machine_runners"]` is added as prefix to the name of the docker machine which is used as name for the EC2 instance by the driver.

The main logic was taken from https://github.com/hashicorp/terraform/issues/22688#issuecomment-1196805096

Closes #521 

## Migrations required

No.

## Verification

- [x] Verify the version logic manually with a `local.tf` only. Returns `true` if version >= 0.16.2
- [x] Check the `config.toml` for versions >= 0.16.2 with `overrides["name_docker_machine_runners"]`. No Name tag present.
- [x] Check the `config.toml` for versions >= 0.16.2 without `overrides["name_docker_machine_runners"]`. No Name tag present.
- [x] Check the `config.toml` for versions < 0.16.2 with `overrides["name_docker_machine_runners"]`. No Name tag equals the name set in the override.
- [x] Check the `config.toml` for versions < 0.16.2 without `overrides["name_docker_machine_runners"]`. Name tag shows `var.environment + "-docker-machine"`
- [x] Replace the current module with this version and check if tags are changed (using version < 0.16.2). There shouldn't be any change in the resource tags.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -28,12 +28,12 @@ locals {
   )
 
   /* determines if the docker machine executable adds the Name tag automatically (versions >= 0.16.2) */
-  docker_machine_version_used          = split(".", var.docker_machine_version)
+  # make sure to skip pre-release stuff in the semver by ignoring everything after "-"
+  docker_machine_version_used          = split(".", split("-", var.docker_machine_version)[0])
   docker_machine_version_with_name_tag = split(".", "0.16.2")
   docker_machine_version_test = [
     for i, j in reverse(range(length(local.docker_machine_version_used)))
-    # make sure to skip pre-release stuff in the semver by ignoring everything after "-"
-    : signum(split("-", local.docker_machine_version_with_name_tag[i])[0] - split("-", local.docker_machine_version_used[i])[0]) * pow(10, j)
+    : signum(local.docker_machine_version_with_name_tag[i] - local.docker_machine_version_used[i]) * pow(10, j)
   ]
 
   docker_machine_adds_name_tag = signum(sum(local.docker_machine_version_test)) <= 0
--- a/locals.tf
+++ b/locals.tf
@@ -32,6 +32,7 @@ locals {
   docker_machine_version_with_name_tag = split(".", "0.16.2")
   docker_machine_version_test = [
     for i, j in reverse(range(length(local.docker_machine_version_used)))
+    # make sure to skip pre-release stuff in the semver by ignoring everything after "-"
     : signum(split("-", local.docker_machine_version_with_name_tag[i])[0] - split("-", local.docker_machine_version_used[i])[0]) * pow(10, j)
   ]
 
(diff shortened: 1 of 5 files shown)


kayman-mk:
Also see #522 for long-wan-ep's test.

kayman-mk:
Still discussing the `MachineName` option in #521

kayman-mk:
Using the `var.overrides["name_docker_machine_runners"]` as prefix for the EC2 name.

long-wan-ep:
Tested out the prefix for the machine name, it looks good to me. One thing to note is that gitlab-runners will add `runner-<SHORT_TOKEN>` as a prefix to the machine name so the name of the instance will end up being `runner-<SHORT_TOKEN>-<NAME_OVERRIDE>-<MACHINE_IDENTIFIER>, but I think it's ok.

npalm (approved):
LGTM, tested with default example

PR #527, opened 2022-08-08
refactor!: remove usage of `create_cache_bucket`

### LINKED ISSUE #525: Refactor cache module
As far as I can see, all resources in the cache module have a count attribute which is set to 1 if the `create_cache_bucket` equals true.

But why should I set the flag to `false` and call the module? It would do nothing at all and I could simply omit the whole call.

The benefit would be a simpler module:
- no `count` attributes and thus simple resources instead of lists
- outputs could return the plain value instead of reading lists
- easier to understand and maintain

The only caller I know is the runner module. But I think there are only small changes to do if any.

I'd like to refactor this module in a PR. Any thoughts and things to consider?

kayman-mk:
## Description

Marks `create_cache_bucket` deprecated to simplify the cache module. In case no cache should be created it is better to not call this module instead of setting the variable to `false`. In this case the module does nothing.

Closes #525 

## Migrations required

In case `create_cache_bucket` was set to `false`: remove the whole module call as it does nothing. Otherwise you should remove the deprecated variable.

## Verification
- [x] deploy over existing 5.1.0 deployment and check plan for changes regarding the cache
- [x] deploy cache with `create_cache_bucket = false`. Throws a validation error
- [x] deploy cache with `create_cache_bucket = true`. Throws a validation error
- [x] deploy cache. Creates the S3 bucket as usual

### DIFF
--- a/modules/cache/variables.tf
+++ b/modules/cache/variables.tf
@@ -40,7 +40,7 @@ variable "tags" {
 }
 
 variable "create_cache_bucket" {
-  description = "This module is by default included in the runner module. To disable the creation of the bucket this parameter can be disabled."
+  description = "(deprecated) If the cache should not be craeted, remove the whole module call!"
   type        = bool
   default     = true
 }
--- a/modules/cache/variables.tf
+++ b/modules/cache/variables.tf
@@ -42,7 +42,12 @@ variable "tags" {
 variable "create_cache_bucket" {
   description = "(deprecated) If the cache should not be craeted, remove the whole module call!"
   type        = bool
-  default     = true
+  default     = null
+
+  validation {
+    condition     = anytrue([var.create_cache_bucket == null])
+    error_message = "Deprecated, don't call the module when not creating a cache bucket."
+  }
 }
 
 variable "cache_lifecycle_clear" {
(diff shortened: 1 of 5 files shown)


npalm:
Some pre terrafom optional module logic

modules/cache/variables.tf
```diff
@@ -40,7 +40,7 @@ variable "tags" {
 }
 
 variable "create_cache_bucket" {
-  description = "This module is by default included in the runner module. To disable the creation of the bucket this parameter can be disabled."
+  description = "(deprecated) If the cache should not be craeted, remove the whole module call!"
   type        = bool
   default     = true
```
npalm, line 45: With a validation rule like below the user gets some nice feedback if still using the variable.
```suggestion
  default     = null
  
  validation {
    condition     = anytrue([var.create_cache_bucket == null])
    error_message = "Deprecated, don't call the module when not creating a cache bucket."
  }
```

npalm (commented):
LGTM, a small suggestion and need to test

modules/cache/variables.tf
kayman-mk: Very good!

kayman-mk:
Tests done. Looks good.

PR #528, opened 2022-08-13
fix: always add the cache policy

### LINKED ISSUE #298: Problem using the shared cache solution with multiple runners (Invalid count argument)
I'm trying to use a shared bucket (as in the examples) together with multiple runners created via `for_each` (not sure if this issue is also relevant without it but I guess so).

```
module "cache" {
  source  = "npalm/gitlab-runner/aws//modules/cache"
  version = "4.21.0"
  environment = "gitlab-runner"
}

module "runner" {
  source  = "npalm/gitlab-runner/aws"
  version = "4.21.0"

  # module will be created for every key in this map
  for_each = var.runners

  # network settings
  aws_region = var.aws_region
  vpc_id     = var.vpc_id
  subnet_ids_gitlab_runner = [var.subnet_id]
  subnet_id_runners        = var.subnet_id
  metrics_autoscaling      = ["GroupDesiredCapacity", "GroupInServiceCapacity"]

  # instance configuration
  environment              = each.key
  instance_type            = each.value.instance_type

(...)

  gitlab_runner_security_group_ids = [var.security_group_id]

    registration_token = var.registration_token_map[each.value.registration_token_key]
    tag_list           = each.value.tags
    description        = each.value.runner_name
    locked_to_project  = "true"
    run_untagged       = "false"
    maximum_timeout    = "3600"
  }

  cache_shared = "true"

  cache_bucket = {
    create = false
    policy = module.cache.policy_arn
    bucket = module.cache.bucket
  }

  }
```

Applying this configuration results in an error: 
```
Error: Invalid count argument

  on ..\..\..\terraform-aws-gitlab-runner\main.tf line 323, in resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance":
 323:   count      = var.cache_bucket["create"] || local.bucket_policy != "" ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.

```

Applying this config is only possible via:
`terraform plan -target=module.cache` first followed by a second `terraform apply`, because `lookup(var.cache_bucket, "policy", "")` is used inside module which is only available after creating the bucket.

```
resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance" {
  count      = var.cache_bucket["create"] || lookup(var.cache_bucket, "policy", "") != "" ? 1 : 0
  role       = aws_iam_role.instance.name
  policy_arn = local.bucket_policy
}
```

See https://github.com/terraform-aws-modules/terraform-aws-transit-gateway/issues/2
See https://github.com/hashicorp/terraform/issues/4149

```
$ terraform init
$ terraform plan -target module.cache
$ terraform apply -target module.cache
$ terraform plan
$ terraform apply
```

The question here is, can we solve this issue somehow?
Do we really need `count      = var.cache_bucket["create"] || lookup(var.cache_bucket, "policy", "") != "" ? 1 : 0` here? Is there a use case where we do not want to add the policy at all? If not, we could ommit the count and solve the problem entirely and make it easier for users?

@npalm WDYT?

kayman-mk:
## Description

As described in #298 the cache module has to be applied separately before creating the Runners. Terraform complains about

```
Error: Invalid count argument

  on ..\..\..\terraform-aws-gitlab-runner\main.tf line 323, in resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance":
 323:   count      = var.cache_bucket["create"] || local.bucket_policy != "" ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

This PR always adds the policy to get rid of the error message during `apply`. As far as we know there is no use cases deploying this module without a cache. Thus we can always add the cache policy to the role.

Closes #298 

## Migrations required

No

## Verification

- [x] create a runner with a cache created separately (using the cache module). No error is shown during `apply`
- [x] create a runner and let the module create the cache. `apply` shows no error.
- [x] destroy everything without error

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ A few option are provided to access the runner instance:
 
 ### GitLab runner cache
 
-By default the module creates a a cache for the runner in S3. Old objects are automatically removed via a configurable life cycle policy on the bucket.
+By default the module creates a cache for the runner in S3. Old objects are automatically removed via a configurable life cycle policy on the bucket.
 
 Creation of the bucket can be disabled and managed outside this module. A good use case is for sharing the cache across multiple runners. For this purpose the cache is implemented as a sub module. For more details see the [cache module](https://github.com/npalm/terraform-aws-gitlab-runner/tree/develop/cache). An example implementation of this use case can be found in the [runner-public](https://github.com/npalm/terraform-aws-gitlab-runner/tree/__GIT_REF__/examples/runner-public) example.
 
--- a/main.tf
+++ b/main.tf
@@ -399,7 +399,6 @@ resource "aws_iam_role_policy_attachment" "user_defined_policies" {
 ### Policy for the docker machine instance to access cache
 ################################################################################
 resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance" {
-  count      = var.cache_bucket["create"] || lookup(var.cache_bucket, "policy", "") != "" ? 1 : 0
   role       = aws_iam_role.instance.name
   policy_arn = local.bucket_policy
 }


npalm (approved):

Neki:
I think this MR introduces an issue with this configuration:

```
  cache_bucket = {
    "create" : false,
    "bucket" : aws_s3_bucket.build_cache.id
  }
```
(use case: S3 cache created outside of this module, because it's shared with other runner types)

When applying this, I get the error:

```
│ Error: Error attaching policy  to IAM Role spot-runners-instance: InvalidParameter: 1 validation error(s) found.
│ - minimum field size of 20, AttachRolePolicyInput.PolicyArn.
│
│
│   with module.ec2_runner.aws_iam_role_policy_attachment.docker_machine_cache_instance,
│   on .terraform/modules/ec2_runner/main.tf line 392, in resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance":
│  392: resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance" {
│
```

since the `policy_arn` is set to the empty string.

kayman-mk:
Had the same problem an hour ago, but no time to create the issue.

100% correct.

Do you have a use case for a Runner without a cache?

Neki:
We do use a cache, but it's created outside of this module (I don't wish this module to create it for me).

That's because we run several Gitlab runner types (Kubernetes and docker-machine), and they should all share the same cache. So it's simpler to create the S3 resources in a separate module, then pass them to the various modules that deploy the runners.

So
* I don't have a use case for a runner without a cache
* but I have a use case for a runner where the cache is not created by this module 😄

kayman-mk:
@Neki It would be wonderful if you could have a quick look at my PR mentioned above. Just did the the `terraform apply` without creating a cache and adding a policy and it worked like a charme.

PR #530, opened 2022-08-16
fix: access cache module via index

kayman-mk:
## Description

Fixes an error applying the cache module. As a `count` was introduced, we have to use an index to access the module's result.

## Migrations required

No

## Verification

None

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -293,8 +293,8 @@ resource "aws_launch_template" "gitlab_runner_instance" {
 ### Create cache bucket
 ################################################################################
 locals {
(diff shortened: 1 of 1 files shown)


kayman-mk:
@npalm This fixes the error mentioned at https://github.com/npalm/terraform-aws-gitlab-runner/issues/525#issuecomment-1216001717

npalm (approved):

long-wan-ep:
@npalm Would it be possible to have this released soon? 5.2.0 currently can't be deployed without this.

npalm:
Released 5.2.1

PR #533, opened 2022-08-23
fix: add `var.environment` to the name of the S3 cache bucket

### LINKED ISSUE #507: cache: environment is not included as bucket name prefix
The [docs](https://registry.terraform.io/modules/npalm/gitlab-runner/aws/latest/submodules/cache#input_environment) describe the `environment` variable as:

>   A name that identifies the environment, used as prefix and for tagging.

However, `environment` appears to only be used for tags and is not included in the prefix (see https://github.com/npalm/terraform-aws-gitlab-runner/blob/5.0.2/modules/cache/main.tf#L15.).  Am I right in thinking the intent was to include this in the prefix?

kayman-mk:
## Description

Adds the name of the environment to the S3 cache bucket name. As #507 explains, the environment name from `var.environment` is not part of the S3 cache bucket name which can cause problems if the module is deployed multiple times.

Furthermore the documentation states that `var.environment` is used as prefix for the resources which was not the case here.

Closes #507 

## Migrations required

No, but the whole cache bucket is lost and you may experience degraded performance of your runners for a while as the cache is cleared.

## Verification

- [x] apply over the latest deployed version. S3 bucket should be recreated with a name prefixed with the environment name. The new S3 bucket is empty
- [x] apply from scratch and make sure that the Runners are working.


tags.tf
```diff
@@ -24,7 +24,7 @@ locals {
     local.tags,
     var.runner_tags,
     # overwrites the `Name` key from `local.tags`
-    var.overrides["name_docker_machine_runners"] == "" ? { Name = format("%s-docker-machine", var.environment) } : { Name = var.overrides["name_docker_machine_runners"] },
+    var.overrides["name_docker_machine_runners"] == "" ? { Name = substr(format("%s-docker-machine", var.environment), 0, 28) } : { Name = var.overrides["name_docker_machine_runners"] },
```
npalm, line 27: Sorry a bit late, made more or less Yesterday the sam fix. I think 20 was even too long, reduced it to 12 and removed docker-machine. Can you rebase the PR?

kayman-mk: Yeap, fixed.

npalm (approved):
Thanks

PR #540, opened 2022-09-01
fix: do not add the cache access policy if there is none

kayman-mk:
## Description

In case the module does not create the cache and `var.cache_bucket["policy"]` is empty, the module can't be deployed as explained in  https://github.com/npalm/terraform-aws-gitlab-runner/pull/528#issuecomment-1234004003.

This PR makes sure to add the policy only if it exists.

## Migrations required

No.

## Verification

- [x] Deploy the module without creating a cache and setting the `policy`. Module is created without any error.


npalm (approved):
And once again thanks for your contribution!

PR #544, opened 2022-09-12
feat: add runners_pull_policies to support multiple pull policies

gmeligio:
## Description

Support multiple Docker pull policies. The pull_policy parameter allows to specify a list of pull policies. The policies in the list will be attempted in order from left to right until a pull attempt is successful, or the list is exhausted.

https://docs.gitlab.com/runner/executors/docker.html#how-pull-policies-work

## Migrations required

NO - A deprecation notice was added to description. In a future release the variable runners_pull_policy could be removed in favor of runners_pull_policies.

## Verification

I have verified the examples (https://github.com/gmeligio/terraform-aws-gitlab-runner/actions/runs/3040828783):
- Verify examples (1.0.11, runner-default)
- Verify examples (1.0.11, runner-docker)
- Verify examples (1.0.11, runner-multi-region)
- Verify examples (1.0.11, runner-pre-registered)
- Verify examples (1.0.11, runner-public)
- Verify examples (latest, runner-default)
- Verify examples (latest, runner-docker)
- Verify examples (latest, runner-multi-region)
- Verify examples (latest, runner-pre-registered)
- Verify examples (latest, runner-public)


npalm (commented):
PR looks good, but test results in runner is not started. Rebased the PR also against develop. Nu clue what the issue is right now. Not more time to dig in further. Got the following error:

```
Failed to restart gitlab-runner.service: Unit not found.

```

locals.tf
```diff
@@ -27,6 +27,8 @@ locals {
     }
   )
 
+  runners_pull_policies = var.runners_pull_policy != "" ? "[${var.runners_pull_policy}]" : "[\"${join("\",\"", var.runners_pull_policies)}\"]"
```
npalm, line 30:
```suggestion
  runners_pull_policies = var.runners_pull_policy != "" ? "[\"${var.runners_pull_policy}\"]" : "[\"${join("\",\"", var.runners_pull_policies)}\"]"
```

template/runner-config.tpl
```diff
@@ -24,7 +24,7 @@ listen_address = "${prometheus_listen_address}"
     disable_cache = ${runners_disable_cache}
     volumes = ["/cache"${runners_additional_volumes}]
     shm_size = ${runners_shm_size}
-    pull_policy = "${runners_pull_policy}"
+    pull_policy = ${runners_pull_policies}
```
npalm, line 27:
```suggestion
    allowed_pull_policies = ${runners_pull_policies}
```

variables.tf
```diff
@@ -232,11 +232,17 @@ variable "runners_helper_image" {
 }
 
 variable "runners_pull_policy" {
-  description = "pull_policy for the runners, will be used in the runner config.toml"
+  description = "Deprecated! Use runners_pull_policies instead. pull_policy for the runners, will be used in the runner config.toml"
   type        = string
   default     = "always"
```
npalm, line 237:
```suggestion
  default     = ""
```

npalm (changes requested):
Just went over the PR again, made some small suggestion which fixed the non starting agent.

- used `allowed_pull_policies` 
- set default for old pull_policy to empty, so by default new one is used.

npalm (approved):
Accepted my suggestions, tested and all good. Thanks for your contribution

cinacio:
> Just went over the PR again, made some small suggestion which fixed the non starting agent.
> 
>     * used `allowed_pull_policies`
> 
>     * set default for old pull_policy to empty, so by default new one is used.

First of all, thanks a lot for all the hard work on this module that makes our lives much easier!

I'm wondring about this change and the above suggestions. It's quite nice that we can now specify allowed pull policies, but if I read the docs correctly this slightly changes the behaviour of the modified directives and now prevents us from specifying a "default" pull policy for runners.

Setting `allowed_pull_policies` will define the policies that can be used in the `.gitlab-ci.yml` file or the runner configuration, but keeping `pull_policy` unset, will keep it at its default value, which is `always` according to the [GitLab runner advanced configuration documentation](https://docs.gitlab.com/15.4/runner/configuration/advanced-configuration.html#the-runnersdocker-section). Thus preventing us from modifying the "default" pull policy at the runner level, and forcing image downloads unless the policy is explicitly set in job configurations. Or am I missing something?

Wouldn't it be better to have two different directives, actually?

kayman-mk:
Also see #511

cinacio:
@npalm As I figured @gmeligio might be busy elsewhere, I took the liberty of creating PR #557 just to set `pull_policy` instead of `allowed_pull_policies`.

I hope that's alright.

PR #545, opened 2022-09-15
feat: Skip yum update

Swsystem-ATuin:
## Description

This will allow for runners to be launched from a known working AMI and not have yum update introduce unexpected issues.
The default is to run the yum update as is current but allows for disabling.

## Migrations required

NO - Existing behaviour not changed

## Verification

Tests based on Example - Spot Runner - Default without issue.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -814,8 +814,8 @@ variable "asg_terminate_lifecycle_lambda_timeout" {
   type        = number
 }
 
-variable "do_yum_update" {
-  description = "Run a yum update as part of starting the instance"
+variable "runner_yum_update" {
+  description = "Run a yum update as part of starting the runner"
   default     = "true"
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -813,3 +813,9 @@ variable "asg_terminate_lifecycle_lambda_timeout" {
   default     = 30
   type        = number
 }
+
+variable "runner_yum_update" {
```
npalm, line 817: Can you convert the variable to a boolean instead of the string.

npalm (changes requested):

variables.tf
Swsystem-ATuin: Variable is now boolean.

npalm (approved):
Thanks

PR #547, opened 2022-09-27
feat: add extra_hosts to config.toml

veshant:
## Description

Adds the missing parameter clone_url to the config.toml of the runner. In case the default clone URL from the GitLab project is not accessible from the runner's infrastructure (e.g. private DNS), the clone_url can be used to specify a different one (e.g. replace the private DNS entry by a public IP address).

Check the [configuration docs for the runner](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runners-section).

## Migrations required

NO

## Verification

 Deploy infrastructure with extra_hosts. Check that the runner can register and config.toml shows the extra_hosts. Runners communicate using updated hosts records.


npalm (approved):

npalm:
Thanks!

PR #553, opened 2022-10-12
chore: fix TfSec issues

kayman-mk:
## Description

This PR fixes all mentioned TfSec issues reported by the pipeline.

- encrypt all resources either with the provided/generated customer managed key or the default AWS key
- specify resources in policies with ARN instead of `*`
- remove unnecessary policies
- allow activation of X-Ray tracing via variable

## Migrations required

No

## Verification

- [x] deploy/destroy runner-default
- [x] termination lambda can write logs
- [x] runners can still access the cache
- [x] access logs are written if the feature is enabled
- [x] update Hapag-Lloyd runners to this version and ensure everything is running


modules/cache/main.tf
```diff
@@ -25,6 +25,8 @@ resource "random_string" "s3_suffix" {
   special = false
 }
 
+# ok, we can go without access logging as it is a cache bucket only
```
npalm, line 28: This is certainly OK for me, and can be made advanced later. But see here a PR on one of my other repos where optional logging is added

https://github.com/philips-labs/terraform-aws-github-runner/pull/2430/files#diff-3d07fd4aceafd6afc7a21efd7392168bb00a3bb8f1a3c5099aa176b8bd8ecb96R65

modules/cache/main.tf
```diff
@@ -64,12 +68,15 @@ resource "aws_s3_bucket_lifecycle_configuration" "build_cache_versioning" {
...
 
+# decision by user whether to use a customer managed key or not. Resource is encrypted either.
+# tfsec:ignore:aws-s3-encryption-customer-key
 resource "aws_s3_bucket_server_side_encryption_configuration" "build_cache_encryption" {
   bucket = aws_s3_bucket.build_cache.id
 
   rule {
     apply_server_side_encryption_by_default {
-      sse_algorithm = "AES256"
+      sse_algorithm     = "aws:kms"
```
npalm, line 78: if no key is provided, is the bucket encrypted with a default key?

npalm (commented):
Nice job so far! Some small comments. So far the PR is working with the default example, without any breaking effect.

modules/cache/main.tf
kayman-mk: Yes, that's true. From the docs: `The default aws/s3 AWS KMS master key is used if this element is absent while the sse_algorithm is aws:kms`.

But we should set the `bucket_key_enabled = true`. This will save up to 99% of the encryption cost.

modules/cache/main.tf
kayman-mk: Good point! We should consider this.

modules/cache/main.tf
kayman-mk: fixed

modules/cache/main.tf
kayman-mk: Logging support added

npalm:
For the access logs I would suggest to support the by the module but let the user inject a bucket. It is quiet common there is already a bucket for logs, even in another account. Shall I propose a implementation?

kayman-mk:
Appreciated. Please go ahead.

logging.tf
```diff
@@ -19,5 +19,8 @@ resource "aws_cloudwatch_log_group" "environment" {
   name              = var.log_group_name != null ? var.log_group_name : var.environment
   retention_in_days = var.cloudwatch_logging_retention_in_days
   tags              = local.tags
-  kms_key_id        = local.kms_key
+
+  # ignored as decided by the user
```
npalm, line 23:
```suggestion
  # as decided by the user
```

kayman-mk:
@npalm Access logging is now active, but no logs are written. Checked the AWS documentation and also tried to enable the logging via AWS console. But no logs appear. Any idea?

kayman-mk:
Needs 15+ minutes but the access logs appear.

Last test is still missing.

npalm (commented):
Nice job, looks good!

kayman-mk:
Last test done. Jobs are still processed.

examples/runner-default/main.tf
```diff
@@ -40,6 +40,10 @@ module "runner" {
   runners_name             = var.runner_name
   runners_gitlab_url       = var.gitlab_url
   enable_runner_ssm_access = true
+  cache_bucket = {
```
npalm, line 43: Maybe I am missing something, but why adding this object to the example. What is causing this need?

npalm (commented):
Looks good, nice work. One small question

npalm:
@kayman-mk I use mostly the default example to test a PR with a simple build and cache test. Seems somthing is changed and the cache is not working anymore. Got the following error:

```
FATAL: received: 404 Not Found                     
Failed to create cache
```

examples/runner-default/main.tf
kayman-mk: Seems that this is from one of my other branches. I will remove it here.

kayman-mk:
> @kayman-mk I use mostly the default example to test a PR with a simple build and cache test. Seems somthing is changed and the cache is not working anymore. Got the following error:
> 
> ```
> FATAL: received: 404 Not Found                     
> Failed to create cache
> ```

Rechecked this. I suspect that there was no cache present in your configuration. I see this message the very first time. But restarting the job shows, that the cache can be accessed.

```
Restoring cache
Using docker image sha256:bba2a6f41b3acaf2219ca7c8af8abd8d7a40953bf0dadd8824b for gitlab-runner-helper-runner-15-1-1:stable with digest gitlab-runner-helper-runner-15-1-1@sha256:77ef1d02e2834df8b283bb267235b74db1458d8610b ...
Checking cache for default-non_protected...
Downloading cache.zip from https://prod-923[34](https://redacted/pipeline/-/jobs/2341959#L34)7-gitlab-runner-cache.s3.dualstack.eu-central-1.amazonaws.com/project/987/default-non_protected 
Successfully extracted cache
```

npalm (approved):

PR #557, opened 2022-10-25
fix: Set pull_policy instead of allowed_pull_policies

cinacio:
## Description

Since #544 was merged in, `runners_pull_policies` is used to set the runner's `allowed_pull_policies` instead of `pull_policy`. That is confusing, and it changes the behaviour of the directive and prevents setting the default pull policy for the runner.

This change switches back to using `runners_pull_policies` to set the runner's `pull_policy`, while still allowing for multiple policies to be defined.

## Migrations required

NO

## Verification

I have verified that the correct `pull_policy` directive is now set on the runners.

### DIFF
--- a/template/runner-config.tpl
+++ b/template/runner-config.tpl
@@ -25,7 +25,7 @@ listen_address = "${prometheus_listen_address}"
     volumes = ["/cache"${runners_additional_volumes}]
     extra_hosts = ${jsonencode(runners_extra_hosts)}
     shm_size = ${runners_shm_size}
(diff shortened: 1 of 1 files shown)


gmeligio:
Thank you @cinacio. Looks good to me.

kayman-mk:
Looks good to me as well. Supporting the `allowed_pull_policies` needs a new variable. The variable description of `runners_pull_policy` also indicates that `pull_policy` should be used in the `config.toml`.

npalm (approved):

PR #564, opened 2022-11-02
Add runners_gitlab_certificate variable

### LINKED ISSUE #563: Support for self-signed certs on self-hosted gitlab instance
Documentation should indicate how to configure the certificate for self-signed certs on gitlab server.

Gitlab supports the following approaches:
https://docs.gitlab.com/runner/configuration/tls-self-signed.html#supported-options-for-self-signed-certificates-targeting-the-gitlab-server

Currently, `tls-ca-file` setting is not available in the module. Perhaps it can be added.
Documentation should be updated either way, and show a working example to make the module easier to use.

baolsen:
## Description

Closes #563 

## Migrations required

NO

## Verification

I've verified that the iniital change works in my own environment where I have self-signed certificates in use.
I have not verified that the changes after enabling pre-commit and tf docs etc work, but they should...


baolsen:
This change simply allows the Gitlab runner to register and run with a self-signed gitlab instance. The *runner helper* image can access such an instance.

For more complex use cases such as within a *user image* being able to access resources on the self-signed instance, there are additional steps needed. I'm still working on these, and may submit the changes as a secondary PR once I have them ironed out.

More info: https://docs.gitlab.com/runner/configuration/tls-self-signed.html#trusting-the-certificate-for-user-scripts 

An example of a more complex use case would be building a new docker image based on an existing docker image that resides on the self-hosted instance. Then, the user-image which is building the new image needs to pull the existing one from the self-hosted gitlab.

baolsen:
I've updated the module to additionally accept a CA cert bundle which is added to the runner-helper-image truststore.

This allows us to trust to various corporate severs (not just gitlab) for fetching build artefacts which might not reside on gitlab, for example.

It just needs some testing.

This completes the functionality needed for this PR.

For user images, the certs folder needs to be mounted on the docker image and then imported into the image truststore as part of the user CI file. This is not included and needs to be specified by the user if it is required (existing functionality allows for it so no extra development is needed here)

kayman-mk:
> For user images, the certs folder needs to be mounted on the docker image and then imported into the image truststore as 
> part of the user CI file. This is not included and needs to be specified by the user if it is required (existing functionality
> allows for it so no extra development is needed 

Would be nice to have an example somewhere.

baolsen:
Alright, I ended up adding the standard pre-commit stuff I use elsewhere with Terraform, so that the examples are updated with terraform-docs content automatically on commit.

Of the precommit stuff, I commented out `terraform validate`, `checkov` and `tfsec` which were not in place on this repo anyway and there was a lot of output from them which I thought can be looked at in another PR. To try to reduce the noise.

It still meant a lot of files (60+) were changed....

@kayman-mk let me know what you think.

I can always remove this last commit and open a new PR for it, if it is too noisy.
In that case, I'd just keep the changes to the examples relating to certificates, and move all the other pre-commit related changes into a different PR.

baolsen:
Related: #553

baolsen:
@kayman-mk please take a look when you can.

kayman-mk:
Any chance to reduce the noise?

The docs are created within the pipeline as far as I remember. So no need to commit them in a feature branch.

npalm (changes requested):
Sorry for keep you waiting that long. Please can you rebase your PR it looks like the diff contains all kind of changes not related to the PR. Hard to find the real change..

baolsen:
Re-opened with less noise: #584

PR #565, opened 2022-11-02
Zip user_data

### LINKED ISSUE #561: user_data is almost fully utilised
User data is nearly at the 16kB limit. 
This causes problems when user wants to add some content for example multiple server certificates. 

Suggest gzipping user data content if larger than 16k

baolsen:
## Description

Closes #561 

## Migrations required

NO

## Verification

Done:
EC2 runner working correctly in my environment for several days now.
On the EC2 Launch Template, User data is visible as b64 + gzip encoded.
On the EC2 instance panel, User data displays as "No user data" for some reason.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -223,7 +223,7 @@ data "aws_ami" "runner" {
 resource "aws_launch_template" "gitlab_runner_instance" {
   name_prefix            = local.name_runner_agent_instance
   image_id               = data.aws_ami.runner.id
-  user_data              = base64encode(local.template_user_data)
+  user_data              = base64gzip(local.template_user_data)
   instance_type          = var.instance_type
   update_default_version = true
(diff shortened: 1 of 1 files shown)


kayman-mk:
Sounds reasonable to me. Found the documentation here: https://cloudinit.readthedocs.io/en/latest/topics/format.html#gzip-compressed-content

GZip content is decoded automatically.

baolsen:
I've tested this with a local copy of the source code and it works fine.

A downside is you can't inspect the user_data element in the AWS UI as it is Gzip compressed. 
We do have cloudwatch logs to compensate for that.

baolsen:
@npalm ready for review and merge.

npalm (approved):
@baolsen thanks!

PR #569, opened 2022-11-08
Add create_cloudwatch_log_group variable

### LINKED ISSUE #559: Cloudwatch logs group destroyed with module
Currently cloudwatch logs group is created and destroyed along with the module.

For audit purposes we have a use case not to delete any logs.

It would be best for us to create the cloudwatch logs group outside the module, and manage its lifecycle separately.

Possible solution would be adding create_cloudwatch_logs_group boolean to the module.

baolsen:
## Description

Closes #559 

## Migrations required

NO

## Verification

None yet

### DIFF
--- a/logging.tf
+++ b/logging.tf
@@ -1,6 +1,6 @@
 resource "aws_iam_role_policy" "instance" {
   count  = var.enable_cloudwatch_logging ? 1 : 0
-  name   = "${local.name_iam_objects}-instance-role"
+  name   = "${local.name_iam_objects}-instance-logging"
   role   = aws_iam_role.instance.name
   policy = templatefile("${path.module}/policies/instance-logging-policy.json", { partition = data.aws_partition.current.partition })
 }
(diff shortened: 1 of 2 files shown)


kayman-mk:
You have to pass the name of the existing log group to the module, haven't you? Just missing a variable for that.

baolsen:
Thanks for the feedback @kayman-mk 

There is an existing variable `log_group_name` for that. 

It is a bit hard to notice because it is not named like the others:
`enable_cloudwatch_logging`
`create_cloudwatch_log_group`
`cloudwatch_logging_retention_in_days`

Perhaps the naming should be standardised. What do you suggest?

kayman-mk:
Looks good to me.

Yeah, there are so many variables around. But standardize these names requires a separate PR. I wouldn't change anything here.

baolsen:
Opened #575 for variable standardisation suggestions.
Will add some verification of this PR (currently none recorded here) and then I think it is good to go.

baolsen:
No longer required for our use case, closing the PR and issue. 
Someone else can open if needed...

PR #572, opened 2022-11-09
Allow custom runner agent IAM role

### LINKED ISSUE #571: Allow custom runner agent IAM role
I have a use case where the role attached to the EC2 instance needs elevated privileges in my account.

This makes it very simple to run CI pipelines without needing to configure specific IAM policies for the pipeline requirements; instead the runner role can effectively perform any action (with some essential gaurdrails in place)

To enable this functionality, we require the ability to specify our own IAM role and to disable the creation of the runner agent role by this module. 
Then, we can pass the name of our role to the module and it uses it as the instance-attached role, thus getting the elevated permissions.

The alternative was to allow our teams to attach the elevated permissions policy onto the runner role using the existing arguments to add policies to the role. However this would allow the teams to attach the elevated permissions onto any role rather than only the specific runner role. So instead, we create an elevated role with the elevated permissions, and only allow this role to be used by the runner when it is created.

baolsen:
## Description

Closes #571 

## Migrations required

NO

## Verification

Verified in my local environment

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -323,11 +323,12 @@ module "cache" {
 ################################################################################
 resource "aws_iam_instance_profile" "instance" {
   name = "${local.name_iam_objects}-instance"
-  role = aws_iam_role.instance.name
+  role = local.aws_iam_role_instance_name
   tags = local.tags
 }
 
 resource "aws_iam_role" "instance" {
+  count                = var.create_runner_iam_role ? 1 : 0
   name                 = "${local.name_iam_objects}-instance"
   assume_role_policy   = length(var.instance_role_json) > 0 ? var.instance_role_json : templatefile("${path.module}/policies/instance-role-trust-policy.json", {})
   permissions_boundary = var.permissions_boundary == "" ? null : "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:policy/${var.permissions_boundary}"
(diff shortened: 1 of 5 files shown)


main.tf
```diff
@@ -323,11 +323,12 @@ module "cache" {
 ################################################################################
 resource "aws_iam_instance_profile" "instance" {
   name = "${local.name_iam_objects}-instance"
-  role = aws_iam_role.instance.name
+  role = local.aws_iam_role_instance_name
   tags = local.tags
 }
 
 resource "aws_iam_role" "instance" {
+  count                = var.create_runner_iam_role ? 1 : 0
```
kayman-mk, line 331: question: do we need a separate `create_runner_iam_role` or can we use something like `var.runner_aim_role_name > ""` instead?

baolsen: Thanks for the feedback.

As users may have differing IAM requirements, I think both are useful to support possible use cases:

 - Default / current behaviour: Create role and use default suggested name.
 - Create role with custom name: `runner_iam_role_name=myrole`
 - Do not create role, use some existing role: `runner_iam_role_name=myrole` and `create_runner_iam_role=false`

kayman-mk: Oh yes, true.

baolsen:
> Please can yu have a look on the failing CI?

Sorted, thanks

PR #580, opened 2022-11-27
ci: lint all files with Megalinter

kayman-mk:
## Description

This PR makes sure that all files are syntactically valid and adhere to the specific format. This makes it easier to read the files in the repository and that the code follows the industry standard.

[Megalinter](https://github.com/oxsecurity/megalinter) is used to have all relevant linters at hand. But some of them were disabled:
- Trivy: Nice linter to report CVEs and other cool stuff. But it reports problems with the Terraform code which can't be disabled by configuration
- TfLint: Has some problems reading referenced policy files. I created a separate workflow for TfLint as it creates valuable output.
- Markdown table formatter: The tables created by TfDoc are not formatted according to the standard. And there is no option to deactivate the check for a section of the file.
- Misspell: CSpell does a great job. No need for a second linter

## Migrations required

No

## Verification

Verified that the job runs and throws an error if a markdown file is not valid.


kayman-mk:
@npalm What do you think about linting the markdown files (auto generated parts are excluded)? Just to make sure to follow the same schema and have valid markdown only.

npalm:
Great idea to start linting arkdown. What would you think about using the megalinter (https://megalinter.io/latest/configuration/)? The megalinter could also be enabled only for markdown but can also be used to lint action yml, bash, docker and any other file in the repo excluding terraform. Since for Terraform all kin dof linting is already in place.

What are you thoughts?

kayman-mk:
Gosh, blocked by my firewall. Have to check later.

Linting all other files would be nice too. Didn't know that there is a Megalinter available. Makes things easier. Up to know I used specific linters for different file types. Good to know that things can be simplified.

Let's change to the one you proposed.

npalm:
> Gosh, blocked by my firewall. Have to check later.
> 
> Linting all other files would be nice too. Didn't know that there is a Megalinter available. Makes things easier. Up to know I used specific linters for different file types. Good to know that things can be simplified.
> 
> Let's change to the one you proposed.

only drawback is that those all-in-one linter like megalinter or superlinter are heavy packed containers. But i like them for using anything expect the main language in the reposiroty.

kayman-mk:
Let's merge #553 first which solves the TfSec problems.

.github/workflows/linter.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  build:
+    name: MegaLinter
+    runs-on: ubuntu-latest
+    steps:
+      # Git Checkout
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.PAT || secrets.GITHUB_TOKEN }}
```
npalm, line 25: I try to avoid PAT, but use app token in stead.

kayman-mk: Copy & paste. PAT removed as not needed.

npalm:
@mkalmykov since we have already tfsec and standard terraform linters. I think we should disable all the Terraform checks in the megalinter. What is your thought?

kayman-mk:
I am not 100% sure that we should disable the Terraform checks in Megalinter. There are some interesting linters here, so may be we should drop the tfsec action.

kayman-mk:
Gotcha! Megalinter is running. There are 3 false warnings about lines in a markdown file which are too long. But the lines mentioned are ok. So I ignore it for now.

- several issues fixed/ignored which were reported by the linters
- I disabled some of the linters as explained in the PR.
- the Terraform format check has been removed as it is done by Megalinter now
- line length for markdown files is set to 132
- removed unused `locals.tags_string`
- globally ignored some Terrascan rules as they can't be ignored inline: Lambda function tracing enabled
- tidy up Go dependencies
- fixed some spelling issues

security_groups.tf
```diff
@@ -2,7 +2,10 @@
 ## Gitlab-runner agent security group ##
 ########################################
 
+# ignores: IAM Access Analyzer Not Enabled
```
npalm, line 5: Just wondering, see you have to ignore this on many plces

1. `ignores: ...` Is this just a comment? Or instructing to ignore?
2. Kics-scan ignore-line. Dow e get anything valuable from kics?

npalm (changes requested):
Great job, looks good. Just a few tiny remarks

.mega-linter.yml
```diff
@@ -0,0 +1,6 @@
+---
+DISABLE_LINTERS:
```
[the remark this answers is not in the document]
kayman-mk: Sure. Mixed something up and thought that comments are not allowed in YAML files. Put everything into the PR.

Will rework.

.vscode/extensions.json
```diff
@@ -1,10 +0,0 @@
-{
```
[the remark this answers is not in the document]
kayman-mk: Tried to Google it as I am not familiar with that. Several sources suggested to not check in these files.

security_groups.tf
kayman-mk: Will recheck the value of KICS. Seems that it has several false positives.

The `ignores:` is just a comment, so we know in the source what's being ignored.

.vscode/extensions.json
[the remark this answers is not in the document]
kayman-mk: It's back now. Seems to make sense to store theses settings in the repository.

security_groups.tf
kayman-mk: Removed nearly all false positive ignores and configured KICS to ignore these rules. Should be covered anyway by other linters.

Suggest to go ahead with KICS and check how it works in daily life.

kayman-mk:
@npalm Fixed the stuff.

npalm (dismissed):

kayman-mk:
Megalinter needs 4+ minutes. Maybe it's better to use dedicated linters to have a fast workflow.

Tried it here: https://github.com/Hapag-Lloyd/Repository-Templates/blob/main/.github/workflows/linter.yml

kayman-mk:
@npalm Finished this one. The linters make sure that we follow the industry standard. Let's do it for some weeks and check the total workflow duration. If we are not satisfied, we can split the linters into single workflows.

.github/workflows/pr-opened.yml
```diff
@@ -15,6 +19,8 @@ jobs:
     steps:
       - uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # ratchet:actions/github-script@v6
         with:
+          # no idea how to break this long line
+          # yamllint disable rule:line-length
```
npalm, line 23: Normally this is possible with | or > like. > should keep the line breaks, > replace them by spaces

```yaml
abc: >
   some text
   some more

npalm (approved):

npalm:
Nice one, the Ci is stail failing, KICS gives one error. And the PR also has no write access. Maybe enalbling this can solve the 403 https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

kayman-mk:
It was green, wasn't it? I wouldn't send the message in case of red pipeline. Strange.

Let me check this.

kayman-mk:
@npalm Seems to be fine now.

npalm (approved):

kayman-mk:
@npalm Need a new approval. Had to solve some conflicts.

npalm (approved):

PR #586, opened 2022-12-01
feat(asg): Add fine-grained options for schedule_config scale_in and scale_out

### LINKED ISSUE #585: feat: Support more fine-grained control over ASG scaling
Currently the ASG scaling can be controlled with the following parameters in `schedule_config`:

- "scale_in_count"
- "scale_out_count"

These set all of the `min_size`, `max_size` and `desired_capacity` properties of the relevant `aws_autoscaling_schedule` resources to the specified value.

For example,  say `scale_out_count=2`, then `min_size=2, max_size=2, desired_capacity=2` are all set for `aws_autoscaling_schedule.scale_out`.

We have a use case where we want different values for these parameters. Specifically, we normally want the ASG to create 0 instances but we will have a separate lambda function which will set the `desired_capacity=1` on demand. 

However, at the moment the runner module makes this impossible as we cannot set the ASG properties to have different values as follows: **`min_size=0, max_size=1, desired_capacity=0`** and then have our lambda change `desired_capacity: 0 => 1`.

baolsen:
## Description

Closes #585 

## Migrations required

NO

## Verification

None. I've locally tested the "try" logic in `terraform console` to ensure it works with / without the overrides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,34 @@ Once you have created the parameter, you must remove the variable `runners_token
 
 Finally, the runner still supports the manual runner creation. No changes are required. Please keep in mind that this setup will be removed in future releases.
 
-### Auto Scaling Group Instance Termination
+### Auto Scaling Group
+
+#### Scheduled scaling
+
+When `enable_schedule=true`, the `schedule_config` variable can be used to scale the Auto Scaling group. 
+
+Scaling may be defined with one `scale_out` scheduled action and/or one `scale_in` scheduled action.
+
+For example:
+```hcl
+  module "runner" {
+    ...
+    enable_schedule = true
+    schedule_config = {
+      # Configure optional scale_out scheduled action
+      scale_out_recurrence = "0 8 * * 1-5"
+      scale_out_count      = 1 # Default for min_size, desired_capacity and max_size
+      # Override using: scale_out_min_size, scale_out_desired_capacity, scale_out_max_size
+
+      # Configure optional scale_in scheduled action
+      scale_in_recurrence  = "0 18 * * 1-5"
+      scale_in_count       = 0 # Default for min_size, desired_capacity and max_size
+      # Override using: scale_out_min_size, scale_out_desired_capacity, scale_out_max_size
+    }
+  }
+```
+
+#### Instance Termination
 
 The Auto Scaling Group may be configured with a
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -148,7 +148,34 @@ Once you have created the parameter, you must remove the variable `runners_token
...
+    schedule_config = {
+      # Configure optional scale_out scheduled action
+      scale_out_recurrence = "0 8 * * 1-5"
+      scale_out_count      = 1 # Default for min_size, desired_capacity and max_size
+      # Override using: scale_out_min_size, scale_out_desired_capacity, scale_out_max_size
+
+      # Configure optional scale_in scheduled action
+      scale_in_recurrence  = "0 18 * * 1-5"
+      scale_in_count       = 0 # Default for min_size, desired_capacity and max_size
+      # Override using: scale_out_min_size, scale_out_desired_capacity, scale_out_max_size
```
kayman-mk, line 173: question: Expected a `scale_out` when leaving the office at 6pm and a `scale_in` when coming back at 8am. Have they been swapped by accident?

kayman-mk:
Nice! Have never noticed this `schedule_config` before.

kayman-mk:
Could you please run a `terraform apply` with a dummy example? Just to make sure that the module is still deployable. Code looks good.

README.md
baolsen: Thanks for the feedback.

They look correct, same as before. 
`scale_out` at the start of the day (create more instances) and `scale_in` at end of the day (terminate).

It does get easily confused. I tried to move the code slightly to clarify / avoid confusion but the logic is identical otherwise.

Maybe clearer if I change from:
`# Configure optional scale_out scheduled action`
to
`# Configure optional scale_out scheduled action`
`# to create instances at the start of my work day`

and
`# Configure optional scale_in scheduled action`
`# to terminate instances at the end of my work day`

What do you think @kayman-mk

npalm (commented):
Can you explain what the change is solving. It seems only defaults are changed which can be done by setting a variable to your need.

baolsen:
Thanks for the feedback @npalm , the "actual" change is in `main.tf` and addresses the issue #585 

Specifically, when the values for `min_size`, 'max_size` and `desired_capacity` need to be _different_ from one another - current functionality only allows setting them all to the _same_ value (specified by `xxx_count`).

kayman-mk (approved):

npalm (approved):

PR #594, opened 2022-12-08
chore: Replace semantic release by release-please

npalm:
## Description

Replacing semantic release by release please. This to make releasing simpler. Release please will create and maintain a PR for the next release on the main branch. Once the PR is merged the release is created.

- example release: https://github.com/npalm/test-repo/releases/tag/5.6.0
- example PR: https://github.com/npalm/test-repo/pull/4

As side effect the develop branch will be removed, only a main branch will be maintained. 

## Migrations required

NO

## Verification

On test repo see links


npalm:
@kayman-mk please can you check the PR, but do not merge. Since I need to do some branching working at the same time as dropping develop as main branch

kayman-mk:
This sounds interesting. Just checked the documentation but I am not 100% sure how it works. Have to check it separately.

.github/workflows/ci.yml
```diff
@@ -48,8 +47,8 @@ jobs:
       - run: terraform validate
 
   docs:
-    # update docs after merge back to develop
-    if: github.event_name != 'pull_request' && contains('refs/heads/develop', github.ref)
+    # update docs after merge back to main
+    if: github.event_name != 'pull_request' && contains('refs/heads/main', github.ref)
```
kayman-mk, line 51: question: Is this really working? I think we are not allowed to push to `main` directly. And if we do, this might create an entry in the changelog which we don't want to have, right?

suggestion: Might be better to add this step to the feature branch pipeline.

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+        id: token
+        with:
+          appId: ${{ secrets.APP_ID }}
+          appPrivateKeyBase64: ${{ secrets.APP_PRIVATE_KEY_BASE64 }}
+          appInstallationType: repo
+          appInstallationValue: ${{ github.repository }}
+
+      # bootstrap-sha and release-as needs to be removed after first release
+      - name: Release
+        uses: google-github-actions/release-please-action@v3
```
kayman-mk, line 23: praise: What a nice tool you have found! Just checked it in my demo repository. Works like a charme and makes things easier.

kayman-mk (approved):

.github/workflows/ci.yml
npalm: You right, but the App installed in the repo is allowed to bypass the branch protection. Is needed to update the changelog.

PR #596, opened 2022-12-10
Fix runner_agent_role_arn

FelikZ:
## Description

Fix missing `:` - `aws_iam_role_instance_arn` is invaild without this fix.

## Migrations required

NO

## Verification

Found, fixed and checked on my own Gitlab cluster.


kayman-mk (approved):
Good spot! Indeed, 2 colons are needed here.

npalm (approved):

PR #604, opened 2022-12-12
ci: fix the TfDocs creation in feature branch

kayman-mk:
The workflow was not running in case the PR was created from a foreign repository. The checkout of the branch didn't succeed.

The TF docs are now generated in the release branch if a commit is pushed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,8 +56,6 @@ jobs:
     steps:
     - name: CHeckout branch
       uses: actions/checkout@v3
-      with:
-        ref: ${{ github.event.pull_request.head.ref }}
 
     - name: Generate TF docs
       uses: terraform-docs/gh-actions@v1.0.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,8 +56,6 @@ jobs:
     steps:
     - name: CHeckout branch
       uses: actions/checkout@v3
-      with:
-        ref: ${{ github.ref }}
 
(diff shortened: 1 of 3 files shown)


kayman-mk:
@npalm Just noticed that the TfDocs creation in the feature branch is not working. Could you please have a look? Tried some fixes but they are not working.

First seen here: #517

.github/workflows/ci.yml
```diff
@@ -56,8 +56,6 @@ jobs:
     steps:
     - name: CHeckout branch
       uses: actions/checkout@v3
-      with:
```
npalm, line 59: Guess that without this, it breaks on PRs from a fork. Maybe better to just generate the docs on the branch for the release PR.

npalm (approved):

npalm:
seems still failing

.github/workflows/ci.yml
kayman-mk: Yeah, but how?

kayman-mk:
Simply using a plain `checkout` does not work as it leaves the repository in a detached head state. But we need a branch to push a commit.

npalm:
I was thinking we can simply run the updates of the docs on the release branch, merging that branch is making the release.

kayman-mk:
Let's give it a go. Better than now.

But the `main` branch is not always in a consistent state as the release might be done several days later.

kayman-mk:
@npalm This one should be ready now. Docs are created in the release branch as discussed.

npalm (approved):

PR #608, opened 2022-12-16
feat: add `amazonec2 userdata` for docker machines

FelikZ:
## Description

Introducing support for [amazonec2-userdata flag](https://github.com/docker/machine/blob/b170508bf44c3405e079e26d5fdffe35a64c6972/drivers/amazonec2/amazonec2.go#L284) that is not mentioned on [driver readme page](https://gitlab.com/gitlab-org/ci-cd/docker-machine/-/blob/main/docs/drivers/aws.md), but it is there and useful.

For example, it can be used to introduce swap for the runner instance or any other user command.

## Migrations required

NO - this is optional flag and defaults to empty string.

## Verification

I have verified on my own runners cluster that user script that I've added works (it creates swap in my case).

```
module "runner" {
  # source  = "npalm/gitlab-runner/aws"
  # version = "5.5.0"

  # ...

  runners_userdata = templatefile("${path.module}/../../templates/swap.tpl", {
    swap_size = "512"
  })
}
```


FelikZ:
But I have no idea how to auto-generate README.md, I think this flag should be documented. (manually edit?)

template/gitlab-runner.tpl
```diff
@@ -14,6 +14,10 @@ ${runners_config}
 
 EOF
 
+cat > /etc/gitlab-runner/runners_userdata.tpl <<- EOF
```
kayman-mk, line 17: issue: File suffix should be `.sh` or similar as this is not a template file.

template/runner-config.tpl
```diff
@@ -65,6 +65,7 @@ listen_address = "${prometheus_listen_address}"
       "amazonec2-iam-instance-profile=%{ if runners_iam_instance_profile_name != "" }${runners_iam_instance_profile_name}%{ else }${runners_instance_profile}%{ endif ~}",
       "amazonec2-root-size=${runners_root_size}",
       "amazonec2-volume-type=${runners_volume_type}",
+      "amazonec2-userdata=%{ if runners_userdata != "" }/etc/gitlab-runner/runners_userdata.tpl%{ endif ~}",
```
kayman-mk, line 68: chore: to be tested without specifying `runners_userdata`

kayman-mk (changes requested):
Good PR! Just 2 little remarks.

template/runner-config.tpl
kayman-mk: Successfully tested. The executors are still running. Nevertheless I had problems with one executor which was not able to update the Ubuntu system. Quite strange. Seems not to be related to this PR.

@npalm Just to be on the safe side: Could you test this as well please?

kayman-mk (approved):

kayman-mk:
@npalm See my comment above. Better you check this as well.

npalm (changes requested):
@FelikZ thanks looks good! Please can you add the example. Which you mention in the PR to either the Readme. Or as comment in the default example. I would prefer the last one (example)

kayman-mk (approved):

kayman-mk:
Did you do a quick check without using the new variable `runners_userdata` to make sure that the agent is still able to start?

kayman-mk:
Ok, it's working. Had a test environment available and deployed your changes. It is working without `runners_userdata`.

PR #609, opened 2022-12-18
Fix "No roles attached to instance profile"

dpad:
## Description

For some reason, the EC2 instances created by the launch template do not automatically associate with the appropriate IAM Role, even though the name matches correctly. This change fixes this by using the ARN of the created role directly.

## Migrations required

NO

## Verification

Changing this line and re-applying on Terraform allowed the EC2 instance to associate with the correct IAM Role as expected.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -248,7 +248,8 @@ resource "aws_launch_template" "gitlab_runner_instance" {
     }
   }
   iam_instance_profile {
-    name = local.aws_iam_role_instance_name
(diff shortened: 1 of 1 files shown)


dpad:
Seems there was an issue on my AWS configuration where the instance profile already existed with the same name, but didn't get linked to the IAM Role. I had to manually remove the instance profile (using the AWS CLI) and the IAM role and recreate everything. So this change can be ignored, sorry to bother you!

internetstaff:
I ran into this after rebuilding a setup. If this fixes the issue with no regressions, it seems worth merging.

PR #618, opened 2023-01-01
feat: add user data of agent to outputs

### LINKED ISSUE #321: Add user_data output
Hello,

in our setup plan always shows `user_data` are to be changed even after same config was just applied. Value from config indeed seems to be the same as one we see in launch_configuration in aws console. We are unable to see what terraform thinks is a new updated value. Plan does not show whole value only its hash.

Is there any way how to see whole value of user_data? If not I would like to suggest to add new module output with value of `local.template_user_data`. This sounds useful for debugging in general since many configurations are effective through user data.

Thnks

kayman-mk:
## Description

Adds a new output `runner_user_data` to the module to show the value of the `user_data` script. Especially useful as the the script is now gzipped.

Closes #321 

## Migrations required

No

## Verification

No verification done. One output added only.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -52,3 +52,8 @@ output "runner_launch_template_name" {
   description = "The name of the runner's launch template."
   value       = aws_launch_template.gitlab_runner_instance.name
 }
+
+output "runner_user_data" {
+  description = "The user data of the Gitlab Runner Agent's launch template."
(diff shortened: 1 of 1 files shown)


kayman-mk:
TfSec problem is fixed with #620

npalm (approved):

PR #625, opened 2023-01-03
ci: validate PR title

kayman-mk:
## Description

Adds a new workflow action which lints the title of the PR. This is to make sure that we have commits adhering to the semantic commit rules only. Otherwise Release-Please is not able to determine the next version number and will not create a release.

### DIFF
--- a/.github/workflows/lint_pr_title.yml
+++ b/.github/workflows/lint_pr_title.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: amannn/action-semantic-pull-request@v5
         env:
-          GITHUB_TOKEN: ${{ secrets.LINT_PR_GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           # Default: https://github.com/commitizen/conventional-commit-types
           types: |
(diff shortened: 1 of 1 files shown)


kayman-mk:
The following settings should be active: Settings > General > Pull Requests: Allow squash merging only. Message shouldn't be the "Default message" as this might cause problems in case there is only one commit in the branch. It should be set to something starting with "Pull request title".

Personally I prefer "Pull request title and description". Especially as we have a PR template. Would be nice to see this in the commits too.

@npalm Could you please check this?

kayman-mk:
@npalm Could you please correct the commit message of "9056fb25170d659ddcbb37291220333a8e427950" in `main`? The prefix should read `feat` instead of `fet`.

.github/workflows/lint_pr_title.yml
```diff
@@ -0,0 +1,66 @@
...
+          # this option and therefore this action allows you to opt-in to using the
+          # special "[WIP]" prefix to indicate this state. This will avoid the
+          # validation of the PR title and the pull request checks remain pending.
+          # Note that a second check will be reported if this is enabled.
+          wip: true
+          # When using "Squash and merge" on a PR with only one commit, GitHub
+          # will suggest using that commit message instead of the PR title for the
+          # merge commit, and it's easy to commit this by mistake. Enable this option
+          # to also validate the commit message for one commit PRs.
+          validateSingleCommit: false
```
npalm, line 63:
```suggestion
          validateSingleCommit: true
```
Would enable this exact for the reason mentioned in the comment

npalm (changes requested):

npalm (approved):

Issue #623, opened 2023-01-01
Remove resources created during runtime (housekeeping)


kayman-mk:
As soon as the module is deployed it creates some resources during runtime. Namely:
- key pairs
- spot instance requests

These resources should be deleted as soon as possible. Afterwards we can get rid of the shell script to terminate the spot requests. I guess nobody is running so that it might happen that spot instances are ordered for outdated spot requests.

Proposal:
In case an Agent is terminated, we should housekeep all unused resources. We already have a Lambda function running for this purpose. It terminates the Executors created by the Agent. We should make this Lambda function mandatory.

- [x] make the lambda function mandatory and remove the variables that should be managed by the module itself (#628)
- [x] delete all unused key pairs (#652)
- [x] terminate all spot instance requests of the terminated Agent (#653)
- [x] amend docs to describe how to remove the module. Should be something like
    - set desired capacity of ASG to 0 so no new instances are created
    - kill all agents, if any. This removes all resources created during runtime (lifecycle hook)
    - Wait for 60s so the lambda function can run
    - run `terraform apply` or `terraform destroy` whatever is applicable
    #654 

Closes #592 (key pairs)
Closes #493 (spot requests)

kayman-mk:
Fully implemented

PR #628, opened 2023-01-04
refactor: make the agent termination lifecycle hook mandatory

kayman-mk:
## Description

Refactors the lambda function which handles killed agents.

- rename it to `terminate-agent-hook` to make clear that it deals with the agents and not with the executors
- make the function mandatory so it always runs as soon as an agent is terminated. This is needed in preparation for #623 as we want to remove unused resources. So the function has to run
- remove all variables which should be better managed by the module and not by the user, e.g. memory size and timeouts. They depend on the module only, so they should be managed by the module.

The variables which are no longer in use are still there, but have been marked as deprecated. So no migration is necessary.

## Migrations required

Remove the declaration of `asg_terminate_lifecycle_hook_heartbeat_timeout` and `asg_terminate_lifecycle_hook_create`. The AWS Lambda function is created always now.

## Verification

- [ ] deploy the whole module and check that the lambda function is still created


modules/terminate-agent-hook/main.tf
```diff
@@ -18,17 +18,17 @@ data "archive_file" "terminate_runner_instances_lambda" {
...
   function_name    = "${var.environment}-${var.name}"
   handler          = "lambda_function.handler"
-  memory_size      = var.lambda_memory_size
+  memory_size      = 128
   package_type     = "Zip"
   publish          = true
   role             = aws_iam_role.lambda.arn
-  runtime          = var.lambda_runtime
-  timeout          = var.lambda_timeout
+  runtime          = "python3.8"
```
npalm, line 30: can we update this without problems to 3.9?.

modules/terminate-agent-hook/main.tf
```diff
@@ -18,17 +18,17 @@ data "archive_file" "terminate_runner_instances_lambda" {
 # tfsec:ignore:aws-lambda-enable-tracing
 resource "aws_lambda_function" "terminate_runner_instances" {
   architectures    = ["x86_64"]
```
npalm, line 20: Is see you fixing options, guess since the lambda is provided by us. Maybe good also to move to arm. Faster and cheaper. I moved all my Node lambda's last year to arm

modules/terminate-agent-hook/variables.tf
```diff
@@ -39,36 +39,12 @@ variable "cloudwatch_logging_retention_in_days" {
...
-  default     = 90
-}
-
 variable "name_iam_objects" {
   description = "The name to use for IAM resources - roles and policies."
   type        = string
   default     = ""
 }
 
-variable "lambda_memory_size" {
```
npalm, line 54: Can you motivate why dropping the configuration. I assume since the lambda is mainted by the module.

variables.tf
```diff
@@ -848,31 +848,31 @@ variable "asg_terminate_lifecycle_hook_name" {
...
 variable "asg_terminate_lifecycle_hook_create" {
-  description = "Boolean toggling the creation of the ASG instance terminate lifecycle hook."
+  description = "(Deprecated and always true now) Boolean toggling the creation of the ASG instance terminate lifecycle hook."
   type        = bool
   default     = true
 }
 
 variable "asg_terminate_lifecycle_hook_heartbeat_timeout" {
-  description = "The amount of time, in seconds, for the instances to remain in wait state."
+  description = "(Deprecated and no longer in use) The amount of time, in seconds, for the instances to remain in wait state."
```
npalm, line 857: If not using, maybe better to set the value to null. And add a condition to check the value is null

variables.tf
```diff
@@ -848,31 +848,31 @@ variable "asg_terminate_lifecycle_hook_name" {
 }
 
 variable "asg_terminate_lifecycle_hook_create" {
-  description = "Boolean toggling the creation of the ASG instance terminate lifecycle hook."
+  description = "(Deprecated and always true now) Boolean toggling the creation of the ASG instance terminate lifecycle hook."
```
npalm, line 851: Maybe add a condition to force true

npalm (changes requested):

modules/terminate-agent-hook/variables.tf
kayman-mk: Yeap. That's the point. We manage the code. So we should know the exact memory size. I think it's error prone to let the user decide it.

By the way: Not sure if it the memory size has been analyzed. The more memory you reserve the more CPUs you get and in turn the function might run faster. We should evaluate that.

kayman-mk: #631 tackles this

modules/terminate-agent-hook/main.tf
kayman-mk: We should do it, yes. Not sure if everything will run smooth. Created #631 for this point.

npalm (approved):
The PR looks good, but since you mention we have to check the memory required for the lambda better. Maybe good to keep the parameter open for configuration. But in the end I think it does not matter much on the scale this lambda is executed. So please go ahead with merging.

PR #650, opened 2023-01-07
chore: pin Terraform provider versions with lock file

kayman-mk:
## Description

To get reproducable results the versions of the providers used in Terraform are fixed. This way we always know which provider was used when the example was applied.

## Migrations required

No.

## Verification

None. Changes in examples only

### DIFF
--- a/.terraform.lock.hcl
+++ b/.terraform.lock.hcl
@@ -1,0 +1,62 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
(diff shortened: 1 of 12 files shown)


npalm (approved):
Looks good, other option is to use the terraform lock file. The latest one is real immutable, since tags are mutable. But this is fine by me

kayman-mk:
Good point, Niek. This is only the 95% solution. Will fix that as long as we have a digest in the file.

npalm (approved):

PR #651, opened 2023-01-09
feat: suppress default tags from module

### LINKED ISSUE #413: Add support to disable default tags
I've hit situation where default module tags `Environment` and `Name` interfere with providers `default_tag` directive. This in result conflict with each Terraform run. 

Introducing something like `default_tags = true` would enable/disable defaults on `local.tags`

Is this a logical feature to add as I'd like to make PR myself.
Maybe there are any workarounds possible? 

 ```
 # module.gitlab-runner.aws_iam_instance_profile.instance will be updated in-place
  ~ resource "aws_iam_instance_profile" "instance" {
        arn         = "arn:aws:iam::1111111111111:instance-profile/staging-instance"
        create_date = "2021-12-03T09:06:50Z"
        id          = "staging-instance"
        name        = "staging-instance"
        path        = "/"
        role        = "staging-instance"
      ~ tags        = {
          + "Environment" = "staging"
            "Name"        = "staging"
        }
        tags_all    = {
            "Environment"    = "staging"
            "Infrastructure" = "v2"
            "ManagedBy"      = "Terraform"
            "Name"           = "staging"
            "Project"        = "Shared"
        }
        unique_id   = "AAAAAAAA"
    }
   ```

kayman-mk:
## Description

#413 describes the problem of interfering `default_tags` of the provider and the tags created by this module. Sometimes it is useful to skip the resource tagging by this module for certain tags.

This PR adds the ability to not add some/all of the default tags created by this module. Set `suppressed_tags` to a list of tags you want to suppress. It works for `tags`, `runner_tags` and `agent_tags`. `role_tags` are not modified as they are used in addition to `tags`.

Closes #413 

## Migrations required

No.

## Verification

- [x] Deploy this version over the running version of the module. Make sure that the tags do not change.
- [x] Set `suppressed_tags = ["Environment"]`. Deploy the module over an already deployed module. Check the plan and make sure that the `Environment` tag is removed.


npalm (approved):
Looks good!, not tested

PR #652, opened 2023-01-09
feat: remove unused SSH keys

### LINKED ISSUE #592: Clean SSH key pairs in AWS
We recently hit the AWS quota for the maximum number of SSH key pairs:

image

Around 4000 of the keys were created by the runner when provisioning machines and left there.

It looks like the runner does not remove the associated key pair when removing a worker.

I would like to be able to:

Not use SSH keys at all (only SSM). I do not know if this is possible, as it looks like docker machine uses those keys for keeping runner-worker communication.

OR

Specify one single key for all workers so they do not get created on demand.

As a temporary fix I created a schedule that cleans all unused CI ssh keys: https://gitlab.com/fluidattacks/universe/-/issues/8300

Thank you so much for all the effort you've put in this module!

kayman-mk:
## Description

Whenever an executor is created it needs an SSH key so the agent is able to contact it. In case the executor is killed this SSH key is not removed. Over time they accumulate and reach the AWS limit. From this point no further SSH keys can be created and thus the module is no longer able to create executors. You won't be able to run any pipeline.

Some users have created their own housekeeping of SSH keys. But this should be done within the module as the keys are created here.

Challenge: Make sure that keys are deleted only which belong to our module. Unfortunately the GitLab Runner does not tag the keys. We use the following logic to determine the keys to delete:

- they are unused
- their name starts with `runner`
- they have the `var.environment`/`var.overrides['name_docker_machine']` somewhere in the name

Solves part of #623 
Closes #592 

## Migrations required

No.

## Verification

- [x] kill an agent manually and check the Cloudwatch logs. The function should log `Removing unused SSH key pairs for environment` and print one line per deleted key. There are no errors in the log.


modules/terminate-agent-hook/iam.tf
```diff
@@ -87,14 +103,56 @@ data "aws_iam_policy_document" "lambda" {
   }
 }
 
+data "aws_iam_policy_document" "ssh_key_housekeeping" {
+  statement {
+    sid = "SSHKeyHousekeepingList"
+
+    effect = "Allow"
+    actions = [
+      "ec2:DescribeKeyPairs"
```
npalm, line 112: Not having much time, but seems DescribeKeyPairs is part of 2 policies, is this correct?

npalm (commented):
Looks good, not time to test

modules/terminate-agent-hook/iam.tf
kayman-mk: removed

modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -87,13 +90,67 @@ def ec2_list(client, **args):
...
     return _terminate_list
 
+
+def remove_unused_ssh_key_pairs(client, executor_name_part):
+    print(json.dumps({
+        "Level": "info",
+        "Message": f"Removing unused SSH key pairs for agent {executor_name_part}"
+    }))
+
+    reservations = client.describe_instances(Filters=[
```
npalm, line 100: Quick quesiton. I am not used to the AWS python sdk. But for example in typescript / javascript API's are paged. So the calls in the function are most like also paged and only returning the first 50 objects. Is this the case?

npalm (changes requested):

modules/terminate-agent-hook/lambda/lambda_function.py
kayman-mk: Oh yes, true. Pagination is done for every 1.000 instances. Guess it's unlikely that you have so many machines running at the same time. But better to fix it now.

kayman-mk:
@npalm Tests completed. 1 SSH key was deleted by the Lambda function.

npalm (approved):
Change looks good to me. Did a bit of testing. As mention in the PR terminate the scenario could be reqplayed by terminating the agent instance. This does not work, the lambda is triggered by the auto scaling group terminating instances. Once terminating the instance by setting the scaling group to 0, the lambda got correctly tirggerd and keys removed.

kayman-mk:
Interesting. I killed the instance via AWS Console and checked the Cloudwatch logs.

PR #653, opened 2023-01-09
feat: cancel spot requests

### LINKED ISSUE #493: cancel-spot-instances.sh fails in Terraform container
This module "npalm/gitlab-runner/aws" is meant to be run in a Terraform container, however script `bin/cancel-spot-instances.sh` doesn't work because it requires bash which doesn't exist in Terraform images. Terraform images only have sh. Besides, the script also requires jq and awscli which don't exist either in Terraform alpine based images.

I get this error:
```
│ Error: local-exec provisioner error
│ 
│   with null_resource.cancel_spot_requests,
│   on main.tf line 142, in resource "null_resource" "cancel_spot_requests":
│  142:   provisioner "local-exec" {
│ 
│ Error running command
│ './.terraform/modules/gitlab-runner/bin/cancel-spot-instances.sh
│ voice-runners': exit status 127. Output: env: can't execute 'bash': No such
│ file or directory
```

kayman-mk:
## Description

Whenever a spot executor is created a spot request is created. Occasionally there might be open spot requests which have not been fulfilled. These spot requests have to be deleted as soon as the agent terminates as no executors are needed.

Challenge: Make sure that spot requests are deleted only which belong to our module. Unfortunately the GitLab Runner does not tag them. We use the following logic for determinination:

- state is open or active
- they are associated to a SSH key which name starts with `runner`
- they have the `var.environment`/`var.overrides['name_docker_machine']` somewhere in the name

Solves part of #623 
Closes #493 

## Migrations required

No.

## Verification

- [x] kill an agent manually and check the Cloudwatch logs. The function should log `Removing open spot requests for environment` and print either `Spot requests deleted` or `No open spot requests found`. There are no errors in the log.


modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -102,19 +163,19 @@ def handler(event, context):
         }))
         try:
             client.terminate_instances(InstanceIds=_terminate_list, DryRun=False)
-            return f"Terminated instances {', '.join(_terminate_list)}"
```
JulianCBC, line 105: Is this removing the log message that confirms that we actually terminated the instances successfully?

Also, would it make sense to split the instance termination out into a separate function too so that the handler is just:

```python
terminate_running_instances(...)
cancel_active_spot_requests(...)
```

kayman-mk: Guess I forgot to replace this with a Json log. Thanks!

kayman-mk: I think we shouldn't create a second lambda as we have to make sure that everything is killed. In case of two functions it could happen, that we kill the EC2s but new spot instances are created because the second lambda function has not been started yet.

I changed the code to terminate the spor requests first and then killing the EC2 instances. This way we should be sure that there is nothing running.

npalm (dismissed):
In general looks good not the time right now to test. Would suggest the following

- Add a module readme to describe shortly what the lambda does
- Add a backlog itme to add unti test.

kayman-mk:
Oh yes, missed to update the `README.md`. It's now done.

Issue for unit tests created.

npalm (approved):

kayman-mk:
merge conflicts fixed

npalm (approved):

PR #656, opened 2023-01-12
ci: add workflow to welcome a new PR

kayman-mk:
## Description

This PR adds a new workflow which posts a comment to every opened PR welcoming the pull request and making some important information available to the author (contribution guidelines, license).

### DIFF
--- a/.github/workflows/pr-opened.yml
+++ b/.github/workflows/pr-opened.yml
@@ -3,6 +3,9 @@ name: PR opened
 
 on:
   pull_request_target:
+    # GITHUB_TOKEN is readonly and the action will fail for Dependabot
+    branches-ignore:
+      - 'dependabot/**'
     types:
       - opened
 
--- a/.github/workflows/pr-opened.yml
+++ b/.github/workflows/pr-opened.yml
@@ -1,5 +1,6 @@
 ---
 name: PR opened
+
 on:
   pull_request_target:
     types:
--- a/.github/workflows/pr-opened.yml
+++ b/.github/workflows/pr-opened.yml
@@ -12,18 +12,12 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - uses: mshick/add-pr-comment@v2
+      - uses: actions/github-script@v6
         with:
-          message: |
-            Hey @${{ github.event.pull_request.user.login }}! 👋
-            
-            Thank you for your contribution to the project. Please refer to the [contribution rules](blob/main/CONTRIBUTING.md) for a quick overview of the process.
-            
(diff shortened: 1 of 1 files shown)


kayman-mk:
Not 100% sure how this behaves if executed from a forked project. But I think `pull_request_target` should do the job.

kayman-mk:
Tested with a forked project. Works like a charm.

.github/workflows/pr-opened.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: PR opened
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+jobs:
+  add-welcome-message:
```
npalm, line 10: Nice job and great idea. I have quickly checked the action you introduced. The action looks nice and filling a need. But any action we add is always a risk as dependency. Both for maintenance and security. What I don't like on the action there is no provenance of the release. No release automation, and no safe guard in the workflows to ensure the generated file in the dir dist (javascript) is based on the source code in the repo. Since we grant the action write access (only PR), we should always be carefull. 

The welcome message we can also do simply with the github script action. This one is maintained at least by GitHub. See https://github.com/actions/github-script#welcome-a-first-time-contributor

Looking forward to you opinion.

npalm (commented):
Nice job, see my remark please. Up to you to make the call

.github/workflows/pr-opened.yml
kayman-mk: Fully agree. Could you please give an example of 

> safe guard in the workflows to ensure the generated file in the dir dist (javascript) is based on the source code in the repo
?

.github/workflows/pr-opened.yml
```diff
@@ -0,0 +1,23 @@
...
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
```
npalm, line 21: Confusing but I think indeed the issue api is the same as for a PR. Did you test it? Maybe good to add a small comment that the API is the same. And the intention is a comment on the PR

npalm (approved):

.github/workflows/pr-opened.yml
kayman-mk: Sure, [tests done](https://github.com/Hapag-Lloyd/test/pull/4).

Yeah, the same API. issues <==> PRs

npalm (approved):

PR #657, opened 2023-01-12
ci: add ChatOps command `/help`

kayman-mk:
## Description

This PR adds ChatOps commands for PRs. This makes it easy for new users to trigger some actions.

Supported commands:
- `/help`: Creates a comment in the PR which mentions one of the maintainers so he can help the user out

Needs #656

### DIFF
--- a/.github/workflows/slash_ops_commands.yml
+++ b/.github/workflows/slash_ops_commands.yml
@@ -1,0 +1,32 @@
+---
+name: Execute SlashOps command
+
+on:
+  repository_dispatch:
+    types: 
+      - help-command
+
+jobs:
+  help-command:
+    name: "SlashOps: Help"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Choose maintainer
+        id: vars
+        run: |
+          maintainers=("npalm" "kayman-mk")
+          
+          RANDOM=$(date +%s)
+          
+          maintainer=${maintainers[ $RANDOM % ${#maintainers[@]} ]}
+
+          echo "maintainer=$maintainer" >> $GITHUB_OUTPUT
+
+      - name: Create comment
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
+          issue-number: ${{ github.event.client_payload.github.payload.issue.number }}
+          body: |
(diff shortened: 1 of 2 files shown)


kayman-mk:
May be the `GITHUB_TOKEN` has to be replaced by a PAT so it works for forks too.

kayman-mk:
Needs #656 so the SlashOps command can be explained there. Otherwise nobody knows them.

kayman-mk:
Future enhancements: Format Terraform code, create Terraform docs

kayman-mk:
Tested with a forked project. `GITHUB_TOKEN` works like a charm.

npalm:
> Future enhancements: Format Terraform code, create Terraform docs. I usually don't have pre-commit running and/or making changes directly in the browser. And the second thing is, that the docs are updated when a release is created. So there is a gap between the merge into `main` and the release creation via PR. Within this time the Terraform docs are wrong/incomplete/...
> 
> Thought about running the Terraform stuff in the feature branch. But if a commit is created I have to rebase locally which is always another step to do. Would be nice to update everything as soon as the PR is ready.

Agree, would be much better to run al the updates before the PR merge. The main branch is a locked branch. Which means we should not commit to the branch directly. For that reason I we moved the docs to the release branch. Assuming frequent releases. Running on the PR before merging had some issue in the past with access to the sources. But agree, doing all those things on the PR is much better.

npalm:
> Future enhancements: Format Terraform code, create Terraform docs. I usually don't have pre-commit running and/or making changes directly in the browser. And the second thing is, that the docs are updated when a release is created. So there is a gap between the merge into `main` and the release creation via PR. Within this time the Terraform docs are wrong/incomplete/...
> 
> Thought about running the Terraform stuff in the feature branch. But if a commit is created I have to rebase locally which is always another step to do. Would be nice to update everything as soon as the PR is ready.

Agree, would be much better to run al the updates before the PR merge. The main branch is a locked branch. Which means we should not commit to the branch directly. For that reason I we moved the docs to the release branch. Assuming frequent releases. Running on the PR before merging had some issue sone past with access to the sources.

> Tested with a forked project. `GITHUB_TOKEN` works like a charm. But I am also a member of the organization owning the base repository.

The GITHUB_TOKEN has only access to the repo for which the CI is running. So in case of the pull_request event that is the fork. And having access to the base should not conflict with this.

npalm (commented):
nice

kayman-mk:
See it in action: https://github.com/kayman-mk/test-own/pull/11

npalm (approved):

PR #671, opened 2023-01-19
fix: upgrade `docker-machine` to support new AWS regions

kayman-mk:
## Description

Upgrades the project to the newest docker-machine to support new AWS regions, e.g. `eu-central-2` (Zurich).

Fixes #660 

## Migrations required

If `docker_machine_version` is set, consider to upgrade to the newest version as well, if you need the new AWS regions.

## Verification

- [ ] deployed to our QA system. Jobs are still processed
- [ ] used docker-machine and created one machine in `eu-central-2` successfully

PR #681, opened 2023-02-02
fix: bump docker machine version due to bug

bwmetcalf:
## Description

This PR updates the docker+machine to the newest version. The fixed bug relevant here is:

> With Docker 23.0.0 release (which happened today), installation of Docker doesn't create an /etc/docker directory
> anymore by default. It needs to be done manually.
> At the same time, Docker Machine's provisioning for some of the supported distributions assumes that it will be there by
> default and starts writing key/certificates files into that directory immediately. Which fails with
>
> `ERROR: tee: /etc/docker/ca.pem: No such file or directory`

This fixes https://gitlab.com/gitlab-org/ci-cd/docker-machine/-/issues/98

## Migrations required

NO

## Verification

Updated in our environment which fixed the issue

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -915,3 +915,4 @@ variable "runner_extra_config" {
   type        = string
   default     = ""
 }
+
--- a/variables.tf
+++ b/variables.tf
@@ -116,7 +116,7 @@ variable "docker_machine_download_url" {
 variable "docker_machine_version" {
   description = "By default docker_machine_download_url is used to set the docker machine version. Version of docker-machine. The version will be ingored once `docker_machine_download_url` is set."
   type        = string
-  default     = "0.16.2-gitlab.15"
+  default     = "0.16.2-gitlab.19"
 }
 
 variable "runners_name" {
--- a/variables.tf
+++ b/variables.tf
@@ -915,4 +915,3 @@ variable "runner_extra_config" {
   type        = string
   default     = ""
 }
-


bwmetcalf:
@npalm I'm not sure what the PR title check is looking for.

bwmetcalf:
> @npalm I'm not sure what the PR title check is looking for.

Fixed.

kayman-mk:
Thanks for this fix!

kayman-mk:
Might supersede #671

kayman-mk (approved):

marvin-w:
This breaks spot instance due to an issue with tagging.

```terraform
docker_machine_version     = "0.16.2-gitlab.15"
docker_machine_options     = ["engine-install-url=https://releases.rancher.com/install-docker/20.10.21.sh"]
```

to fix it.

kayman-mk:
As far as I understood the issues here, the `engine-install-url` was a try to fix the Docker problems. So better remove this again and use the 5.9.1 module with the fixed docker+machine. This should make the `docker_machine_options` obsolete.

marvin-w:
No it only fixes half of the problem. This solutions works only for on demand instances due to https://gitlab.com/gitlab-org/gitlab-runner/-/issues/29213#note_1263789121

kayman-mk:
Let's hope they fix it in a couple of hours.

JulianCBC:
Hm. Looks like the spot instance issues are bigger than can be fixed quickly, there's another fork of docker-machine which has a fix, but it doesn't have the Docker 23.0 fix. Hopefully they can combine forces to get both of these issues sorted in the same fork.

I'm currently using this to work around these issues:

```hcl
  # FIXME: Use non-gitlab version of docker machine to work around spot provisioning issue
  docker_machine_download_url = "https://gitlab.com/cki-project/docker-machine/-/releases/v0.16.2-gitlab.18-cki.1/downloads/docker-machine-Linux-x86_64"

  # FIXME: Use old version of Docker to work around Docker 23 not creating /etc/docker
  docker_machine_options     = ["engine-install-url=https://releases.rancher.com/install-docker/20.10.21.sh"]
```

marvin-w:
> Hm. Looks like the spot instance issues are bigger than can be fixed quickly, there's another fork of docker-machine which has a fix, but it doesn't have the Docker 23.0 fix. Hopefully they can combine forces to get both of these issues sorted in the same fork.
> 
> I'm currently using this to work around these issues:
> 
> ```terraform
>   # FIXME: Use non-gitlab version of docker machine to work around spot provisioning issue
>   docker_machine_download_url = "https://gitlab.com/cki-project/docker-machine/-/releases/v0.16.2-gitlab.18-cki.1/downloads/docker-machine-Linux-x86_64"
> 
>   # FIXME: Use old version of Docker to work around Docker 23 not creating /etc/docker
>   docker_machine_options     = ["engine-install-url=https://releases.rancher.com/install-docker/20.10.21.sh"]
> ```

Yep this also works. Mine from https://github.com/npalm/terraform-aws-gitlab-runner/pull/681#issuecomment-1414344235 worked for us - it just used the previously good docker-machine version and the old (non 23.0) version from the engine.

JulianCBC:
The CKI project seems to be on the ball and has released an update since my last comment. I'll deploy this later today and verify that it fixes the problem.

https://gitlab.com/cki-project/docker-machine/-/releases/v0.16.2-gitlab.19-cki.1

@npalm @kayman-mk as GitLab seems to only be focused on their specific use cases for docker-machine, would it make any sense at all to switch to the CKI maintained version of docker-machine for this project?

kayman-mk:
It's definitely worth to think about it. Any results from your tests?

JulianCBC:
We've been running the `v0.16.2-gitlab.19-cki.1` version of docker-machine on our production runners for a couple of days now without incident, so in my environment (mostly defaults, spot instances) it's a drop-in replacement.

My main reason for asking the question was that the Gitlab fork doesn't seem to be receptive to bug reports that don't relate to Gitlab's specific deployment whereas CKI seems to be doing some level of development to keep the software working, so as we're targeting AWS instead of Google Cloud, it seems like a better fit.

kayman-mk:
Yeah, seems to be a good option. I did some tests as well with the CKI docker+machine (see #696) and proposed #697 to switch to this driver as the problems with the GitLab versions are not tackled.

PR #697, opened 2023-02-15
fix!: switch to docker+machine from CKI project

### LINKED ISSUE #696: Working configurations of docker+machine and GitLab Runner
Due to various problems with the docker+machine, GitLab Runner and Docker v23 release, I took some time to develop working configurations for spot and on-demand instances.

What I have running at the moment (spot and on-demand):

```hcl
version = "5.9.1"

docker_machine_options = ["engine-install-url=https://releases.rancher.com/install-docker/20.10.21.sh"]
docker_machine_version = "0.16.2-gitlab.15"
gitlab_runner_version  = "15.1.1"
```

Upgrading to the newest version of docker+machine will never work as long as you stay with the GitLab fork. You can use the following configuration:

```hcl
version = "5.9.1"

gitlab_runner_version       = "15.8.2"
docker_machine_download_url = "https://arr-cki-prod-docker-machine.s3.amazonaws.com/v0.16.2-gitlab.19-cki.2/docker-machine-Linux-x86_64"

# not needed if you run on-demand instances only; choose according to your setup
runner_instance_spot_price    = "100" # empty string/null does not work here!
docker_machine_spot_price_bid    = "100" # empty string/null does not work here!
```

The spot price bid issue is tackled in #697 

This version should also support the "new" AWS regions `eu-central-2`, ... But I wasn't able to check this.

kayman-mk:
## Description

Due to various problems of the GitLab docker+machine driver (especially with spot instances), this PR switches to the version provided by [CKI](https://gitlab.com/cki-project/docker-machine).

- `amazonec2-spot-price` is removed from the machine options if `on-demand-price` is used as the new driver does not support an empty value.
- GitLab Runner is updated to 15.8.2 due to some important fixes related to Docker v23

Closes #696 

## Migrations required

Yes. But in general it should run without any changes.

Switching to the CKI project means that you have to change the `docker_machine_version` (append a -cki.2 or whatever). As not all docker+machine versions from GitLab are published by CKI you might have to set the `docker_machine_download_url` to get the old version.

We suggest to get rid of the `docker_machine_version` and  `docker_machine_download_url` as long as you don't have any special use case. The default will work in nearly all cases.

## Verification

- [x] run this infrastructure with Gitlab 15.7.7
- [x] set `docker_machine_spot_price_bid = "on-demand-price"`. `config.toml` does not show a `amazonec2-spot-price` entry
- [x] do not set a  `docker_machine_spot_price_bid`. `config.toml` does not show a `amazonec2-spot-price` entry
- [x] set `docker_machine_spot_price_bid = "2.50"`. `config.toml` shows a `amazonec2-spot-price=2.50` entry
- [x] use on-demand instances only. Set `runners_request_spot_instance = false`.

Post condition for all tests: Runners are working and processing jobs.


kayman-mk:
This mitigates all problems due to the buggy GitLab docker+machine. See #696

JulianCBC:
Looks good to me, I'll be migrating my instance to this configuration shortly to verify.

JulianCBC:
> Looks good to me, I'll be migrating my instance to this configuration shortly to verify.

Yeah, deployed and ran a few jobs and it all works.

You can consider this verification and an approval from me!

kayman-mk:
All tests were successful.

kayman-mk:
@npalm Any chance to get this merged?

variables.tf
```diff
@@ -110,7 +110,7 @@ variable "docker_machine_spot_price_bid" {
 variable "docker_machine_download_url" {
   description = "(Optional) By default the module will use `docker_machine_version` to download the GitLab mantained version of Docker Machine. Alternative you can set this property to download location of the distribution of for the OS. See also https://docs.gitlab.com/runner/executors/docker_machine.html#install"
```
npalm, line 111: Can you update this description, maybe refer to https://gitlab.com/cki-project/docker-machine and quickly explain why we switch.

npalm (dismissed):

npalm:
@kayman-mk looks good, a small comment. Please go ahead with merging.

kayman-mk:
@npalm Overlooked a hardcoded url in the template and fixed that. Could you please approve again.

I think it is a major change as you have to change the `docker_machine_version` (append a `-cki.2` or whatever). As not all docker+machine versions from GitLab are published by CKI you might have to set the `docker_machine_download_url` to get the old version.
Yes, in general, you shouldn't overwrite theses variables as the default should be good enough. But users might run into trouble here. So I propose a major change which is more likely to be seen.

variables.tf
kayman-mk: Fixed the description but didn't mentioned any reason there as I think it is too much information for the variable description. I put some information into the description of this PR.

kayman-mk:
Installed in my production environment. First pipeline ran without problems. Let's wait until Monday and process another 1000 jobs to be safe.

npalm (approved):

PR #704, opened 2023-02-23
fix: install gitlab-runner after docker+machine driver

### LINKED ISSUE #703: [docker+machine] Gitlab Runner is started before installation is finished
# Describe the bug

## Summary

A premature start of the `gitlab-runner` service causes the Runner Manager to pick up jobs before the machine is actually ready to pick up jobs. This causes the jobs to get stuck: they seems to be executing on GitLab but they get stuck in pergatory because we then restart the `gitlab-runner` service.

## Walk along

When the Runner Manager is starting and the `userdata` scripts are executing, GitLab Runner is started before the actual installation is finished:

image

This causes the Runner Manager to pick up jobs and trying to create Machines, but it first fails because `docker-machine` executable is not yet installed because that happens later in the `userdata` script as can be seen below:

image

Here you can see that pending jobs have already been picked up while we are still creating a dummy machine:

image

Later in the `userdata` script we call `service gitlab-runner restart`:

image

Then the Runner manager is back up again but we can see that jobs that have been picked up are lost (`builds: "0"):

image

This causes the job that has been picked to get stuck:

image

# To Reproduce

1. Have jobs that are in a `stuck` state: no Runner can pick them up.
2. Provision a Docker Machine Runner Manager that would be able to pick up those `stuck` jobs
3. The jobs will get picked up by the Runner Manager prematurely
4. The jobs will go from `pending` to `running` but they will actually never execute

# Expected behavior

The `gitlab-runner` service should not start prematurely.

# Additional context

When we install the `gitlab-runner` package here:

https://github.com/npalm/terraform-aws-gitlab-runner/blob/1f75ba4a20a8b6ad0fb894c68005544469a10475/template/gitlab-runner.tpl#L82-L86

as a final step the `gitlab-runner` service actually get registered in systemd AND the service is started up. I think this is where the `gitlab-runner` service is actually started prematurely even before we have `docker-machine` installed.

I think a possible solution would be:

* Install all prerequisites (`docker-machine` + creating the dummy machine) before installing the actual `gitlab-runner` package
* Switch to a manual installation where we have control over when the `gitlab-runner` service gets started (main solution). See: https://docs.gitlab.com/runner/install/linux-manually.html#using-binary-file. In step 4 you can see you have to manually start the service, giving us more control.

Thanks for coming to my TED talk 😄

kayman-mk:
## Description

As pointed out in #703 the GitLab Runner is installed and started before the docker+machine driver. But as the docker+machine driver needs some time for a complete installation (certificates, ...) it might happen that the GitLab Runner already tries to process jobs from GitLab. These jobs will fail.

This PR changes the installation order of the GitLab Runner project and the docker+machine driver. The driver is installed first, the GitLab Runner second. This avoids running a non-working GitLab Runner setup.

⚠️ The restart of the GitLab Runner after executing user's post install script has been removed as a restart could possibly loose some jobs being processed by the GitLab Runner. If this is needed by your script, you have to add it there and be aware, that this might cause problems with the first jobs being processed.

Closes #703 

## Verification

- [x] start a job in GitLab then start the only Runner processing this job. The job should succeed.
- [x] there are no new errors during startup regarding docker, docker-machine and GitLab Runner


template/gitlab-runner.tpl
```diff
@@ -176,6 +170,12 @@ chkconfig --add remove_gitlab_registration
...
 
+if ! ( rpm -q gitlab-runner >/dev/null )
+then
+  curl --fail --retry 6 -L https://packages.gitlab.com/install/repositories/runner/gitlab-runner/script.rpm.sh | bash
+  yum install gitlab-runner-${gitlab_runner_version} -y
+fi
+
 ${post_install}
 
 service gitlab-runner restart
```
tmeijn, line 181: suggestion: change this to (a redundant) `gitlab-runner start`. A restart might cause the runner to lose the jobs it has picked up already.

```suggestion
service gitlab-runner start
```

kayman-mk: Let's go one step further: If it's redundant, we can remove it as it doesn't do anything. If it is there due to the `${post_install}`: As this script is under user control it should also start/restart, ... the Runner if needed.

Just tested this PR and checked the logs. After the `yum` command and before any other command in the startup script, I read: `Feb 23 07:59:31 ip-100-64-24-219 systemd: Started GitLab Runner.`

tmeijn: I'm just a bit afraid of [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence) 😅. Especially since I just started using this module.

tmeijn: Just tested it this way, notice the sleep as stand in for long post_install command.

```shell
if ! ( rpm -q gitlab-runner >/dev/null )
then
  curl --fail --retry 6 -L https://packages.gitlab.com/install/repositories/runner/gitlab-runner/script.rpm.sh | bash
  yum install gitlab-runner-${gitlab_runner_version} -y
fi

${post_install}

sleep 60

service gitlab-runner restart
```

Resulted in job error, not loss. But still, seems very bad.

image

kayman-mk: Also true. But as far as I can see, the commit is from 2017, first installing the GitLab Runner, then the docker+machine and then restarting the GitLab Runner.

As we know from the new logs, the GitLab Runner is up and running. Jobs are processed. So I am fine with this change. @npalm Do you remember why this restart is there?

kayman-mk: Checked the docs for [manual installation](https://docs.gitlab.com/runner/install/linux-manually.html#using-binary-file) of the GitLab Runner. They install the Runner and starts it. That's all. No second start/restart at all.

npalm (approved):

npalm:
I have no issues with the current main, but swapping the logic is fine to me. @kayman-mk up to you to merge (or not)

PR #707, opened 2023-02-24
fix: set correct lifecycle prefix for shared cache

### LINKED ISSUE #539: Is the cache cleanup lifecycle working?
I think there are some issues with cleaning up the S3 cache.

# S3 Lifecycle Prefix
After installing a fresh Gitlab Runner and S3 cache I noticed that the prefix within the S3 cache is `project` and not `runner`. But this is the default setting when a new cache is created. This was also mentioned in #200 
This could be 
- a configuration issue of the user (but I didn't found any special configuration on my side)
- a change within the Gitlab Runner project (renamed cache directory) which we didn't notice

https://github.com/npalm/terraform-aws-gitlab-runner/blob/5b6e342467557ca3e5c6567ee01fb3b0d2ccc5de/modules/cache/variables.tf#L59-L63

# S3 Lifecycle
The lifecycle rule touches files within the bucket not its content. This means that objects which are not updated (no longer in use, e.g. caches from deleted Gitlab projects) are deleted. But I think this does not happen very often as the majority of all projects in Gitlab is never deleted or archived.

What I see in my cache is that it constantly grows as old objects within the cache zip file are never deleted. This results in higher storage costs and slower pipelines as the amount of data grows over the time.

# Suggestion
1. Change the default setting of the lifecycle prefix as mentioned above to activate the clean up.
2. Clean the cache files, e.g. unzip the cache, delete old files, zip and upload the cache again.

What do you think?

tmeijn:
## Description

Sets the lifecycle prefix correctly based on whether the cache is shared or not.
For more context see validation in #539.

## Migrations required

NO - users of a shared cache should finally get a correct lifecycle prefix! 🚀 

## Verification

```hcl
cache_shared = false
```

Result:

image

```hcl
cache_shared = true
```

Result:

image

Closes #539

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -327,8 +327,10 @@ module "cache" {
   cache_bucket_set_random_suffix       = var.cache_bucket_set_random_suffix
   cache_bucket_versioning              = var.cache_bucket_versioning
   cache_expiration_days                = var.cache_expiration_days
+  cache_lifecycle_prefix               = var.cache_shared ? "project/" : "runner/"
   cache_logging_bucket                 = var.cache_logging_bucket
   cache_logging_bucket_prefix          = var.cache_logging_bucket_prefix
+
 
   kms_key_id = local.kms_key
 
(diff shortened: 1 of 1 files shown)


kayman-mk (approved):

kayman-mk:
Thanks for this fix.

PR #752, opened 2023-03-16
feat!: allow to set all docker options for the Executor

kayman-mk:
## Description

Adds a new variable `runners_docker_options` which holds all values for the `[runners.docker]` section and makes the single variables
- `runners_image`
- `runners_privileged`
- `runners_disable_cache`
- `runners_additional_volumes`
- `runners_shm_size`
- `runners_docker_runtime`
- `runners_helper_image`
- `runners_pull_policy`

obsolete.

## Migrations required

Yes, as the minimum Terraform version is 1.3.0 to support optional block variables with defaults.

A migration script is provided to restructure the variables. See `/migrations/migrate-to-7-0-0.sh`. Attention Mac users: The script will not work out of the box as the `sed` implementation is different. Use a Docker container with Alpine or Ubuntu to run the script.

```hcl
module "gitlab_ci_runner" {
  ...
  runners_docker_options {
    # set whatever is necessary
  }
```

## Verification

- [x] Use current configuration and ensure that the `config.toml` remains unchanged
- [x] Set all new block variables and ensure that the `config.toml` is valid (use `gitlab-runner verify)
- [x] Check that the default settings with Terraform 1.3 work as expected
- [x] Verify all docker settings against the documentation to ensure correct names

The runner starts in both cases and is available in Gitlab. No example tested but used our active configuration at Hapag-Lloyd.


kayman-mk:
Just to check the branch after messed up with `Hapag-Lloyd:kayma/network-mode`. See #511

PR #754, opened 2023-03-17
feat: show `config.toml` and user data in Terraform plan

tmeijn:
### What

Adds a new variable `show_user_data_in_plan` which is `false` by default (there could be sensitive data leaked).

* Writes the rendered `config.toml` to a file
* Writes the agent's user data to a file.

### Why

* By writing the rendered `config.toml` to a file we get an actual diff of the `config.toml` when we change values in our Terraform plan. In addition this would help us with the refactoring work were currently doing.
* Even better ability to see the impact of changes we make.

## Migrations required

NO

## Verification

`show_user_data_in_plan = true`

image

Example of `config.toml` showing a diff in plan:

image


kayman-mk:
Appreciated. Makes it easier to check for changes without deploying the module. Simplifies the development process.

.mega-linter.yml
```diff
@@ -2,6 +2,8 @@
 DISABLE_LINTERS:
   # Has some problems reading referenced policy files. I created a separate workflow for TfLint as it creates valuable output.
   - TERRAFORM_TFLINT
+  # Seems very, very slow compared to other linters.
+  - TERRAFORM_KICS
```
kayman-mk, line 6: issue: This is not related to this PR. I have seen bad execution times of the KICS linter as well. The output is valuable, but we have to optimize the execution time. Idea is to run it in parallel with the Megalinter. This saves up to 3 minutes (the time it needs to pull the Megalinter image).

To be done in a separate PR.

locals.tf
```diff
@@ -91,3 +91,14 @@ locals {
...
+
+resource "local_file" "config_toml" {
+  content  = local.template_runner_config
+  filename = "${path.module}/debug/runner_config.toml"
+}
+
+resource "local_file" "user_data" {
+  count    = var.enable_user_data_diffing ? 1 : 0
+  content  = nonsensitive(local.template_user_data)
+  filename = "${path.module}/debug/user_data.sh"
```
kayman-mk, line 103: suggestion: Include the `/debug` folder in `.gitignore` so we don't check in the files by accident.

variables.tf
```diff
@@ -948,3 +948,9 @@ variable "runner_extra_config" {
   type        = string
   default     = ""
 }
+
+variable "enable_user_data_diffing" {
```
kayman-mk, line 952: suggestion: Can we rethink the name? I have no alternative at hand, but `diffing` sounds a little bit strange.

kayman-mk (changes requested):

kayman-mk:
See #755 for KICS issue

variables.tf
tmeijn: WYDT about `show_user_data_in_plan` ?

tmeijn: Well, depends on how you look at it. The objective here is to show what changes are actually in those files. To do that I have to write them to somewhere so that we get a diff from `local_file` as shown in the screenshot in the PR body. For me personally writing the file is a clever workaround, but was not the main goal, insofar that if you could not write it, I wouldn't have done it initially.

tmeijn:
@kayman-mk apart from the PR linter error, this one is ready for re-review.

kayman-mk (approved):
Good job! Makes it very easy to check for configuration changes.

kayman-mk (approved):

PR #757, opened 2023-03-20
refactor!: rename variables and prefix with `runner_manager`, `runner`, `runner_worker` and global scope

### LINKED ISSUE #575: Minor: Standardise variable names
Some variable names could be standarised.

First example, all of the below relate to cloudwatch log groups:
`log_group_name`
`enable_cloudwatch_logging`
`create_cloudwatch_log_group`
`cloudwatch_logging_retention_in_days`

Probably best to leave this issue for a little while to collect and discuss other cases of standardisation and then to a single PR to resolve.

It may be helpful to also compile a "migration list" so that users can easily upgrade between different named variables if there are many.

Note, changing these would be breaking changes.

kayman-mk:
## Description

Moves all variables to a new block:
- `runner_manager` in case it configures the "main" process which sets the defaults for all runners
- `runner` in case it configures the runner created by the runner manager
- `runner_worker` in case it configures the docker/docker+machine
- or leave it as it is, if it is a global scope, e.g. common tags, the environment, ...

Closes #575 

## Migrations required

Yes. Use the provided script to migrate to version 7.0

## Verification


kayman-mk:
Did some experiments today. I think, it is not a good idea to use blocks.

They look good and structure the variables into groups. Blocks can be grouped in other blocks, which gives a good strucuture for the variables.

But the drawback is, that I can't document these variables in a way to get support from my IDE.

So I will go with the following:
```hcl
module "gitlab_runner" {
 agent_var_1 = 1
 agent_var_2 = 2

 executor_var_ 1 = 1
}
```

instead of

```hcl
module "gitlab_runner" {
 agent {
  var_1 = 1
  var_2 = 2
 }

 executor {
  var_1 = 1
 }
}
```

tmeijn:
I'm a bit confused...

```hcl
module "gitlab_runner" {
 agent {
  var_1 = 1
  var_2 = 2
 }

 executor {
  var_1 = 1
 }
}
```
AFAIK this syntax is not possible for modules and can only be defined for provider resources and datasources.

kayman-mk:
You are right. There is a `=` missing, `agent = {...}`. But anyway, I will not use it due to the documentation problems.

tmeijn:
FYI: https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/3996/diffs#note_1342739227 and https://gitlab.com/gitlab-org/gitlab/-/issues/398139

kayman-mk:
@tmeijn Counted 126 variables in the module so far. I will follow your suggestion and group them into a map despite the documentation issues I had. I think it is worth to reference the GitLab/Docker Machine documentation and not to repeat everything here again.

kayman-mk:
Closed in favor of #810 which includes the renaming and restructuring.

PR #768, opened 2023-03-24
fix: disable outputting config.toml by default

tmeijn:
## Description

In #754 I unfortunately made a major oversight: the `local_file` doesn't persist between machines (duh 🤦🏾 ). So for example in a CI/CD pipeline the file is not present so it wants to render the whole file as new every time. This defeats the main purpose that I had intended for this feature... Also if the pipeline applies the plan, there will be a perpetual diff, even if you save the file in repository.

I still think this is useful locally for debugging purposes, but therefore I propose we disable this mechanism by default. In addition, when these variables are enabled I changed the location to the Root module for easier access if you want to debug the rendered files.

## Migrations required

YES/NO - Feature is quite new, if someone wants to use keep this as introduced they now need to explicitly enable it.

## Verification

```hcl
output_runner_config_to_file = true
output_user_data_to_file     = true
```

Results:

image


kayman-mk:
Do we really need two variables here? It's a debug feature.

npalm:
> Do we really need two variables here? It's a debug feature.

Agree we should not introduce two variables. And since the vars are both mentioned for debug. I would suggest the following:

- introduce an object which encloses both vars
- add a condtion to the current variable to mention it is moved

kayman-mk:
@npalm @tmeijn Are we good to go here now?

npalm:
Will be able to check open PRs Sunday or Monday

kayman-mk (approved):

kayman-mk:
@npalm As you mentioned to check the open PRs: Shall we merge?

PR #769, opened 2023-03-27
ci: do not check the links in `CHANGELOG.md`

kayman-mk:
## Description

`CHANGELOG.md` is an auto-generated file. It makes no sense to check whether the links are correct or not as we shouldn't change this file.

## Migrations required

No

## Verification

- [x] Ensure the `CHANGELOG.md` is no longer checked


kayman-mk:
@npalm I will tackle the `checkov` issues in another PR. Seems that we should fix the KICS version and do not use `latest`.

npalm (approved):

Issue #474, opened 2022-04-28
capacity-not-available when using spot instances


schealex:
Hi everyone,

I'm not really sure if that is an AWS issue or maybe a config issue on our side but right now we are having huge troubles getting runners up for our jobs. It worked fine for a couple month but since a few days we have hours of downtime due to not getting spot instances as requested by the runner:

![image](https://user-images.githubusercontent.com/3441348/165766045-3930cb28-73b1-4f1e-b3fa-9849e60c9b5d.png)

Is this something we can address on AWS? Is there something we can do to prevent this?

If there is not is there a way to do a mixed configuration where you have at least 1 on-demand instance and up to X spot instances - so that we can make sure that there is always at least one instance to handle jobs?

Regards
Alex

kayman-mk:
I deployed the module two times: One is using spot instances and the other is running on on-demand instances (for Terraform jobs). Just copy your module definition and change the two or three variables.

rgarrigue:
There's #249 on this topic.

@kayman-mk about to do the same, I'm wondering about the behavior : if I've the spot one with autoscaling idle count > 1 and the on-demand with autoscaling idle count = 0, will it favor spot ones until eventually there can't be any spot, then it spins up on-demand ?

kayman-mk:
I don't think so. I guess it's the runner contacting your GitLab instance next which receives the job. But you can't control that. 

Would be a nice feature to use spot instances first and on-demand instances if no spot instances are available. But this didn't happen here.

erickfaustino:
Hi, @kayman-mk
Could you guide me on how to disable spot instances?

I've unset the  spot bid price variable and also set in `docker_machine_options` not to request spot instances, but it doesn't work and keeps spawning spot instances

```
module "gitlab-runner" {
  source  = "npalm/gitlab-runner/aws"
  version = "4.41.1"

  asg_terminate_lifecycle_hook_create = false
  aws_region                          = "us-east-1"
  docker_machine_options              = ["amazonec2-instance-type=${var.runner_instance_type}", "amazonec2-request-spot-instance=false", "amazonec2-use-ebs-optimized-instance=true", "amazonec2-root-size=${var.runner_root_disk_size}", "amazonec2-ami=ami-0c4f7023847b90238"]
  docker_machine_version              = "0.16.2-gitlab.12"
  enable_asg_recreation               = true
  enable_cloudwatch_logging           = false
  enable_docker_machine_ssm_access    = true
[...]

Issue #435, opened 2022-01-20
EC2 created from launch configuration keeps terminating.


cmbailey87:
Current using  version = "4.25.0"
and I have my ec2 that is created from the launch group to be a spot instance request. 
https://github.com/npalm/terraform-aws-gitlab-runner/issues/344 

Something that has started recently is the EC2 created from the launch group( the EC2 that deploys the docker machines ) will occasionally, once a week, terminate. 
I have to terminate all the spot requests ( because some show as active and fulfilled ) but there isnt an EC2 running...
OR I have to redeploy the code to the one failing ENV.

I have deployed this same code to multiple ENVs...but only one of the ENVs has this happening. All in the same region.
All ENVs have the same spot requests.

has anyone ran into this issue???
I have my instances set on a schedule and Idle time... could this be the culprit?
if this fixed in a newer version???

thanks in advance.

cmbailey87:
![image](https://user-images.githubusercontent.com/26728086/150378969-27fd809a-14f5-47ea-8df8-14ceef690817.png)
![image](https://user-images.githubusercontent.com/26728086/150382829-4f2bdf4e-d639-483e-a9e8-42cf9356b55c.png)


think i found my issue...i am running into a capacity issues for this account... but only this account.

npalm:
Would suggest to check your AWS limits indeed, please can you update / close the issue when resolved. Thanks for taking the time for sharing!

npalm:
still having this issue?

cmbailey87:
Sorry @npalm , I was out of the office for an extended period...
The issue was resolved, I changed the  EC2 type to a medium , instead of a micro.. The micros seem to be very popular and not alot of the instances were available as spot instances...

Mediums on the other are a plenty.

I will close out the issue. thanks again.

PR #777, opened 2023-03-31
feat(spotfleet): add supports spot fleets for spot instances allowing us to use multiple instance types and AZs

Kadeux:
## Description

The [fork](https://gitlab.com/cki-project/docker-machine/-/tree/v0.16.2-gitlab.19-cki.2) of docker-machine used in this module has released a version that supports [AWS Spot Fleet](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-fleet.html) 3 months ago.

In our case, spot fleets are useful when using spot instances because it fixes the issue when there is no capacity available for a certain instance type in a certain AZ. With spot fleets, we can use multiple instance types and AZ so when there is no capacity on one instance type, AWS will take the next instance type and so on.

This fixes issues [#76](https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/76) [#435](https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/435) [#77](https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/77) and [#474](https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/474)

## Migrations required
NO our solution is backward compatible so if you don't plan to use spot fleets, no migration is required.
YES if you want to use spot fleets you will have to do the following changes:
`docker_machine_instance_type` to `docker_machine_instance_types`
`subnet_id` to `subnet_ids`

You also have to put these 2 new parameters:
`use_fleet` and `key_pair_name` 

### Important
We have seen that the [fork](https://gitlab.com/cki-project/docker-machine/-/tree/v0.16.2-gitlab.19-cki.2) of docker-machine this module is using consume more RAM using spot fleets.
For comparison, if you launch 50 machines in the same time, it consumes ~1.2GB of RAM. In our case, we had to change the `instance_type` of the runner from `t3.micro` to `t3.small`.

## Verification

```hcl
  docker_machine_instance_types = ["r5a.xlarge", "r5.xlarge", "r5ad.xlarge", "r5d.xlarge", "m5a.2xlarge"]
  use_fleet                     = true
  key_pair_name                 = "<key-pair-name>"
```
We are using this solution in production for a week now with 150+ developers and over 20 000 jobs a day without any issue for now.


variables.tf
```diff
@@ -31,6 +31,12 @@ variable "subnet_id" {
   default     = "" # TODO remove as soon as subnet_id_runners and subnet_ids_gitlab_runner are gone. Variable is mandatory now.
 }
 
+variable "subnet_ids" {
```
npalm, line 34: Just a quick response, need to dig futher in the details.

This variable seems similar to `subnet_id`. The variable you introduced here was used in the pas. But since the agent is is only deployed once it was refactored to single subnet. 

The subnets here are meant for the runner executors and not for the agent as far I can see. Maybe we should think about a better name to link better to the runner executors. 

In the comment you mention the firs tis used. I think this is not right. In another module I maintain I use the spot fleet API as well. Here we provide the api a list of subnets, instance types and price strategy. The fleet api creates instances matching this requirements.

Kadeux: I think my comment is not explicit enough, all the subnets in the variable are used for agents, but only the first one is used for the runner itself since it can only be in one subnet. Maybe the best strategy here is to used one variable for the subnet of the runner.

template/gitlab-runner.tftpl
```diff
@@ -48,6 +48,14 @@ fi
 
 sed -i.bak s/__REPLACED_BY_USER_DATA__/$token/g /etc/gitlab-runner/config.toml
 
+if [[ "${use_fleet}" == "true" ]]
```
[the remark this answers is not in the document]
Kadeux: Yes I think I will create a SSH key pair by using `tls_private_key` instead of asking a public key form the user.

Kadeux:
> > You also have to put these 4 new parameters:
> > `use_fleet` `public_key` `secure_parameter_store_runner_private_key` and `key_pair_name`
> 
> Would it be a good idea to have the module create these?

I finally found a way to create the SSH key without asking one from the user.

README.md
```diff
@@ -369,6 +369,47 @@ module "runner" {
...
+
+  aws_region  = "eu-west-3"
+  environment = "spot-runners"
+
+  vpc_id                   = module.vpc.vpc_id
+  subnet_ids               = module.vpc.private_subnets
+
+  docker_machine_instance_types             = ["t3a.medium", "t3.medium", "t2.medium"]
+  use_fleet                                 = true
+  key_pair_name                             = "<key_pair_name>
```
marvin-w, line 393:
```suggestion
  key_pair_name                             = "<key_pair_name>"
```

README.md
```diff
@@ -369,6 +369,47 @@ module "runner" {
...
+Since spot instances can be taken over by AWS depending on the instance type and AZ you are using, you may want multiple instances types in multiple AZs. This is where spot fleets come in, when there is no capacity on one instance type and one AZ, AWS will take the next instance type and so on. This update has been possible since the [fork](https://gitlab.com/cki-project/docker-machine/-/tree/v0.16.2-gitlab.19-cki.2) of docker-machine supports spot fleets.
+
+We have seen that the [fork](https://gitlab.com/cki-project/docker-machine/-/tree/v0.16.2-gitlab.19-cki.2) of docker-machine this module is using consume more RAM using spot fleets.
+For comparison, if you launch 50 machines in the same time, it consumes ~1.2GB of RAM. In our case, we had to change the `instance_type` of the runner from `t3.micro` to `t3.small`.
+
+#### Configuration example
+```hcl
+module "runner" {
+  # https://registry.terraform.io/modules/npalm/gitlab-runner/aws/
+  source  = "npalm/gitlab-runner/aws"
```
marvin-w, line 383: Does this need to be changed to the new repo path?

Kadeux: This repo path is still working since it is not the path for the GitHub repo, it is the path for this terraform [repo](https://registry.terraform.io/modules/npalm/gitlab-runner/aws/latest)

marvin-w:
I have this running on production for a week now without any problem. Nice work!

variables.tf
```diff
@@ -43,6 +49,12 @@ variable "metrics_autoscaling" {
   default     = null
 }
 
+variable "key_pair_name" {
+  description = "The name of the key pair used by the runner to connect to the docker-machine executors."
+  type        = string
+  default     = "fleet-key"
```
kayman-mk, line 55: issue: All resources should be prefixed by `var.environment`, so better to calculate the default somewhere else. Or maybe the variable is not needed at all and can be hard-coded?

kayman-mk (changes requested):
Highly appreciated as we had some problems in the last week to order spot instances. Could you please check the two minors?

kayman-mk:
@Kadeux Could you please have a look at my commits?

Renamed resources/variables to make clear that we deal with fleets and set the old default for the `key_pair_name` again. Otherwise the variable becomes mandatory and the resource is now prefixed by `var.environment`.

kayman-mk:
Will give it a try tomorrow morning in my test environment. I am very excited about this change.

kayman-mk:
@Kadeux The docs are updated automatically during the release cycle. So don't worry.

kayman-mk:
Good job. Checked with/without fleet support. Both use cases are working.

kayman-mk (approved):

Kadeux:
I believe the removal of `nonsensitive` in the output of `runner_user_data` was a mistake. In fact we are printing the SSH private key. I think we should add `nonsensitive`.

kayman-mk:
I think so do and we have to check the Cloudwatch logs, .... for similar problems. Yours will be fixed with #821

PR #778, opened 2023-03-31
feat: add an IAM policy to grant the runner access to the KMS key

jasonjanderson:
## Description

Fixes some issues with accessing the S3 bucket when the bucket has KMS encryption enabled.

Allowing  the runner role the ability to use the KMS key. We mainly ran into issue with this around access to the caching S3 bucket. Without access to the KMS key, the runners can't pull from or push to the S3 bucket created by this project.

In case `var.enable_kms` is set to `true`, the module automatically adds a policy to access the KMS key.

## Migrations required

No

## Verification
Custom:
```
module "runner" {
  source      = "github.com/jasonjanderson/terraform-aws-gitlab-runner?ref=kms"
  aws_region  = data.aws_region.current.name
  environment = "terraform"

  #   ssh_public_key = local_file.public_ssh_key.content

  vpc_id                   = module.vpc.vpc_id
  subnet_ids_gitlab_runner = module.vpc.private_subnets
  subnet_id_runners        = element(module.vpc.private_subnets, 0)

  runners_name       = "terraform"
  runners_gitlab_url = "https://gitlab.com"

    registration_token = data.gitlab_project.init-register.runners_token
    tag_list           = "terraform"
    description        = data.gitlab_project.init-register.name
    locked_to_project  = "true"
    run_untagged       = "false"
    maximum_timeout    = "3600"
  }


  docker_machine_spot_price_bid = "on-demand-price"
  docker_machine_instance_type  = "t3.large"

  enable_kms     = true
  enable_cloudwatch_logging            = true
  cloudwatch_logging_retention_in_days = 14
  agent_tags                           = local.terraform_runner_tags
  runner_tags                          = local.terraform_runner_tags
  docker_machine_iam_policy_arns       = [aws_iam_policy.terraform_runner.arn]
  runners_add_dind_volumes             = true
  runners_executor                     = "docker+machine"
  runners_monitoring                   = true
  runners_request_concurrency          = 10
  runners_machine_autoscaling = [{
    periods    = ["\"* * 8-19 * * mon-fri *\""]
    idle_count = 1
    idle_time  = 1800
    timezone   = "America/Chicago"
  }]
}
```


policies/instance-kms-policy.json
```diff
@@ -0,0 +1,18 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Action": [
+                "kms:Encrypt*",
+                "kms:Decrypt*",
+                "kms:ReEncrypt*",
+                "kms:GenerateDataKey*",
+                "kms:Describe*"
```
kayman-mk, line 10: suggestion: I think `Decrypt` and `GenerateDataKey` should be sufficient according to [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html). Could you please check this?

kayman-mk (changes requested):
Good job! Thanks for implementing the KMS policy in the module. Makes the module easier to use.

policies/instance-kms-policy.json
jasonjanderson: Updated the policy to only include `Decrypt` and `GenerateDataKey`. Validated that the change is working correctly. 

If the `SecureString` SSM parameters are ever updated to use the CMK, `kms:Encrypt` will have to be added to this policy.

kayman-mk:
@npalm Any chance to test this change (have no access to a computer right now)? It looks good to me. I think we have to create a release note so users can remove their own KMS policy.

npalm:
> @npalm Any chance to test this change (have no access to a computer right now)? It looks good to me. I think we have to create a release note so users can remove their own KMS policy.

will do my best to test it today or tomorrow. Thx for the ping

main.tf
```diff
@@ -370,6 +370,32 @@ resource "aws_iam_role" "instance" {
   tags = merge(local.tags, var.role_tags)
 }
 
+################################################################################
+### Policy for the instance to use the KMS key
+################################################################################
+resource "aws_iam_policy" "instance_kms_policy" {
+  count = var.enable_kms ? 1 : 0
```
npalm, line 377: Introduction of the count let the policy created on line 472 fails when "enable_kms" is set

```
│ Error: Invalid count argument
│
│   on ../../main.tf line 472, in resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance":
│  472:   count = var.runners_executor == "docker+machine" ? (var.cache_bucket["create"] || lookup(var.cache_bucket, "policy", "") != "" ? 1 : 0) : 0
│
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To
│ work around this, use the -target argument to first apply only the resources that the count depends on.
╵
╷
```

main.tf
```diff
@@ -370,6 +370,32 @@ resource "aws_iam_role" "instance" {
...
+
+  name        = "${local.name_iam_objects}-kms"
+  path        = "/"
+  description = "Allow runner instance the ability to use the KMS key."
+  policy = templatefile("${path.module}/policies/instance-kms-policy.json",
+    {
+      kms_key_arn = var.enable_kms && var.kms_key_id == "" ? aws_kms_key.default[0].arn : "arn:aws:kms:${var.aws_region}:${data.aws_caller_identity.current.account_id}:key/${var.kms_key_id}"
```
npalm, line 384: I doubt this is correct. On other places in the module var.kms_id is pointing to the ARN as far I can see.

npalm (changes requested):
@jasonjanderson thx for the contribution.

- Tested with a deployment without KMS key, all fine
- When setting `enable_kms` and not provide a manged key the module failes. See inline comment.

Dit not have the time to explore solution right now.

main.tf
jasonjanderson: Fixed this. var.kms_key_id is now assumed to be the full ARN. Checked a few places where var.kms_key_id is used and they accept the ARN. Some accept the ARN or the key ID.

main.tf
kayman-mk: Didn't see this behavior here. Switched from 6.2.0 to this version without any problems. Have `enable_kms` set and tried with/without `kms_key_id`.

main.tf
kayman-mk: Yeah, horrible in the provider. I updated the variable description with this finding.

main.tf
kayman-mk: @npalm How to proceed here? Any chance to reproduce it? To me it seems not related to this change.

Haven't we removed these `lookup`s in the past? May be we forgot this one.

npalm: My bad, ran everything again without a problem

npalm (approved):
@jasonjanderson thanks for your work, all looks good. @kayman-mk for me the PR is ready.

kayman-mk (approved):

PR #798, opened 2023-04-13
chore: replace references pointing to `npalm/` by `cattle-ops/`

kayman-mk:
## Description

This PR removes all references pointing at `npalm/` resources with `cattle-ops/` as the repository was moved some time ago and everything should point to the new destination.


kayman-mk:
The Megalinter will succeed as soon as #799 is merged.

The first commit is the "real" one. All others are dealing with linter issues.

npalm (approved):

PR #723, opened 2023-03-02
feat!: restructure module variables

kayman-mk:
## Description

Restructures all input variables of the agent and executor.

Closes #467 (add network_mode to docker.runner config)
Closes #513 (wait_for_services_timeout paramete)
Closes #819

## Migrations required

YES. Check the script in `/migrations/migrate-to-7-0-0.sh`.

PR #810, opened 2023-04-20
refactor!: group variables for better overview

kayman-mk:
## Description

Groups variables into objects to
- reduce the number of variables (currently 118)
- to gain a better overview of all configuration settings

Creates new groups of variables:
- `runner_manager` in case it configures the "main" process which sets the defaults for all runners
- `runner` in case it configures the runner created by the runner manager
- `runner_worker` in case it configures the docker/docker+machine or leave it as it is, if it is a global scope, e.g. common tags, the environment, ...

## Migrations required

Yes and a script is provided to do that. It covers 98% of all migrations (see migrations/migrate-to-7-0-0.sh)

## Verification


kayman-mk:
Branch is based on #757 Commits beginning from here are the relevant ones.

kayman-mk:
@npalm This one is ready now. It rewrites the complete `variables.tf` and restructures it. Could you please do a review here? Extensive tests still to be done. But I will do it on the target branch `refactor-variables` this PR points to. There are some other major changes pending.

#723

examples/runner-certificates/README.md
```diff
@@ -34,7 +34,7 @@ Create a PEM-encoded `.crt` file containing the public certificate of your Gitla
 module {
   ...
   # Public cert of my companys gitlab instance
-  runners_gitlab_certificate = file("${path.module}/my_gitlab_instance_cert.crt")
+  agent_gitlab_certificate = file("${path.module}/my_gitlab_instance_cert.crt")
```
npalm, line 37: Should this not be runner_gitlab.... ?

npalm (changes requested):
Looks in general good. Will run a check also locally.

npalm:
Scopting with refactor will not create a new release, and also will not be mentioned int he release note as far I know. Maybe better to just scope it as feat.

kayman-mk:
The `!` should be the solution as it creates a major release. Let me quickly check this.

kayman-mk:
@npalm The only problem I see, is, how to manually update the release docs. Tried some changes here and there, but Release-Please overwrites them. Any idea how to change the docs manually?

I would like to have a reference to the migration script at a prominent place.

examples/runner-certificates/README.md
kayman-mk: Thanks for pointing at this. I took the chance to validate all `README.md` files and found some other problems too. Fixed.

kayman-mk:
@npalm Rebasing is a pain here. Can we merge to target branch `kayma/group-variables` to start testing the v7 release?

kayman-mk:
Merged into the pre-release branch for testing.

PR #815, opened 2023-04-26
fix: remove explicit aws_s3_bucket_acl

### LINKED ISSUE #814: AccessControlListNotSupported: The bucket does not allow ACLs
# Describe the bug

When attempting to apply the `gitlab-runner` module, the creation of the `build_cache_acl` resource fails.

```bash
╷
│ Error: error creating S3 bucket ACL for production220111411-570948727822-gitlab-runner-cache-dqgh6n63: AccessControlListNotSupported: The bucket does not allow ACLs
│ 	status code: 400, request id: YCWBQCJFN3ND3QGF, host id: DToeb/qB4wpmHrMBT/SK7zt8/doNeP4vpKnfbo0gHaN4M/QKrW7LZazQpj7iS5fHwzvEBsOBYb8=
│ 
│   with module.runner.module.cache[0].aws_s3_bucket_acl.build_cache_acl,
│   on .terraform/modules/runner/modules/cache/main.tf line 41, in resource "aws_s3_bucket_acl" "build_cache_acl":
│   41: resource "aws_s3_bucket_acl" "build_cache_acl" {
│ 
╵
```

# To Reproduce

1. plan an entirely new runner module instantiation keeping the default of the cache bucket enabled.
2. try to apply that plan

# Expected behavior

The cache bucket ACL creates without error.

# Additional context

This is due to changes AWS applied to bucket security defaults in April 2023. See this related issue in the Terraform AWS provider for mitigation steps, specifically under the "Standalone aws_s3_bucket_acl resource with default security settings" heading: https://github.com/hashicorp/terraform-provider-aws/issues/28353

It looks like the private ACL resource could be removed completely, or an additional `aws_s3_bucket_ownership_controls` will be required as per the notes in the related issue.

ryancausey:
This resolves an issue related to the April 2023 S3 API changes. More info can be found here: https://github.com/hashicorp/terraform-provider-aws/issues/28353

Closes #814

## Description

Remove the explicit private ACL that leads to an error during the apply phase as with the new security defaults, it should no longer be needed. I'm not 100% on the implications of removing this resource for existing deployments, so if that's a concern we can go the route of adding the explicit `aws_s3_bucket_ownership_controls` resource as per the related issue.

## Migrations required

NO

## Verification

- Solves the issue in #814 using the recommendations from: https://github.com/hashicorp/terraform-provider-aws/issues/28353
- manually tested and checked that the cache is still accessible

### DIFF
--- a/modules/cache/main.tf
+++ b/modules/cache/main.tf
@@ -38,12 +38,6 @@ resource "aws_s3_bucket" "build_cache" {
   force_destroy = true
 }
 
-resource "aws_s3_bucket_acl" "build_cache_acl" {
-  bucket = aws_s3_bucket.build_cache.id
-
-  acl = "private"
-}
-
 resource "aws_s3_bucket_versioning" "build_cache_versioning" {
   bucket = aws_s3_bucket.build_cache.id
 


ryancausey:
/help: lint PR title check failing with Error: Resource not accessible by integration

ryancausey:
/help

lint PR title check failing with Error: Resource not accessible by integration

ryancausey:
@kayman-mk I tried this using `var.runners_executor = "docker"` and the cache gets deployed but I'm getting 403 errors when the runner tries to push or pull from the cache. I'm allowing the module to create the cache and I am not specifying my own nor am I using the shared cache.

It looks to me like the only place the `local.bucket_policy` gets attached to anything is in `resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance"`. This appears to indicate that the bucket policy is never attached to the runner instance itself when not using `docker+machine`. Is this analysis correct?

I just wanted to confirm that I'm not missing something and that this isn't a side effect of removing the S3 bucket's private ACL resource.

ryancausey:
Looks to me like it is a separate issue so I opened #816 and #815.

kayman-mk (approved):
Thanks for your contribution.

PR #817, opened 2023-04-26
fix: allow s3 cache access for the "docker" runner executor

### LINKED ISSUE #816: 403 forbidden error when using cache with the "docker" runners executor
# Describe the bug

When using "docker" as the value for `var.runners_executor`, the runner is not able to access the S3 bucket created by default as a runner cache.

# To Reproduce

1. Plan a new instance of the runner using "docker" as the value for `var.runners_executor`.
2. Apply the resulting plan.
3. Trigger a GitLab pipeline that uses the newly deployed runner and attempts to pull from and push to the GitLab CI cache.

# Expected behavior

Pushing to and pulling from the cache should not result in a 403 forbidden error.

# Additional context

I believe the problem lies in that `resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance"` is what attaches the bucket policy that allows access to the cache bucket, but it only does so if `var.runners_executor` is set to "docker+machine". I locally modified the code to remove this check and the runner was then able to access the cache normally.

ryancausey:
Closes #816

## Description

This fixes the "docker" runner executor not having access to the S3 bucket by always attaching the bucket policy if it exists.

## Migrations required

NO

## Verification

Ran changes locally and observed the issue was resolved with this change.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -545,7 +545,7 @@ resource "aws_iam_role_policy_attachment" "docker_machine_cache_instance" {
   /* If the S3 cache adapter is configured to use an IAM instance profile, the
      adapter uses the profile attached to the GitLab Runner machine. So do not
      use aws_iam_role.docker_machine.name here! See https://docs.gitlab.com/runner/configuration/advanced-configuration.html */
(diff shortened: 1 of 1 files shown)


kayman-mk:
Yes, we shouldn't exclude the `docker` executor here.

kayman-mk (approved):
Thanks for the quick fix you provided.

PR #821, opened 2023-04-27
chore: make the `runners_user_data` a sensitive output

kayman-mk:
## Description

Some credentials are leaking through the output `runners_user_data`. We don't want to see these in our pipeline logs.

## Migrations required

No

## Verification

No verification done.


npalm (approved):

kayman-mk:
Looks pretty strange, as `nonsensitive` was introduced here, but the PR title says `sensitive`.

PR #822, opened 2023-04-28
feat: add option to read Gitlab Runner Registration token from SSM

### LINKED ISSUE #776: Security: registration token is stored in plan text in user-data

The runner token is stored in the SSM Parameter store but the registration token is just there in plain text.
It would make sense to get it dynamically from another parameter. That parameter can possibly be managed externally.

# Describe alternatives you've considered

# Suggest a solution

1. (option 1). Create an SSM parameter with provided registration token. Get the token dynamically from it.
2. (option 2). Point user-data to the ssm parameter to get the token from it. This way the SSM parameter can be managed externally and the token doesn't end up in the terraform state.

# Additional context

Not sure if it's worth creating another issue. The provided IAM policy allows reading *all* SSM parameters. It doesn't seem like it's required and access should be scoped down to one resource (or 2 if the registration token is placed in there as well).

Conoreby:
## Description

Adds the ability to read the Gitlab registration token from SSM. If no registration token is passed in, it will look in SSM to find the token to use. This prevents the token from being leaked as part of the user_data.

```hcl
module "gitlab_runner" {
  # ...
    registration_token = "" # this is the default value too
    # ...
  }

  secure_parameter_store_gitlab_runner_registration_token_name = "name-of-ssm-parameter-holding-the-registration-token"
```

Closes #776
Precondition for #186 to get rid of pre-registered runners.

## Migrations required

NO

## Verification

I modified the runner-default example to not pass in a registration token and added the token to SSM instead. Then I started up the runner and confirmed that it successfully registered with Gitlab.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -55,10 +55,12 @@ locals {
   # Ensure max builds is optional
   runners_max_builds_string = var.runners_max_builds == 0 ? "" : format("MaxBuilds = %d", var.runners_max_builds)
 
-  # Define key for runner and registration token for SSM
-  secure_parameter_store_registration_token_key = "${var.environment}-${var.secure_parameter_store_registration_token_key}"
-  secure_parameter_store_runner_token_key       = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
-  secure_parameter_store_runner_sentry_dsn      = "${var.environment}-${var.secure_parameter_store_runner_sentry_dsn}"
+  #
+  secure_parameter_store_gitlab_runner_registration_token_name = var.secure_parameter_store_gitlab_runner_registration_token_name
+
+  # Define key for runner token for SSM
+  secure_parameter_store_runner_token_key  = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
+  secure_parameter_store_runner_sentry_dsn = "${var.environment}-${var.secure_parameter_store_runner_sentry_dsn}"
 
   # Custom names for runner agent instance, security groups, and IAM objects
   name_runner_agent_instance = var.overrides["name_runner_agent_instance"] == "" ? local.tags["Name"] : var.overrides["name_runner_agent_instance"]
--- a/locals.tf
+++ b/locals.tf
@@ -55,9 +55,10 @@ locals {
   # Ensure max builds is optional
   runners_max_builds_string = var.runners_max_builds == 0 ? "" : format("MaxBuilds = %d", var.runners_max_builds)
 
-  # Define key for runner token for SSM
-  secure_parameter_store_runner_token_key  = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
-  secure_parameter_store_runner_sentry_dsn = "${var.environment}-${var.secure_parameter_store_runner_sentry_dsn}"
+  # Define key for runner and registration token for SSM
+  secure_parameter_store_registration_token_key = "${var.environment}-${var.secure_parameter_store_registration_token_key}"
+  secure_parameter_store_runner_token_key       = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
+  secure_parameter_store_runner_sentry_dsn      = "${var.environment}-${var.secure_parameter_store_runner_sentry_dsn}"
 
   # Custom names for runner agent instance, security groups, and IAM objects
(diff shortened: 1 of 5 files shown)


npalm:
@Conoreby thanks, can you have a look on the failing workflows. Can you also add a note to the top-level readme about the options users have to provide the registration token.

template/gitlab-runner.tftpl
```diff
@@ -32,10 +32,17 @@ then
   [[ "$valid_token_response" != "200" ]] && valid_token=false
 fi
 
+gitlab_runner_registration_token=${gitlab_runner_registration_token}
+# fetch registration token from SSM
+if [[ "$gitlab_runner_registration_token" == "__GITLAB_REGISTRATION_TOKEN_FROM_SSM__" ]]
+then
+    gitlab_runner_registration_token=$(aws ssm get-parameters --names "${secure_parameter_store_registration_token_key}" --with-decryption --region "${secure_parameter_store_region}" | jq -r ".Parameters | .[0] | .Value")
```
kayman-mk, line 39: question: any special reason to prefer `get-parameters` over `get-parameter` to read a single parameter?

template/gitlab-runner.tftpl
```diff
@@ -32,10 +32,17 @@ then
...
+fi
+
 if [[ "${runners_token}" == "__REPLACED_BY_USER_DATA__" && "$token" == "null" ]] || [[ "$valid_token" == "false" ]]
 then
   token=$(curl ${curl_cacert} --request POST -L "${runners_gitlab_url}/api/v4/runners" \
-    --form "token=${gitlab_runner_registration_token}" \
+    --form "token=$gitlab_runner_registration_token" \
```
kayman-mk, line 45: suggestion: revert the change here to use curly brackets for all parameters.

variables.tf
```diff
@@ -615,6 +615,12 @@ variable "gitlab_runner_registration_config" {
   }
 }
 
+variable "secure_parameter_store_registration_token_key" {
```
kayman-mk, line 618: suggestion: rename to something like `secure_parameter_store_gitlab_runner_registration_token_name`. `gitlab_runner` as it is the GitLab Runner registration token, `name` as this is used by `aws ssm` and in the AWS Console. What do you think?

variables.tf
```diff
@@ -615,6 +615,12 @@ variable "gitlab_runner_registration_config" {
...
+  description = "The key name used store the Gitlab runner registration token in Secure Parameter Store"
```
kayman-mk, line 619: issue: I guess this is a copy & paste issue as the module does not store anything in the parameter. It uses the parameter to retrieve the GitLab registration token.

suggestion: `The name of the SSM parameter to read the GitlLb Runner registration token from.`

locals.tf
```diff
@@ -55,9 +55,10 @@ locals {
   # Ensure max builds is optional
   runners_max_builds_string = var.runners_max_builds == 0 ? "" : format("MaxBuilds = %d", var.runners_max_builds)
 
-  # Define key for runner token for SSM
-  secure_parameter_store_runner_token_key  = "${var.environment}-${var.secure_parameter_store_runner_token_key}"
-  secure_parameter_store_runner_sentry_dsn = "${var.environment}-${var.secure_parameter_store_runner_sentry_dsn}"
+  # Define key for runner and registration token for SSM
+  secure_parameter_store_registration_token_key = "${var.environment}-${var.secure_parameter_store_registration_token_key}"
```
kayman-mk, line 59: issue: do not prepend the `environment` here. This is done for the other parameters as they are managed by this module and we have to create unique names. But your new parameter is not managed by this module, but outside (by the caller). Thus we should simply use the name provided by the caller.

kayman-mk (changes requested):
You made a good job here to make the module safer. Thank you.

locals.tf
Conoreby: Make sense to me, will change.

template/gitlab-runner.tftpl
Conoreby: This is an intentional change to not output the registration token in plaintext in user data.
With the string interpolation on the template using the curly bracket parameter will output the token in plain text in the user data. By using the environment variable instead, whether the token is fetched from SSM or not, it won't appear in plain text anymore.

kayman-mk (dismissed):

PR #860, opened 2023-06-01
fix: correct the bugs of major version 7 (pre-release)

kayman-mk:
## Description

Corrects all bugs of the pre-release version 7 of the module before roll-out.


kayman-mk:
First test with the Hapag-Lloyd test setup successfully done. Runner and Executors are up and running. Jobs were processed.

kayman-mk:
Second test with the live environment. Module conversion with script worked out of the box. Runner were up and running. All jobs were processed.

Spot instances, shared cache and fleet support.

kayman-mk:
Another test with a non-shared cache created by the module: Works

kayman-mk:
Like before, but the Agent is running on an on-demand instance. Works also fine.

kayman-mk:
Do not specify the `runner_worker` block: Works fine. I see default values in the `config.toml`

kayman-mk:
Switched from `docker+machine` to `docker`. Wow, how easy that is. Everything up and running and all jobs were processed on the Runner.

Issue #819, opened 2023-04-27
Breaking Change Ahead: Version 7


kayman-mk:
I will add some information on how to migrate to the new version in the next days. Still some time to go and not 100% discussed internally.

ToDo:
- [x] get an approval for #810 and then
- [x] test #723 

# Major Version 7

## Main reasons
  - harmonize the variable names
  - harmonize the documentation
  - remove deprecated variables
  - gain a better overview of the features provided
  - upgrade to Terraform 1.3+ for optional variables in maps

## Feature added
  - all possible Docker settings
  - `idle_scale_factor`

kayman-mk:
@npalm Can we go ahead with this one?

npalm:
NO objection

kayman-mk:
Great, I will release the changes on Sunday, September 10th at 1900z

kayman-mk:
Changes have been released on Sunday. Will closely monitor the issues here within this week.

kayman-mk:
❗ Just found out that the conversion script has a bug. Please check the instance type of the Runner Worker machines. I guess they are all `m5.large` (the module default).

See #975

ryancausey:
@kayman-mk here's what I get from `journalctl`. This is using runner version 16.2.2.

```
 journalctl -u gitlab-runner.service --no-pager
-- Logs begin at Wed 2023-10-25 21:06:51 UTC, end at Wed 2023-10-25 21:24:42 UTC. --
Oct 25 21:09:22 ip-10-0-2-65.us-west-2.compute.internal systemd[1]: Started GitLab Runner.
Oct 25 21:09:22 ip-10-0-2-65.us-west-2.compute.internal gitlab-runner[3374]: Runtime platform                                    arch=amd64 os=linux pid=3374 revision=30ac8e7f version=16.2.2
Oct 25 21:09:22 ip-10-0-2-65.us-west-2.compute.internal gitlab-runner[3374]: Starting multi-runner from /etc/gitlab-runner/config.toml...  builds=0 max_builds=0
Oct 25 21:09:22 ip-10-0-2-65.us-west-2.compute.internal gitlab-runner[3374]: Running in system-mode.
Oct 25 21:09:22 ip-10-0-2-65.us-west-2.compute.internal gitlab-runner[3374]:
[...]

Issue #513, opened 2022-06-20
wait_for_services_timeout parameter not supported


DnD-Magnum:
Hello,

First, thank you very much for this very good module ! 
 
We search to increase [wait_for_services_timeout](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runnersdocker-section) runner parameter but, it seems that this parameter is not implemented in the module.

Do you plan to develop it in the roadmap?

Best regards,
Geoffrey

kayman-mk:
See #511 Created the PR yesterday to restructure the variable sections and make it easy to add all other Docker options.

PR #861, opened 2023-06-02
feat: add support for `wait_for_services_timeout` option

tibuntu:
## Description

This PR adds the `wait_for_services_timeout` docker setting for the executor (see https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runnersdocker-section).

I've found #513 while looking for a solution to our problem. After migrating from the Kubernetes executer to the docker-machine executer, we have had a couple of Gitlab jobs that are always waiting 30s before actually running the defined steps.

I am aware of #819 but I believe this might be a quick win.

## Migrations required

 NO

## Test the change

In order to test my change, I recommend to set
```hcl
  debug = { "output_runner_config_to_file": true, "output_runner_user_data_to_file": false }
```

and then run `terraform plan`.

It will print the locally rendered `config.toml` that now contains the new setting:
```
[...]
  pre_clone_script = ""
  request_concurrency = 1
  output_limit = 4096
  limit = 0
  wait_for_services_timeout = 30
[...]
```

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -364,7 +364,7 @@ variable "runners_output_limit" {
 variable "runners_wait_for_services_timeout" {
   description = "How long to wait for Docker services. Set to -1 to disable. Default is 30."
   type        = number
-  default     = "30"
+  default     = 30
 }
 
 variable "userdata_pre_install" {
--- a/variables.tf
+++ b/variables.tf
@@ -361,6 +361,12 @@ variable "runners_output_limit" {
   default     = 4096
 }
 
+variable "runners_wait_for_services_timeout" {
+  description = "How long to wait for Docker services. Set to -1 to disable. Default is 30."
+  type        = number
+  default     = "30"
+}
+
 variable "userdata_pre_install" {
   description = "User-data script snippet to insert before GitLab runner install"
   type        = string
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -361,6 +361,12 @@ variable "runners_output_limit" {
   default     = 4096
 }
 
+variable "runners_wait_for_services_timeout" {
+  description = "How long to wait for Docker services. Set to -1 to disable. Default is 30."
+  type        = string
```
kayman-mk, line 366: question (blocking): Any reason to choose a `string`? Looks like a number.
```suggestion
  type        = number
```

kayman-mk (changes requested):
Thanks for your work! Looks good, but one question.

variables.tf
tibuntu: I wasn't 100% sure how Terraform would handle `-1` since this is a valid value for the runner config to disable the feature. But since I was curious, I've tested it myself and indeed number is the correct type here. I will change it ASAP.

kayman-mk (approved):

kayman-mk:
Thanks for your contribution, @tibuntu

PR #874, opened 2023-06-06
fix(runner-config): wait_for_services_timeout needs to be an integer …

dfrkp:
## Description

The current runners config leads to this error when installing the gitlab runner service:
```sh
FATAL: toml: line 33 (last key "runners.docker.wait_for_services_timeout"): incompatible types: TOML value has type string; destination has type integer 
```


kayman-mk (approved):
Thanks for the fix, @dfrkp

PR #876, opened 2023-06-09
feat: add new authentication method for GitLab >= 16

Kadeux:
## Description

GitLab released a new [authentication architecture for their runners](https://docs.gitlab.com/ee/architecture/blueprints/runner_tokens) since the version 16.0.0. This MR handle this new architecture while maintaining backward compatibility.

## Migrations required

Highly recommended as the old GitLab Runner registration method will be removed with GitLab 17.

Migration steps:
1. create a group access token for your GitLab group (needs the owner role and the api scope). To be refreshed manually once a year.
2. store the token in a SSM parameter and set the variable `runner_gitlab.access_token_secure_parameter_store_name` to this SSM parameter
3. remove `runner_gitlab_registration_config.registration_token`. No longer needed.
4. add `type = `project or group`, `group_id = <GitLab group number>` (for group runners) or `project_id = <GitLab project id>` (for project runners) to the `runner_gitlab_registration_config`  section.

## Verification

- running the old authentication method on GitLab 16: success
- running the new authentication method on GitLab 16: success


template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,45 @@ then
...
+      --form "access_level=${gitlab_runner_access_level}" \
+      | jq -r .token)
+  else
+    runner_type_param=""
+    if [ "${gitlab_runner_type}" = "group_type" ]; then
+      runner_type_param='--form group_id=${gitlab_runner_group_id}'
+    elif [ "${gitlab_runner_type}" = "project_type" ]; then
+      runner_type_param='--form project_id=${gitlab_runner_project_id}'
+    fi
+
```
kayman-mk, line 68: suggestion: It might happen, that the module caller forgot to set both `group_id` and `project_id`. Better to throw an error here to make the caller aware of this situation.

main.tf
```diff
@@ -69,13 +69,18 @@ locals {
...
       secure_parameter_store_region                                = var.aws_region
       gitlab_runner_registration_token                             = lookup(var.gitlab_runner_registration_config, "registration_token", "__GITLAB_REGISTRATION_TOKEN_FROM_SSM__")
+      gitlab_token                                                 = var.gitlab_token != "" ? var.gitlab_token : "__GITLAB_TOKEN_FROM_SSM__"
       gitlab_runner_description                                    = var.gitlab_runner_registration_config["description"]
       gitlab_runner_tag_list                                       = var.gitlab_runner_registration_config["tag_list"]
       gitlab_runner_locked_to_project                              = var.gitlab_runner_registration_config["locked_to_project"]
       gitlab_runner_run_untagged                                   = var.gitlab_runner_registration_config["run_untagged"]
       gitlab_runner_maximum_timeout                                = var.gitlab_runner_registration_config["maximum_timeout"]
+      gitlab_runner_type                                           = var.gitlab_runner_registration_config["type"]
```
kayman-mk, line 81: question: What about having a dedicated variable to indicate if the new authentication method shall be used? This makes it explicit which authentication method to use.

```
# in locals.tf
locals {
  # maybe a good idea to use this variable in `gitlab-runner.tfpl` too
  use_new_runner_authentication_gitlab_16 = contains(keys(gitlab_runner_registration_config), "type")
}
```

kayman-mk (changes requested):
Thanks for solving this issue, @Kadeux 

I have some questions and suggestions which came into my mind. Especially because everyone needs to adjust his config at the moment.

Could you please also update the default example and describe how to configure the new authentication method?

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,45 @@ then
...
-  token=$(curl ${curl_cacert} --request POST -L "${runners_gitlab_url}/api/v4/runners" \
-    --form "token=$gitlab_runner_registration_token" \
-    --form "tag_list=${gitlab_runner_tag_list}" \
-    --form "description=${gitlab_runner_description}" \
-    --form "locked=${gitlab_runner_locked_to_project}" \
-    --form "run_untagged=${gitlab_runner_run_untagged}" \
-    --form "maximum_timeout=${gitlab_runner_maximum_timeout}" \
-    --form "access_level=${gitlab_runner_access_level}" \
-    | jq -r .token)
+  if [[ "${gitlab_runner_version}" < "16.0.0" ]]
```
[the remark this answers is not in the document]
Kadeux: I think using `use_new_runner_authentication_gitlab_16` from your last comment is better

kayman-mk:
Linter still complains about àn unused variable `use_new_runner_authentication_gitlab_16`. Is it still work in progress?

Kadeux:
> Linter still complains about an unused variable `use_new_runner_authentication_gitlab_16`. Is it still work in progress?

Sorry I didn't see it, I still don't know why did I put this variable here :)

Tiduster:
@kayman-mk : Any additional remarks on this change?

Do not hesitate to ask for more changes if required.

Best regards,

kayman-mk:
Environment for testing:
GitLab 16.0.8
GitLab Runner 16.0.2

Upgraded my runner to this version and verify that jobs are still processed without changing my configuration: success

kayman-mk:
Tried to use the new authentication. But it didn't work.

```hcl
  runner_gitlab_access_token_secure_parameter_store_name = aws_ssm_parameter.gitlab_runner_registration_token_new.name

    # registration_token = var.gitlab_runner_registration_token.value
    tag_list           = "test"
    description        = "runner for testing new version (${each.value.availability_zone})"
    locked_to_project  = "false"
    run_untagged       = "false"
    maximum_timeout    = "10800"

    type = "group"
    group_id = 123
  }
```

It still uses the old registration method. The new `glrt-` registration token wasn't used.

kayman-mk:
Regarding the registration process of the runners: The module needs a PAT to access GitLab and register the runner.

I think in an enterprise environment we shouldn't give a PAT to our module. I suggest that users register the runner manually and we use the `glrt-` token then. What do you think @Kadeux @npalm @Tiduster

kayman-mk:
Seems that my question is obsolete as there are access tokens on group level available. https://about.gitlab.com/blog/2023/07/06/how-to-automate-creation-of-runners/

variables.tf
```diff
@@ -336,6 +339,12 @@ variable "runner_gitlab_registration_token_secure_parameter_store_name" {
   default     = "gitlab-runner-registration-token"
 }
 
+variable "runner_gitlab_access_token_secure_parameter_store_name" {
```
kayman-mk, line 342: question: as we usually use group access tokens: How to handle the token expiration? The agent is no longer able to register. How to inform the user? Is there an option to refresh the token automatically?

variables.tf
```diff
@@ -302,6 +302,9 @@ variable "runner_gitlab_registration_config" {
     registration_token = optional(string, "")
     tag_list           = optional(string, "")
     description        = optional(string, "")
+    type               = optional(string, "") # mandatory if gitlab_runner_version >= 16.0.0
```
kayman-mk, line 305: thought: would be nice to have a `condition` here to validate the values.

kayman-mk (changes requested):

README.md
```diff
@@ -328,6 +328,40 @@ module "runner" {
...
+  subnet_id_runners        = element(module.vpc.private_subnets, 0)
+
+  runners_name       = "docker-default"
+  runners_gitlab_url = "https://gitlab.com"
+
+  runner_gitlab_access_token_secure_parameter_store_name = "gitlab_token_ssm"
+
+
+  runner_gitlab_registration_config = {
+    type               = "instance_type"
```
kayman-mk, line 354: issue: `instance_type` is not a valid value, is it? `project_type`?

Either `group_id` or `project_id` is missing too.

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,48 @@ then
...
-  token=$(curl ${curl_cacert} --request POST -L "${runners_gitlab_url}/api/v4/runners" \
-    --form "token=$gitlab_runner_registration_token" \
-    --form "tag_list=${gitlab_runner_tag_list}" \
-    --form "description=${gitlab_runner_description}" \
-    --form "locked=${gitlab_runner_locked_to_project}" \
-    --form "run_untagged=${gitlab_runner_run_untagged}" \
-    --form "maximum_timeout=${gitlab_runner_maximum_timeout}" \
-    --form "access_level=${gitlab_runner_access_level}" \
-    | jq -r .token)
+  if [ "${use_new_runner_authentication_gitlab_16}" == "false" ]
```
kayman-mk, line 47: suggestion: flip the if/else as the `else` part contains the standard mechanism to be used in future.

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,48 @@ then
...
+      | jq -r .token)
+  else
+    runner_type_param=""
+    if [ "${gitlab_runner_type}" = "group_type" ]; then
+      if [ ! -z "${gitlab_runner_group_id}" ]; then
+        echo "If the runner type is group_type, you must specify a group_id".
+      fi
+      runner_type_param='--form group_id=${gitlab_runner_group_id}'
+    elif [ "${gitlab_runner_type}" = "project_type" ]; then
+      if [ ! -z "${gitlab_runner_project_id}" ]; then
```
kayman-mk, line 66: issue: `! -z` means not empty. `-z` is better here to show the error in case the string is missing. Same as above.

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,48 @@ then
...
+        echo "If the runner type is project_type, you must specify a project_id".
```
kayman-mk, line 67: suggestion: raise an exit state with `exit`. The GitLab Runner is not able to work. Same as above.

kayman-mk (changes requested):
Nice feature you implemented! Appreciated to be able to use the module with GitLab 17+

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,48 @@ then
...
+      --form "tag_list=${gitlab_runner_tag_list}" \
+      --form "description=${gitlab_runner_description}" \
+      --form "locked=${gitlab_runner_locked_to_project}" \
+      --form "run_untagged=${gitlab_runner_run_untagged}" \
+      --form "maximum_timeout=${gitlab_runner_maximum_timeout}" \
+      --form "access_level=${gitlab_runner_access_level}" \
+      | jq -r .token)
+  else
+    runner_type_param=""
+    if [ "${gitlab_runner_type}" = "group_type" ]; then
```
kayman-mk, line 60: suggestion: use `group` and `project` and remove the `_type` suffix. Already part of the variable name (also in variables.tf)

variables.tf
kayman-mk, line 342: suggestion: the variables were grouped with the v7 release. The prefix `runner_gitlab` already indicates that the variable should be moved to the `runner_gitlab` block and the prefix removed.

template/gitlab-runner.tftpl
```diff
@@ -39,17 +39,48 @@ then
...
+      if [ ! -z "${gitlab_runner_group_id}" ]; then
+        echo "If the runner type is group_type, you must specify a group_id".
```
kayman-mk, line 62: suggestion: use a prefix like `ERROR: ` to make clear that we have a problem here. Same as below.

variables.tf
Kadeux: I didn't find any clue about auto refreshing tokens, but I think that GitLab sends notification when a token is about to expire.

README.md
Kadeux: `instance_type` is a valid value and doesn't require neither `group_id` or `project_id` since it is at the instance level.

kayman-mk: Oh yes, you are right. It's clearly stated in the docs. https://docs.gitlab.com/ee/api/runners.html

kayman-mk:
Looks good. Needs a final test on my side.

kayman-mk:
Successfully tested the new authentication schema. Had some problems as I added the new variables into the wrong blocks and got no error. Guess this is one of the downsides using maps as variables.

kayman-mk:
Tested both scenarios with success. @Kadeux Could you please have a look at my (minor) changes, please?

kayman-mk (approved):
Great! Let's go with the new authentication method. Thanks for your work.

PR #905, opened 2023-07-12
fix: add missing defaults

npalm:
While checking the PR, I was not able to switch from main to the refactory-varialbes branch for the example-default.

- fix a missing default
- add a null check in stated of a string check


kayman-mk:
Good spot, thanks!

PR #962, opened 2023-09-18
feat: add `MaxGrowthRate` to limit the number of instances added in parallel

### LINKED ISSUE #851: Support MaxGrowthRate

The module already supports parameters like idleCount.

Maybe it should also support MaxGrowthRate: https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runnersmachine-section

rohaquinlop:
## Description

This feature allows to support `MaxGrowthRate` which limits the number of instances added in parallel by the Runner. The default of `0` (unlimited) is used if no value is given.

The parameter is added to `runner_worker_docker_machine_instance` as this is the only Runner type for which it makes sense. All other types are not able to scale and thus it is not a `runner_worker` parameter.

See [advanced-configuration](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runnersmachine-section)

Closes #851 

## Migrations required

NO

## Verification

- do not set the parameter --> `MaxGrowthRate` in config file is 0.
- set the parameter to any value --> `MaxGrowthRate` in config file is "any value"

### DIFF
--- a/examples/runner-default/versions.tf
+++ b/examples/runner-default/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "5.14.0"
+      version = "5.15.0"
     }
     local = {
       source  = "hashicorp/local"
--- a/main.tf
+++ b/main.tf
@@ -107,6 +107,7 @@ locals {
       runners_spot_price_bid            = var.runner_worker_docker_machine_instance_spot.max_price == "on-demand-price" || var.runner_worker_docker_machine_instance_spot.max_price == null ? "" : var.runner_worker_docker_machine_instance_spot.max_price
       runners_ami                       = var.runner_worker.type == "docker+machine" ? data.aws_ami.docker-machine[0].id : ""
       runners_security_group_name       = var.runner_worker.type == "docker+machine" ? aws_security_group.docker_machine[0].name : ""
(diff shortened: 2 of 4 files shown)


dsalaza4:
Hi @npalm @kayman-mk,

A peer review here would be greatly appreaciated! 🙏🏽

kayman-mk:
Seems that the formatting is broken. Could you please run a `terraform fmt -recursive` to fix that?

kayman-mk:
Looks good at first hand. But I would prefer not to mention this parameter in the examples as it is quite specific and an advanced configuration.

rohaquinlop:
@kayman-mk already added your suggestions, thanks in advance for your help!

kayman-mk:
Did you test the change without specifying the `MaxGrowthRate`?

rohaquinlop:
@kayman-mk No, I was trying to test it but I don't understand completely the documentation. I'm still trying on it.

kayman-mk:
If you are talking about the `tests/` directory: Don't waste your time. This is still work in progress. I test my PRs manually by deploying them.

kayman-mk:
Ok, just checked, that the parameter is set as expected. Works fine.

kayman-mk (approved):
Good job! Thanks for adding this feature.

dsalaza4:
@kayman-mk Thank you so much for all the help!

PR #998, opened 2023-10-01
chore: do all dependency updates once a week on the same day

kayman-mk:
## Description

Terraform updates were scheduled for Sunday, everything else for once a week which could be another weekday. Now all dependency updates are done once a week.


kayman-mk:
We have to see if this is working. Usually I assume it is better to have one PR per update to easily track which update breaks the code.

PR #1019, opened 2023-10-27
fix: delete the 'runner_config_toml_rendereded' output variable

### LINKED ISSUE #1018: "Error: Output refers to sensitive values" when applying fleet configuration
# Describe the bug

When applying a spot-fleet configuration, the apply fails with this error:

```
Error: Output refers to sensitive values

  on outputs.tf line 56:
  56: output "runner_user_data" {

To reduce the risk of accidentally exporting sensitive data that was
intended to be only internal, Terraform requires that any root module
output containing sensitive data be explicitly marked as sensitive, to
confirm your intent.

If you do intend to export this data, annotate the output value as
sensitive by adding the following argument:
    sensitive = true
```

This is because that output is exporting `local.template_user_data` which is templated using `local.template_gitlab_runner` which is templated using `tls_private_key.fleet[0].private_key_pem` which is flagged as sensitive according to the documentation. https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key#private_key_pem

# To Reproduce

This is reproducible using any fleet configuration. Relevant parts of my configuration:

```hcl
  environment = "test-runners"

  vpc_id    = var.vpc_id
  subnet_id = var.public_subnets[2]

    registration_token = var.runners_token
    description        = "Test Runners"
    locked_to_project  = "true"
    run_untagged       = "true"
    maximum_timeout    = "3600"
  }

  runner_instance = {
    name       = "Test Runners"
    ssm_access = true
  }

  runner_gitlab = {
    url = "https://gitlab.com"
  }

    private_address_only = false
    idle_count           = 0
    root_size            = 16
    subnet_ids           = var.public_subnets
    types = [
      "m6i.large", "m6a.large", "m5zn.large", "m5a.large", "m5.large",
    ]
  }

  runner_worker_docker_machine_fleet = {
    enable = true
  }
```

# Expected behavior

Apply proceeds without errors.

JulianCBC:
## Description

Due to calls to `nonsensitive` in case the data is sensitive or not calling the function in the other case, the code gets very complex. The output variable `runner_config_toml_rendereded` was introduced for easier debugging, but we still have the option to write the whole Runner configuration to a local file (`debug.write_runner_config_to_file`).

This "fixers" that issue by just deleting the output variable.

Closes #1018

## Migrations required

Usually not. For debugging purposes, set the `debug.write_runner_config_to_file` to `true`. The configuration is written to a file on your local disk (directory `debug/`)

## Verification

Only linted as this is only deleting an output variable.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -57,8 +57,3 @@ output "runner_user_data" {
   description = "(Deprecated) The user data of the Gitlab Runner Agent's launch template. Set `var.debug.output_runner_user_data_to_file` to true to write `user_data.sh`."
   value       = nonsensitive(local.template_user_data)
 }
-
-output "runner_config_toml_rendered" {
-  description = "The rendered config.toml given to the Runner Manager."
-  value       = local.template_runner_config
-}


kayman-mk:
The fix is definitely needed. But the code looks rather complicated: `var.runner_worker_docker_machine_fleet.enable == true ? nonsensitive(local.template_runner_config) : local.template_runner_config`. It should simply return the template. But we have to wrap it with `nonsensitive` sometimes. I understand, why this is needed.

Maybe we find another solution to export the Runner's configuration? This was the original intend of this output variable. But I guess they will all have this sensitivity problem.

JulianCBC:
> The fix is definitely needed. But the code looks rather complicated: `var.runner_worker_docker_machine_fleet.enable == true ? nonsensitive(local.template_runner_config) : local.template_runner_config`. It should simply return the template. But we have to wrap it with `nonsensitive` sometimes. I understand, why this is needed.

I completely agree: the goal of this pull request was to start the discussion and to that end I chose the "smallest" fix that wasn't completely insane or removing functionality. We definitely need a better solution than this.

Some options from my perspective:

1. Just delete this output - it's not something people should need day-to-day unless they're either paranoid or debugging the module
2. Only output something in this output if a flag is set - essentially turning it into a debug aid only
3. Move the `nonsensitive()` call up into the inputs to the template, making that always non-sensitive - this would mean that the private key is effectively leaked into the state without any _attempt_ to flag it as being private
4. Pass the private key to the EC2 instance another way

kayman-mk:
Voting for 2 and delete the output. The module writes the configuration to a local file. This should be sufficient for debugging.

Would be nice to get rid of sensitive/non sensitive. And I wouldn't consider this as a major change, as it is for debugging only.

JulianCBC:
Thanks @kayman-mk, I've re-done this pull request to just delete the output variable.

kayman-mk:
PR updated as we use it as commit message.

kayman-mk (approved):
Thanks for fixing this bug.

JulianCBC:
Thanks @kayman-mk for your help in sorting this out and fixing up the PR message.

PR #1021, opened 2023-10-27
refactor: extract data sources to its own dedicated file

lmilbaum:
## Description

Reducing main.tf file size will improve its readability. The data sources are moved to a separate file.

## Migrations required

Nn

## Verification

No verification done. Just moving the data sources.


kayman-mk:
Refactoring the ` main.tf`  is a good idea. I usually prefer to have files grouped by functionality. So I think it would be a good idea to split it into a
- ` runner.tf`
- ` runner-worker.tf`
- ...
and put the `data`  statements into the respective files. I guess some of them are so general that they will end up in a non specific ` data.tf`

What do you think?

lmilbaum:
> Refactoring the ` main.tf` is a good idea. I usually prefer to have files grouped by functionality. So I think it would be a good idea to split it into a
> 
> * ` runner.tf`
> * ` runner-worker.tf`
> * ...
> 
> and put the `data` statements into the respective files. I guess some of them are so general that they will end up in a non specific ` data.tf`
> 
> What do you think?

I like the suggested approach. Will push a change soon.

kayman-mk:
Sorry, wasn't aware that there are only some data sources. Thought we have many more. So I combined everything into one file again. One file per data source is too much overhead.

kayman-mk:
Thanks for investing your time to improve the module, @lmilbaum.

kayman-mk (approved):

PR #1032, opened 2023-11-08
fix: remove the deprecated runner_user_data output variable

### LINKED ISSUE #1031: Using `nonsensitive` in `runner_user_data` variable causes `terraform plan` to fail
# Describe the bug

As of version 7.2.0 of this module, which includes #876 to allow using the new runner registration method, the `runner_user_data` output variable's `value` is passed through `nonsensitive` first. When switching to use the new auth method, this results in the following error:

```bash
╷
│ Error: Invalid function argument
│ 
│   on .terraform/modules/runner/outputs.tf line 58, in output "runner_user_data":
│   58:   value       = nonsensitive(local.template_user_data)
│     ├────────────────
│     │ while calling nonsensitive(value)
│     │ local.template_user_data is "trimmed for brevity"
│ 
│ Invalid value for "value" parameter: the given value is not sensitive, so
│ this call is redundant.
╵
```

It appears that there are configurations where the `template_user_data` will not be sensitive, so `nonsensitive` cannot be used here.

# To Reproduce

1. Instantiate the runner module using the new authorization method, and do not pass any values to the module that are sensitive.
2. Run a `terraform plan`
3. See error

# Expected behavior

The `terraform plan` should not fail due to the `nonsensitive` redundant call error.

# Additional context

Runner module instantiation example that caused this:

```hcl
# ssm.tf
resource "aws_ssm_parameter" "gitlab_runner_access_token" {
  name        = "gitlab_runner_access_token"
  description = "The access token for the GitlabRunner to use to register."
  type        = "SecureString"
  value       = var.gitlab_token
}

# iam.tf
data "aws_iam_policy_document" "gitlab_token_access" {
  statement {
    sid       = "GitLabTokenAccess"
    actions   = ["ssm:GetParameter"]
    resources = [aws_ssm_parameter.gitlab_runner_access_token.arn]
  }
}

resource "aws_iam_policy" "gitlab_token_access" {
  name_prefix = "gitlab_token_access"
  description = "This policy allows access to the GitLab token stored in SSM."
  policy      = data.aws_iam_policy_document.gitlab_token_access.json
}

# main.tf
module "runner" {
  source  = "npalm/gitlab-runner/aws"
  version = "7.2.0"

  # We need to override the environment name to be less than 64 characters in length.
  #
  # The module code uses the environment name + some postfix string by default, so we
  # use that limited to a max length of 21 to allow for postfix string lengths.
  environment = substr(var.environment_name, 0, 21)

  vpc_id    = data.aws_vpc.main.id
  subnet_id = data.aws_subnets.public_subnet_ids.ids[0]

    description       = "Ephemeral runner for test project."
    locked_to_project = "true"
    run_untagged      = "false"
    maximum_timeout   = "3600"
    type              = "project"
    project_id        = "123456"
  }

  runner_instance = {
    # Buff our runner instance size since we aren't using the docker+machine. This means
    # the jobs run directly on a runner, so a t3.micro instance might not cut it.
    type = "m5.large"
    # To allow the runner to access the internet without requiring a NAT gateway, we must
    # give them a public IP.
    use_eip              = true
    private_address_only = false
    name                 = "test-${var.environment_name}"
    # Allow SSM access to help debug if runner issues arise.
    ssm_access = true
    tag_list   = join(", ", var.runner_tags)
  }

  runner_gitlab = {
    url                                      = "https://gitlab.com"
    runner_version                           = "16.2.2"
    access_token_secure_parameter_store_name = aws_ssm_parameter.gitlab_runner_access_token.name
  }

  runner_cloudwatch = {
    enable = false
  }

  # This has the jobs run on the same EC2 instance as the agent, no autoscaling is used.
  runner_worker = {
    type = "docker"
  }

  runner_worker_cache = {
    random_suffix = true
  }

  runner_networking = {
    security_group_ids = [data.aws_security_group.rds.id]
  }

  # This ends up taking precedence over the `subnet_id` input above, but that input is
  # required.
  # https://github.com/cattle-ops/terraform-aws-gitlab-runner/blob/5100efd3445c3f06e5089d970da5a3a0341624eb/main.tf#L177C42-L177C79
    subnet_ids = data.aws_subnets.public_subnet_ids.ids
  }

  # We need to attach additional policies to allow the runner to get the new API token.
  runner_role = {
    policy_arns = [aws_iam_policy.gitlab_token_access.arn]
  }
}
```

ryancausey:
## Description

Depending on the value of the output a call to `nonsensitive` is needed. Instead of introducing a conditional call, we remove the output variable. The value is still written to a local file, see `debug.write_runner_user_data_to_file`.

Closes #1031.

## Migrations required

YES - any use of the `runner-user-data` output must be changed to use `var.debug.write_runner_user_data_to_file`. As this is used for debug only, it shouldn't be a problem for your live environment.

## Verification

None. Just removing the output variable.


ryancausey:
I just noticed #1018 and #1019, so I've changed this to always mark this variable as sensitive.

kayman-mk:
As this variable is already deprecated and used for debugging only, we should simply delete it (like in #1019). Users can set `debug.write_runner_user_data_to_file` to `true` to write the data to the local disk.

outputs.tf
```diff
@@ -55,7 +55,8 @@ output "runner_launch_template_name" {
 
 output "runner_user_data" {
   description = "(Deprecated) The user data of the Gitlab Runner Agent's launch template. Set `var.debug.output_runner_user_data_to_file` to true to write `user_data.sh`."
-  value       = nonsensitive(local.template_user_data)
+  value       = local.template_user_data
+  sensitive   = true
 }
```
kayman-mk, line 60: According to #1019 and changing the `nonsensitive` back and forth, I suggest to remove the whole output. It's deprecated and the value is also written to a local file in the `debug/` directory.

kayman-mk (changes requested):

kayman-mk (approved):

kayman-mk:
Thanks for improving the module, @ryancausey

PR #1033, opened 2023-11-08
fix: retry the jq installation in case of errors

ryancausey:
## Description

In case of network connection problems, the `yum` operations ends with an error. This PR adds a retry to install `jq`.

## Migrations required

NO

## Verification

Confirmed this works for my setup by vendoring this module with these changes.


kayman-mk:
Could you please share the error message you have seen in the logs?

ryancausey:
> Could you please share the error message you have seen in the logs?

```bash
Nov  8 21:43:26 user-data:
Nov  8 21:43:26 user-data: Error downloading packages:
Nov  8 21:43:26 user-data:  jq-1.5-1.amzn2.0.2.x86_64: [Errno 256] No more mirrors to try.
Nov  8 21:43:26 user-data:
```

The added retry logic eventually retries this package install enough times to make it work. When it was a single try, this would fail causing the userdata script would fail and thus causing the runner to fail to register and start.

kayman-mk (changes requested):

kayman-mk:
Installing `jq` always shouldn't do any harm, but removes the `if`. Thought about integrating this into the loop above. But this seems to be a bit more tricky.

kayman-mk (approved):

kayman-mk:
Thanks for fixing this issue, @ryancausey

PR #1034, opened 2023-11-09
fix: run the GitLab Runner deregistration process at shutdown

### LINKED ISSUE #1014: remove_gitlab_registration script doesnt run on termination of instance
it seems like the intention is for the /etc/init.d/remove_gitlab_registration script to run on instance termination to runlevel 0. This doesn't seem to be happening. As far as i can tell the script works in itself, and when triggered manually with a stop command it works ok - it just doesn't actually run when a machine is terminated in aws. So when we use terraform to destroy them it's not being run. 

This used to work for us on older versions of the module but stopped working at some point. Needing us to manually remove the runners from the gitlab UI now. I don't know if its due to some chance or timing related difference on AWS's side, though.

kayman-mk:
## Description

We have a script which removes the GitLab Runner from GitLab in case it is shutdown. This script wasn't run automatically. This PR adds a Systemd service which starts the removal script at shutdown automatically.

Closes #1014 

## Migrations required

No

## Verification

Verified manually that the Runner is removed from GitLab at shutdown.

### DIFF
--- a/template/gitlab-runner.tftpl
+++ b/template/gitlab-runner.tftpl
@@ -155,77 +155,40 @@ then
   docker-machine rm -y dummy-machine
   unset HOME
   unset USER
-
 fi
 
-# A small script to remove this runner from being registered with Gitlab.
-cat <<REM > /etc/rc.d/init.d/remove_gitlab_registration
+# A small script to remove this runner from being registered with Gitlab. Executed at shutdown.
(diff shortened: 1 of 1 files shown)


kayman-mk:
I didn't figure out why the script wasn't running using `chkconfig`. My AWS contact advised me to use a Systemd service as mentioned in #1014.

PR #1063, opened 2024-01-06
feat: support setting thorughput and iops for ebs volumes

tibuntu:
## Description

These changes will add support for setting `throughput` and `iops` to the EBS device that is being attached to the docker_machine instance. Because setting `iops` is now possible, this will also allow the usage of `io1` and `io2` EBS volumes types.

## Migrations required

No - If you consider my defaults for the two new variables as sane.

## Verification

I have adapted the module which is already by changing the `runner_worker_docker_machine_instance` input variables:

```
    private_address_only = false
    types                = ["m5.large", "t3.xlarge", "m6a.large"]
    subnet_ids           = [module.vpc.public_subnets[0]]
    runners_idle_time    = 180
    root_size            = 30
    volume_type          = "io2"
    volume_iops = 6000
    volume_throughput = 250
  }
```

The volume was created successfully when testing the runner via Gitlab:

Screenshot 2024-01-06 at 12 14 09


main.tf
```diff
@@ -330,6 +330,8 @@ resource "aws_launch_template" "fleet_gitlab_runner" {
     ebs {
       volume_size = var.runner_worker_docker_machine_instance.root_size
       volume_type = var.runner_worker_docker_machine_instance.volume_type
+      iops        = var.runner_worker_docker_machine_instance.volume_type == "gp3" || var.runner_worker_docker_machine_instance.volume_type == "io2" || var.runner_worker_docker_machine_instance.volume_type == "io1" ? var.runner_worker_docker_machine_instance.volume_iops : null
```
kayman-mk, line 333: nitpick: to make the expression a little bit shorter, I suggest to use

```suggestion
      iops        = contains(["gp3", "io1", "io2"], var.runner_worker_docker_machine_instance.volume_type ? var.runner_worker_docker_machine_instance.volume_iops : null
```

variables.tf
```diff
@@ -683,6 +685,8 @@ variable "runner_worker_docker_machine_instance" {
     subnet_ids                 = optional(list(string), [])
     types                      = optional(list(string), ["m5.large"])
     volume_type                = optional(string, "gp2")
+    volume_throughput          = optional(number, 125)
+    volume_iops                = optional(number, 3000)
```
kayman-mk, line 689: question: These are the default values today? I am not able to check this in my environment now.

variables.tf
```diff
@@ -666,7 +666,9 @@ variable "runner_worker_docker_machine_instance" {
     start_script = Cloud-init user data that will be passed to the Runner Worker. Should not be base64 encrypted.
     subnet_ids = The list of subnet IDs to use for the Runner Worker when the fleet mode is enabled.
     types = The type of instance to use for the Runner Worker. In case of fleet mode, multiple instance types are supported.
-    volume_type = The type of volume to use for the Runner Worker.
+    volume_type = The type of volume to use for the Runner Worker. `gp2`, `gp3`, `io1` or `io2` are supported.
```
kayman-mk, line 669: suggestion: This drops `sc1` and `st1` support. I guess nobody uses the hard drives and the cloud default is `gp2`. So I am fine with that. But we should add a validation.

```hcl
validation {
  condition = contains(["gp2", "gp3", "io1", "io2"], var.runner_worker_docker_machine_instance.volume_type)
  message = "Supported volume types: gp2, gp3, io1 and io2"
}
```

variables.tf
tibuntu: Simply looking at our current volumes: Yes

AFAIK these are the gp3 defaults by AWS

Screenshot 2024-01-15 at 11 12 23

variables.tf
tibuntu: Great feedback, thank you. I have added the validation block.

main.tf
tibuntu: Makes total sense, changed.

kayman-mk (approved):

kayman-mk:
Thanks for your contribution, @tibuntu

PR #1080, opened 2024-02-06
feat: error checking for initial API token call

MrGiga:
## Description

If the API token is invalid or expired, the current implementation does not return that error to inform the end user. This implementation adds error checking, returns that error and then does not continue. This should make debugging much easier compared to it's current state. From my research there are two possible error formats.

First error response: 401 unauthorized
{"message":"401 Unauthorized"}

Second possible error response: Token has expired
{"error":"invalid_token","error_description":"Token is expired. You can either do re-authorization or token refresh."}

## Migrations required

None

## Verification

Provide a bad or expired API token to the module. It should fail and output the reason for failure.


MrGiga:
Code is slightly incomplete due to the following:

1. I am looking for input on how the error should be outputted 
2. What should be outputted to cover scenario where token is null and there is no message or error_description.

kayman-mk:
> Code is slightly incomplete due to the following:
> 
> 1. Is there specific semantics that the error output needs to follow?
No. It's fine as it is.

> 2. What should be outputted to cover scenario where token is null and there is no message or error_description? Currently it just continues.
I think, we ignore this for now, as the GitLab API specification says, that this is the format of error messages. So we can expect an error message. By the way: Good job to include the `error_description` as well!

https://docs.gitlab.com/ee/api/rest/#sudo
https://docs.gitlab.com/ee/api/rest/#data-validation-and-error-reporting

template/gitlab-runner.tftpl
```diff
@@ -58,8 +58,16 @@ then
...
       $runner_type_param \
-      --form "access_level=${gitlab_runner_access_level}" \
-      | jq -r '.token')
+      --form "access_level=${gitlab_runner_access_level}")
+
+    token = $(echo response | jq -r '.token')
+    if [[ "$token" == null ]]
+        message = $(echo response | jq -r '.message // .error_description')
+        if [[ "$message" != null ]]
+            echo "ERROR: API call returned $message".
```
kayman-mk, line 67: suggestion: Give some more information to the user which API call failed to make debugging easier.

```suggestion
            echo "ERROR: Couldn't register the Runner. GitLab API call returned $message".
```

kayman-mk (changes requested):
Thanks for contributing. This makes the debugging of the registration process easier.

MrGiga:
> > Code is slightly incomplete due to the following:
> > 
> > 1. Is there specific semantics that the error output needs to follow?
> >    No. It's fine as it is.
> 
> > 2. What should be outputted to cover scenario where token is null and there is no message or error_description? Currently it just continues.
> >    I think, we ignore this for now, as the GitLab API specification says, that this is the format of error messages. So we can expect an error message. By the way: Good job to include the `error_description` as well!
> 
> https://docs.gitlab.com/ee/api/rest/#sudo https://docs.gitlab.com/ee/api/rest/#data-validation-and-error-reporting

My thoughts are that if there is no token and it can't get a token, then the script should stop there. This will prevent the logs from being convoluted with further error messages from the runner it's self.

template/gitlab-runner.tftpl
MrGiga: I have updated the code as suggested.

kayman-mk (approved):

PR #1084, opened 2024-02-08
fix: rename `pull_policies` to `pull_policy` in `config.toml`

### LINKED ISSUE #1065: issue with pull_policies
# Describe the bug

The `pull_policies` argument is not working as intented. From gitlab's own documentation only `pull_policy` and `allowed_pull_policies` are allowed: 

https://docs.gitlab.com/runner/executors/docker.html#:~:text=Use%20the%20following%20values%20for,local%20version%20does%20not%20exist.

# To Reproduce

Instead of:
```  
runner_worker_docker_options = {
    privileged            = "true"
    pull_policies = ["if-not-present"]
  }
```

Try:
```
  runner_worker_docker_options = {
    privileged            = "true"
    allowed_pull_policies = ["always", "if-not-present"]
  }
```

Running a runner with `if-not-present` crash makes it error out, as only `always` is allowed.

# Expected behavior

If not present works accordingly.

# Additional context

I think pull_policies should become allowed_pull_policies in the outputted config.toml, but I'm not sure as it is kinda like a breaking change to pipelines...

kayman-mk:
## Description

The `runner_worker_docker_options` mentioned a `pull_policies` attribute which was directly used in `config.toml` which is wrong. The correct name is `pull_policy`. `pull_policy` is ignored by the Runner and has no effect. Instead a pull policy of `always` is used automatically.

This PR renames the `pull_policies` to `pull_policy` before inserting it into the configuration file. If `pull_policies` is used at the moment it has no effect.

Closes #1065 

## Migrations required

In case `var.runner_worker_docker_options.pull_policy` is used, rename `pull_policy` to `pull_policies` and make sure that it is a `list(string)`, e.g. `pull_policies = ["always"]`

## Verification

Still to be done

- [X] `pull_policies` not given --> `pull_policy = ["always"] is used.
- [X] `pull_policies` given --> Specified value used


kayman-mk:
I think the above code makes no sense at all. Better to rename `pull_policies` before inserting it into the configuration file. This way we know that it accepts a list of strings as `pull_policy` implies one value only (at least for me).

PR #1095, opened 2024-02-29
feat: enable docker IPv6 support on executors

### LINKED ISSUE #1094: Support Docker option for IPv6

I'm trying to use the runner in a IPv6 only environment. The runner itself works well and is able to connect to my Gitlab instance, but when the job starts within its own docker container it gets a connection timeout to Gitlab as its not able to make use of the IPv6 networking. 

To allow this to work, `enable_ipv6` needs to be set in the Docker configuration otherwise docker networks don't support it. 

See https://docs.docker.com/config/daemon/ipv6/ and https://gitlab.com/gitlab-org/gitlab-runner/-/issues/36994

# Suggest a solution

Add the enable_ipv6 flag to [runner_worker_docker_options](https://github.com/cattle-ops/terraform-aws-gitlab-runner?tab=readme-ov-file#input_runner_worker_docker_options). Note the `enable_ipv6` documentation is currently missing, see the issue above.

kayman-mk:
## Description

Adds a configuration flag `runner_worker_docker_options.enable_ipv6` for the GitLab Runner to enable IPv6 support on the executors which is needed to run in IPv6 networks only. The configuration is set in `config.toml` section `[runners.docker]`.

This setting is still undocumented. See [GitLab issue](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/36994).

Closes #1094 

## Verification

- [x] there is no `enable_ipv6` option in the config, if the value is not set
- [x] `enable_ipv6` is set in the config, if a value is given
- [x] if the option is set, the Runner starts and does not report errors
- [x] if the option is set, the Executors start and do not report errors


kayman-mk:
@william00179 Please cross-check

rk295:
Does this alone enable v6? Reading the [docs](https://docs.docker.com/config/daemon/ipv6/) suggests we also need to configure the docker daemon as well.

I have been using this as a way to enable v6 in the engine:

```
  runner_worker_docker_machine_ec2_options = [
    "engine-opt=ipv6=true",
    "engine-opt=ip6tables=true",
    "engine-opt=experimental=true",
  ]
```

Looking at the Gitlab runner [source](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/main/common/config.go#L260) seems to suggest this only takes affect if the `FF_NETWORK_PER_BUILD` flag is on.

kayman-mk:
Yeah, true. We should have something like `ip_version = ["6"]` to enable ipv6 only which should configure everything in the background. For the time being, I will add it to the documentation.

william00179:
Hi @kayman-mk, 

Thanks for jumping onto this so quickly. 

I have been trying to get this working and for my case which is the docker runner, not the docker+machine executor. 

I have tried to enable IPv6 in docker but the various configuration methods are all  leading to errors when the job tries to run. I have asked for some [additional information](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/36994#note_1800534069) on the required configuration to work with the gitlab runner. 

For this to work, the feature flag needs to be set in the runners config.toml, see https://docs.gitlab.com/runner/executors/docker.html#create-a-network-for-each-job

There are some additional considerations to be made here also. Do we use a ULA subnet which would keep the IPv4 style of the containers not becoming globally routable for those who don't know better, or do we allow GUA which will then require some changes to the ENI to allocate an IPv6 prefix which can then be used. 

I can't think of a use-case for when a user would want the job to be globally routable, but perhaps it exists. 

When I can get a working IPv6 configuration I will let you know the required configuration for the module.

Issue #1062, opened 2024-01-05
'remove-gitlab-registration' script can invalidate in-use runner authentication token


schmidt-galen-heb:
# Describe the bug

We're in the process of upgrading to `7.2.2` + the [new runner registration workflow](https://docs.gitlab.com/ee/ci/runners/new_creation_workflow.html), and discovered an unfortunate race condition with `remove-gitlab-registration`.

Sometimes when cycling the agent instance, the AWS ASG will spin up the new instance before tearing down the old one. In that case, the following sequence of events can occur:

| Old Instance | New Instance |
| ------------- | ------------- |
| _starts_ | _n/a_ |
| [obtains `Runner Token 1` from Gitlab](https://github.com/cattle-ops/terraform-aws-gitlab-runner/blob/8b73d42fe21602a343155743cb5d4beea0733b89/template/gitlab-runner.tftpl#L32-L81) | _n/a_ |
| [adds `Runner Token 1` to SSM](https://github.com/cattle-ops/terraform-aws-gitlab-runner/blob/8b73d42fe21602a343155743cb5d4beea0733b89/template/gitlab-runner.tftpl#L82) | _n/a_ |
| _running_ | _n/a_ |
| _running_  | _starts_ |
| _running_ | [fetches and validates `Runner Token 1` from SSM](https://github.com/cattle-ops/terraform-aws-gitlab-runner/blob/8b73d42fe21602a343155743cb5d4beea0733b89/template/gitlab-runner.tftpl#L21-L28) |
| _running_ | _finishes startup_ |
| _begins shutdown_ | _running_ |
| [remove_gitlab_registration.sh](https://github.com/cattle-ops/terraform-aws-gitlab-runner/blob/8b73d42fe21602a343155743cb5d4beea0733b89/template/gitlab-runner.tftpl#L174-L180) runs and invalidates `Runner Token 1` | _running_ |
| _finishes shutdown_ | _receives 403 from Gitlab because `Runner Token 1` is now invalid_ |

# To Reproduce

The exact steps to reproduce seem to be tricky because the ASG seems inconsistent in how it's cycling the instances, but I believe it's:
1. Upgrade module to `7.2.2`
[...]

PR #1099, opened 2024-03-14
feat: add graceful terminate option to terminate-agent-hook

long-wan-ep:
## Description

Resolves https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1029.

Adds optional functionality to the `terminate-agent-hook` module for graceful termination, which users can enabled/configure using new input variables. When graceful terminate is enabled, the lambda will give running jobs a chance to finish before the runner instances are terminated. When graceful terminate is disabled, the original behavior of the `terminate-agent-hook` is used.

Updated `.pylintrc` to use `/usr/local/lib/python3.12/site-packages/`, because pylint was not able to find the boto3 dependencies in `/usr/local/lib/python3.11/site-packages/`. Seems like megalinter is using python 3.12 now.

## Migrations required

No

### With graceful terminate enabled
1. Manually deployed a runner with:
```
runner_worker_graceful_terminate = {
    enabled      = true
    retry_period = 60
    timeout      = 300
}
```
2. Started a GitLab job using the runner
3. Manually triggered instance refresh in the runner ASG
4. Verified the new graceful termination steps in the lambda runs successfully
5. Verified the existing steps in the lambda runs successfully

### With graceful terminate disabled
1. Manually deployed a runner
2. Started a GitLab job using the runner
3. Manually triggered instance refresh in the runner ASG
4. Verified the lambda runs successfully without graceful termination steps


long-wan-ep:
Currently if graceful terminate is enabled, the runner is highly likely to run into https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1062, since graceful terminate delays runner termination during a runner refresh. I think we should tackle that issue first before we implement this feature.

kayman-mk:
#1062 will be solved via #1102 soon.

variables.tf
```diff
@@ -750,6 +750,17 @@ variable "runner_worker_docker_machine_autoscaling_options" {
...
+variable "runner_worker_graceful_terminate" {
+  description = "Gracefully terminate Runner Worker, by waiting a set amount of time for running jobs to finish before termination."
+  type = object({
+    enabled      = optional(bool, false)
+    timeout      = optional(number, 1800)
+    retry_period = optional(number, 300)
+    job_timeout  = optional(number, 3600)
+  })
+  default = {}
+}
```
kayman-mk, line 762: suggestion: remove this variable and enable the graceful termination always. Is there any reason why someone should not use this?

kayman-mk (commented):
Thanks for your work, @long-wan-ep 

I am still trying to understand the termination process. It looks like that we are not waiting until all jobs on the workers/executors are finished but we are killing them at time of GitLab Runner shutdown.

variables.tf
```diff
@@ -750,6 +750,17 @@ variable "runner_worker_docker_machine_autoscaling_options" {
   default = []
 }
 
+variable "runner_worker_graceful_terminate" {
+  description = "Gracefully terminate Runner Worker, by waiting a set amount of time for running jobs to finish before termination."
+  type = object({
+    enabled      = optional(bool, false)
+    timeout      = optional(number, 1800)
```
kayman-mk, line 757: docs: this should be the maximum duration of the longest job running on the workers/executors. 1800s seems good for me, but a documentation should be added. Setting this value too low will result in not gracefully shutdown the Runner.

modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -10,12 +10,105 @@
...
+        print(json.dumps({
+            "Level": "error",
+            "Message": "Failure waiting for command to be successful"
+        }))
+        raise error
+
+    if command_response['Status'] == "Success":
+        print(json.dumps({
+            "Level": "info",
+            "Message": f"gitlab-runner service stopped, SSM command response: {command_response}"
```
kayman-mk, line 106: nitpick: This seems to be incorrect as we do not know if the service has been stopped already. It might still be running, waiting for jobs to be finished, right?

modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -232,26 +325,97 @@ def handler(event, context):
...
+        
+        ec2_client = boto3.client("ec2", region_name=region)
+    
     # make sure that no new instances are created
-    cancel_active_spot_requests(ec2_client=client, executor_name_part=os.environ['NAME_EXECUTOR_INSTANCE'])
+    cancel_active_spot_requests(ec2_client=ec2_client, executor_name_part=os.environ['NAME_EXECUTOR_INSTANCE'])
 
     # find the executors connected to this agent and terminate them as well
-    _terminate_list = ec2_list(client=client, parent=event_detail['EC2InstanceId'])
+    _terminate_list = ec2_list(client=ec2_client, parent=instance_id)
```
kayman-mk, line 410: issue: we shouldn't terminate the instances in case of a graceful shutdown. This is done by the GitLab Runner itself. Otherwise the shutdown is no longer graceful.

modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -270,7 +434,7 @@ def handler(event, context):
             "Message": "No instances to terminate."
         }))
 
-    remove_unused_ssh_key_pairs(client=client, executor_name_part=os.environ['NAME_EXECUTOR_INSTANCE'])
+    remove_unused_ssh_key_pairs(client=ec2_client, executor_name_part=os.environ['NAME_EXECUTOR_INSTANCE'])
```
kayman-mk, line 437: note: as far as I believe, we can remove the SSH keys immediately. They are copied during startup into local files on the GitLab Runner instance and thus no longer needed.

modules/terminate-agent-hook/iam.tf
```diff
@@ -134,6 +203,65 @@ data "aws_iam_policy_document" "spot_request_housekeeping" {
...
+
+  # Permit the function to send SSM commands to the GitLab Runner instance
+  statement {
+    sid = "GitLabRunnerGracefulTerminateSSMSendEC2"
+    actions = [
+      "ssm:SendCommand"
+    ]
+    effect = "Allow"
+    resources = [
+      "arn:${data.aws_partition.current.partition}:ec2:${data.aws_region.this.name}:${data.aws_caller_identity.this.account_id}:instance/*"
```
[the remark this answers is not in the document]
long-wan-ep: The SSM command is sent to the gitlab runner instance not the workers, which doesn't have `gitlab-runner-parent-id`.

modules/terminate-agent-hook/lambda/lambda_function.py
```diff
@@ -232,26 +325,97 @@ def handler(event, context):
     :param event: see https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-concepts.html#gettingstarted-concepts-event
     :param context: see https://docs.aws.amazon.com/lambda/latest/dg/python-context.html
     """
-    event_detail = event['detail']
 
-    if event_detail['LifecycleTransition'] != "autoscaling:EC2_INSTANCE_TERMINATING":
-        sys.exit()
+    # if graceful terminate is enabled then the event will be in SQS message format
```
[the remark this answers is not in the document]
long-wan-ep: Yes, updated comment to provide better explanation.

variables.tf
long-wan-ep: I made it an optional feature because I thought there might be users who don't need/want graceful termination and want to keep the existing behavior, but I agree that it should be an improvement for most users. We could set `enabled` default to `true` if we want to give users the choice, or if you don't think that's necessary then we can remove the `enabled` field and make it mandatory.

variables.tf
long-wan-ep: Updated docs for `timeout` in README.

modules/terminate-agent-hook/lambda/lambda_function.py
long-wan-ep: The SSM command sent to the runner instance will stop the gitlab-runner service and only return success if the service is inactive.

modules/terminate-agent-hook/lambda/lambda_function.py
long-wan-ep: This section would only run after:
1. graceful terminate is successful
OR
2. we reach the graceful terminate timeout

In the first case, the SSM command should cleanup the workers using `docker-machine rm`, though if for some reason it fails then the workers could be left running.

In the second case, I believe it's possible for the runner instance to be terminated by the lifecycle hook before gitlab runner can remove the workers, leaving the workers orphaned.

I think having this section would act as a back up for these situations.

modules/terminate-agent-hook/lambda/lambda_function.py
long-wan-ep: I have this line after the graceful terminate section to prevent it from running multiple time, since the graceful terminate section of the code is allowed to fail and retry.

modules/terminate-agent-hook/lambda/lambda_function.py
kayman-mk: So the waiter above returns only in case the service has been stopped? This could last up to 1800s (by default), but as far as I understand the waiter, it definitely returns after 3*10 seconds, doesn't it? Is this a problem? Wouldn't we return a failure then?

modules/terminate-agent-hook/lambda/lambda_function.py
kayman-mk: Yes, sounds reasonably to me. Let's keep it here.

modules/terminate-agent-hook/lambda/lambda_function.py
long-wan-ep: The SSM command sent to the instance does not wait for the service to finish stopping(1800s), it will:
1. attempt to stop `gitlab-runner.service`(non blocking)
2. wait 5s
3. check the status of `gitlab-runner.service`
4. exit 0(SSM command status = "Success") if the service is stopped and docker machines removed, exit 1(SSM command status = "Failed") otherwise

This `if` statement checks the status of the SSM command, not the waiter. Also, if the waiter reaches max retries and the SSM command hasn't finished running, then it should fail and the function will exit with failure to be retried via SQS.

Sorry it's a bit confusing, let me know if that doesn't make sense.

long-wan-ep:
Since https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1062 has been resolved, this is no longer blocked.

kayman-mk:
@long-wan-ep There is another PR ongoing (#1117 ). On a first view it seems to be that their implementation is more robust. Could you have a look please and comment on that?

long-wan-ep:
@kayman-mk The implementation from #1117 is much simpler and works, let's close this MR in favour of the other solution.

PR #1105, opened 2024-04-03
feat: add field `create_aws_s3_bucket_public_access_block` to variable `runner_worker_cache`

KevinSnyderCodes:
## Description

Some organizations may disallow configuring block public access settings on individual S3 buckets. For example, the organization may use account level configuration to block public access on all buckets.

To support this, we add the field `create_aws_s3_bucket_public_access_block` to the `runner_worker_cache` variable.

We add `count` to the `aws_s3_bucket_public_access_block` resource to control its creation and use the `moved` keyword to ensure that existing instances of this resource are not recreated due to this change.

## Migrations required

No. Migrations are automatically handled by `moved` keyword.

## Verification

Applied this module with the `create_aws_s3_bucket_public_access_block` field set to `false`. The `aws_s3_bucket_public_access_block` resource was not created. All other resources were created as expected.


variables.tf
```diff
@@ -421,18 +421,19 @@ variable "runner_worker_cache" {
...
+    bucket                                   = optional(string, "")
+    bucket_prefix                            = optional(string, "")
+    create                                   = optional(bool, true)
+    expiration_days                          = optional(number, 1)
+    include_account_id                       = optional(bool, true)
+    policy                                   = optional(string, "")
+    random_suffix                            = optional(bool, false)
+    shared                                   = optional(bool, false)
+    versioning                               = optional(bool, false)
+    create_aws_s3_bucket_public_access_block = optional(bool, true)
```
kayman-mk, line 436: docs: could you please add a description to the docs above? E. g. that it is useful in case public access blocks are denied on organizational level.

modules/cache/moved.tf
```diff
@@ -0,0 +1,4 @@
+moved {
```
kayman-mk, line 1: suggestion: I think a file name like `state_migration.tf` makes it clearer, what happens in this file. So I suggest to rename the file.

variables.tf
```diff
@@ -765,3 +766,8 @@ variable "debug" {
   })
   default = {}
 }
+
+variable "create_aws_s3_bucket_public_access_block" {
```
kayman-mk, line 770: issue: please remove this unused variable

variables.tf
kayman-mk, line 436: nitpick: please sort the variables by name

kayman-mk (changes requested):

kayman-mk:
Did some quick tests in my environment. Looks good to me.

kayman-mk (approved):
Thanks for improving this module.

modules/cache/variables.tf
```diff
@@ -91,3 +91,8 @@ variable "kms_key_id" {
   type        = string
   default     = ""
 }
+
+variable "create_aws_s3_bucket_public_access_block" {
```
kayman-mk, line 95: issue: please add a description otherwise the `kics` job fails

kayman-mk (changes requested):

kayman-mk (approved):

Issue #1109, opened 2024-04-11
Support for glrt- tokens


Unibozu:
Hi

This is following https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/876 which added support for auto-registering the runner with the [Next GitLab Runner Token Architecture](https://docs.gitlab.com/ee/architecture/blueprints/runner_tokens/index.html#comparison-of-current-and-new-runner-registration-flow).

Personal or Group Access Tokens are impossible to securely scope with Gitlab, they grant read/write access to every repositories settings under the group or user. Not ideal.

# Describe the solution you'd like

Gitlab now recommends to manually register the runners first (https://docs.gitlab.com/ee/architecture/blueprints/runner_tokens/index.html#proposal) which provides a GitLab runner token starting with `glrp-`. Passing this token as `token` in the `config.toml` allows the runner to properly register.

It would make sense for the module to support both methods: auto-registration with PAT for ease of use, or directly passing a GLRT for the runner to use for people who want extra security. This is ignoring registration tokens that are meant to disappear once Gitlab 17 is released in May.

It is possible to use the GLRT right now by manually setting the value of the `-runner-token` SSM parameter created by the module and restarting the instance. But this doesn't play well with terraform.

A simple implementation would be to seed the token to `aws_ssm_parameter.runner_registration_token` instead of `null`.

Thoughts?

I'm happy to implement, I wanted a sounding board first - we need the solution soon before Gitlab 17 releases in May.

Best

woz5999:
Update: The new workflow is no longer recommended, it's required. Per an email they sent out today:

"The legacy workflow that uses runner registration tokens is deprecated and will be disabled in GitLab.com with the deployment of GitLab 17, after 2024-05-22."

woz5999:
@kayman-mk Any thoughts here? This seems like a pretty huge problem on the horizon.

kayman-mk:
At the moment, it's nice that everything is managed by the module. But if GitLab deprecated the workflow and we have to register the Runner beforehand, we should do this. And to be honest, this would make the module simpler again, supports grouping of Runners in the UI. But you have to find a way to renew the tokens, I guess.

My proposal: We need an option to tell the module to simply use a given token without modifing it. Then we should already be fine. In case the token becomes invalid, the Runner writes a log entry.

It seems to be feasible on short notice as the coding is already in place (80% or so).

What do you think?

woz5999:
[...]

PR #1115, opened 2024-04-20
feat: add option to use a pre-registered Runner

### LINKED ISSUE #1074: Module update removes registered runner but does not create a new runner
# Describe the bug

Updating GitLab Runner version by re-deploying a Terraform project using this module with the new "Next GitLab Runner Token Architecture", as the new EC2 instance is created, GitLab Runner is removed from the GitLab instance, but a new one is not registered.

Before "Next GitLab Runner Token Architecture", the GitLab Runner was simply updated.

# To Reproduce

1. Update `runner_gitlab = { runner_version = "16.7.0" }` to a newer version (eg. `16.8.0`)
2. `terraform plan` + `terraform apply`
3. Observe new master runner instance being created on EC2 and old one destroyed
4. Observe runner being deleted from GitLab instance
5. Wait and observe no new runner being created on the GitLab instance
6. Have no runner

# Expected behavior

A GitLab Runner to either be recreated or re-use the existing runner registration.

# Additional context

Upgrade to "Next GitLab Runner Token Architecture" was performed with a completely fresh setup, no old SSM parameters remained. We deploy using and GitLab's Terraform state handling through GitLab's CI (running on a instance completely separated from any auto-deployed ones from this module).

We have to fully destroy the runner and recreate it to get the runner re-registered. EC2 system logs show nothing different from the past, but I can supply a log if no other ideas are available.

kayman-mk:
## Description

GitLab announced then [Next GitLab Runner Token Architecture](https://docs.gitlab.com/ee/architecture/blueprints/runner_tokens/index.html#proposal). Runners have to be registered manually.

This PR adds a new import parameter `runner_gitlab.preregistered_runner_token_ssm_parameter_name` holding the name of a SSM parameter (type: `SecuredString`). This parameter contains the GitLab Runner token obtained from GitLab. All other registration methods will still work, but have been marked as deprecated and will be removed with [v8.0.0](https://github.com/cattle-ops/terraform-aws-gitlab-runner/milestone/4) end of the year.

This also solves the problems with Runners removed from GitLab at shutdown, resulting in new Runners not able to start.

Closes #1074 and #1109 

## Verification

- [x] deployed the module using the new registration version. Runner is online.
- [x] deployed the module using the old authentication schema. Runner is online.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -5,12 +5,12 @@ Common pitfalls are documented in [pitfalls.md](pitfalls.md).
 ## Configuration
 
 The examples are configured with defaults that should work in general. The examples are in general configured for the
-region Ireland `eu-west-1`. The only parameter that needs to be provided is the GitLab are the registration token and the
-URL of your GitLab instance. The token can be found in GitLab in the runner section (global, group or repo scope).
-Create a file `terraform.tfvars` and the registration token.
+region Ireland `eu-west-1`. The only parameter that needs to be provided is the name of a SSM parameter holding the Runner
+registration token and the URL of your GitLab instance. The Runner token is created in GitLab during the Runner registration process.
+Create a file `terraform.tfvars` and put the Runner registration token in the SSM parameter.
 
 ```hcl
-registration_token = "MY_TOKEN"
+preregistered_runner_token_ssm_parameter_name = "my-gitlab-runner-token-ssm-parameter-name"
 gitlab_url   = "https://my.gitlab.instance/"
 ```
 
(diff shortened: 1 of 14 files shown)


kayman-mk:
We already have several options to configure the Runner registration. I used a new parameter to be able to remove all the old methods easily with v8.0.0

kayman-mk:
@woz5999 @lpsm-nuageit Can you please have a quick look here? Are there any better names for the new variable coming into your mind?

kayman-mk:
Still to do: update examples to new registration method, update docs and describe new registration method

lpsm-nuageit:
@kayman-mk I believe we can use the same config nomenclatures that gitlab runner uses in the helm chart version. 

But since we're in a different stream, I think the name `preregistered_runner_token_ssm_parameter_name` makes more sense and avoids ambiguity.

woz5999:
This all looks reasonable to me

czomo:
Are there any plans to support AWS Secret Manager as secret store for Gitlab tokens?

bck01215:
The docs are still unclear to me (new to project) `preregistered_runner_token_ssm_parameter_name` gets created so it can't exist beforehand. Does that mean I need to pass the token into the resource after it's created?

PR #1117, opened 2024-04-23
feat: implement graceful shutdown of GitLab Runner

### LINKED ISSUE #1029: Option to gracefully terminate runner

When the `terminate-agent-hook` runs, workers are terminated and running jobs are interrupted. We would like an option to gracefully terminate runners, so that the running jobs are given a chance to complete

# Describe alternatives you've considered

We previous disabled the creation of `terminate-agent-hook` and used our own hook + lambda to handle graceful termination, but `terminate-agent-hook` was made mandatory so we can no longer do this.

# Suggest a solution

We suggest adding an option to gracefully terminate runners in the `terminate-agent-hook` lambda. We can contribute our graceful termination logic to `terminate-agent-hook` if it works for you. Here is a brief summary of our solution:

1. Configure the gitlab-runner service to gracefully stop:
ie. 
```
cat <<EOF > /etc/systemd/system/gitlab-runner.service.d/kill.conf
[Service]
# Time to wait before stopping the service in seconds
TimeoutStopSec=600
KillSignal=SIGQUIT
EOF
```
2. Send a message to a SQS queue when a runner's terminate lifecycle hook triggers
3. Lambda triggers from SQS message
4. Lambda sends command to runner EC2 instance to stop the gitlab-runner service, done via AWS SSM document
5. Lambda waits for the SSM document to finish executing
  a. If the gitlab-runner service successfully stopped, lambda completes lifecycle hook
  b. If the gitlab-runner service has not successfully stopped, error is thrown and the SQS message goes back to the queue to be retried in the next run
7. Lambda terminates any workers still running

tmeijn:
## Description

Based on the discussion https://github.com/cattle-ops/terraform-aws-gitlab-runner/discussions/1067:

1. Move the EventBridge rule that triggers the Lambda from `TERMINATING` to `TERMINATE`. The Lambda now functions as an "after-the-fact" cleanup instead of being responsible of cleanup _during_ termination.
2. Introduces a shell script managed by Systemd, that monitors the target lifecycle of the instance and initiates GitLab Runner graceful shutdown.
3. Makes the heartbeat timeout of the ASG terminating hook configurable, with a default of the maximum job timeout + 5 minutes, capped at `7200` (2 hours).
4. Introduces a launching lifecyclehook, allowing the new instance to provision itself and GitLab Runner to provision its set capacity before terminating the current instance.

### Todos

- [ ] Provide a diagram
- [ ] Investigate https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/1117#issuecomment-2110177977

## Migrations required

No, except that if the default behavior of immediately terminating all Workers + Manager, the `runner_worker_graceful_terminate_timeout_duration` variable should be set to 30 (the minimum allowed).

### Graceful terminate

1. Deploy this version of the module.
2. Start a long running GitLab job.
3. Manually trigger an instance refresh in the runner ASG.
4. Verify the job keeps running and has output. Verify from the instance logs that GitLab Runner service is still running.
6. Once remaining jobs have been completed, observe that GitLab Runner service is terminated and instance is put into `Terminating:Proceed` status

### Zero Downtime deployment

1. Deploy this version of the module.
2. Start multiple, long running GitLab jobs, twice the capacity of the GitLab Runner.
3. Manually trigger an instance refresh in the runner ASG.
4. Verify the jobs keep running and have output. Verify from the instance logs that GitLab Runner service is still running.
5. Verify new instance gets spun up, while the current instance stays `InService`.
7. Verify new instance is able to provision its set capacity.
8. Verify new instance starts picking up GitLab jobs from the queue before current instance gets terminated.
9. Observe that there is zero downtime.
10. Once remaining jobs have been completed, observe that GitLab Runner service is terminated and current instance is put into `Terminating:Proceed` status

Closes #1029


main.tf
```diff
@@ -634,15 +634,16 @@ resource "aws_iam_role_policy_attachment" "eip" {
...
+  name                                   = var.runner_terminate_ec2_lifecycle_hook_name == null ? "terminate-instances" : var.runner_terminate_ec2_lifecycle_hook_name
+  environment                            = var.environment
+  asg_arn                                = aws_autoscaling_group.gitlab_runner_instance.arn
+  asg_name                               = aws_autoscaling_group.gitlab_runner_instance.name
+  cloudwatch_logging_retention_in_days   = var.runner_cloudwatch.retention_days
+  name_iam_objects                       = local.name_iam_objects
+  name_docker_machine_runners            = local.runner_tags_merged["Name"]
+  role_permissions_boundary              = var.iam_permissions_boundary == "" ? null : "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:policy/${var.iam_permissions_boundary}"
+  kms_key_id                             = local.kms_key
+  asg_hook_terminating_heartbeat_timeout = var.runner_worker_graceful_terminate_timeout_duration == null ? min(7200, tonumber(coalesce(var.runner_gitlab_registration_config.maximum_timeout, 0)) + 300) : var.runner_worker_graceful_terminate_timeout_duration
```
tmeijn, line 646: **todo:** move this to a local variable

modules/terminate-agent-hook/variables.tf
```diff
@@ -60,3 +60,14 @@ variable "enable_xray_tracing" {
...
 }
+
+variable "asg_hook_terminating_heartbeat_timeout" {
+  description = "Duration the ASG should stay in the Terminating:Wait state."
+  type        = number
+  default     = 30
+
+  validation {
+    condition     = var.asg_hook_terminating_heartbeat_timeout >= 30 && var.asg_hook_terminating_heartbeat_timeout <= 7200
+    error_message = "AWS support heartbeat timeout in the range 30 - 7200."
```
tmeijn, line 71:
```suggestion
    error_message = "AWS only supports heartbeat timeout in the range of 30 to 7200."
```

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
 
+#############################
+# GRACEFUL SHUTDOWN SERVICE #
+#############################
+
+# set timeout for runner service to wait before stopping
+mkdir /etc/systemd/system/gitlab-runner.service.d
+cat <<EOF > /etc/systemd/system/gitlab-runner.service.d/kill.conf
+[Service]
+# according to the GitLab Runner docs, this initiates a graceful shutdown of the service
```
tmeijn, line 257: **todo:** link to docs

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+TimeoutStopSec=0
+EOF
+
+cat <<'EOF' > /opt/monitor_runner.sh
+#!/bin/bash
+
+# Fetch instance ID since we need it for the lifecycle hook actions.
+imds_token=$(curl -sSf -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 300")
+instance_id=$(curl -sS -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/instance-id")
+region=$(curl -s -H "X-aws-ec2-metadata-token: $imds_token" http://169.254.169.254/latest/meta-data/placement/region)
```
tmeijn, line 268:
```suggestion
region=$(curl -s -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/placement/region")
```

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+
+# Function to send complete-lifecycle-action
+send_complete_lifecycle_action() {
+    # Add your logic to send complete-lifecycle-action here
```
tmeijn, line 272: **todo:** remove

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+
+    # Retrieve the lifecycle hook names associated with the Auto Scaling Group
+    lifecycle_hooks=$(aws --region $region autoscaling describe-lifecycle-hooks --auto-scaling-group-name $asg_name --query "LifecycleHooks[*].LifecycleHookName" --output text)
+    if [ -z "$lifecycle_hooks" ]; then
+    echo "No lifecycle hooks found for Auto Scaling Group: $asg_name"
+    exit 1
+    fi
+    echo "Lifecycle Hooks: $lifecycle_hooks"
+
+    echo "Will send complete actions in 10 seconds..."
```
tmeijn, line 297:
```suggestion
    # This is needed to let CloudWatch Logs catch up.
    echo "Will send complete actions in 10 seconds..."
```

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+# Function to monitor the existence of the Runner process
+monitor_process() {
+    while true; do
+        systemctl --no-block stop gitlab-runner.service
+        sleep 5
+        status=$(systemctl is-active gitlab-runner.service)
+        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
+          # TODO: maybe also Docker Machine Cleanup here?
+          # Currently the Lambda is resposible for this.
+          echo "GitLab Runner service $status..."
```
tmeijn, line 327: **todo:** elaborate next action.

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+          send_complete_lifecycle_action
+        else
+          echo "GitLab Runner Serivce still running, sleeping..."
+        fi
+        sleep 5 # Adjust the interval as needed
```
tmeijn, line 332:
```suggestion
        sleep 5
```

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+main() {
+    # Poll whether the instance is in the terminating state using instance metadata API
+    imds_token=$(curl -sSf -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 300")
+    target_lifecycle_state=$(curl -sSf -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/autoscaling/target-lifecycle-state")
+
+    if [ "$target_lifecycle_state" = "Terminated" ]; then
+        # Monitor the existence of the Runner process
+        monitor_process
+    else
+        echo "Instance target lifecycle state is $target_lifecycle_state. Not doing anything."
```
tmeijn, line 346:
```suggestion
        echo "Instance target lifecycle state is $target_lifecycle_state. No action required."
```

variables.tf
```diff
@@ -752,6 +752,12 @@ variable "runner_worker_docker_machine_autoscaling_options" {
   default = []
 }
 
+variable "runner_worker_graceful_terminate_timeout_duration" {
+  description = "Amount of time to wait for GitLab Runner to finish picked up jobs. Defaults to the `maximum_timeout` configured + `5m`. Maximum allowed is `7200` (2 hours)"
+  type        = number
+  default     = null
+}
+
```
tmeijn, line 760: **todo:** move this variable to the `runner_worker` "namespace".

tmeijn (commented):
Some todo's during self-review.

tmeijn:
Thank you @long-wan-ep, your MR definitely helped in making this MR better so thank you for proposing your solution 🙏🏾 . I'd like to invite you to review this MR and come with any questions or suggestions you might have!

tmeijn (commented):
Some open questions.

@kayman-mk, is there a reason we do not have `terraform_docs` in pre-commit, nor as a check in CI?

.pre-commit-config.yaml
```diff
@@ -1,11 +1,10 @@
 repos:
   - repo: https://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.64.1
+    rev: v1.89.0
     hooks:
       - id: terraform_fmt
         args:
           - --args=-recursive
-      - id: terraform_tflint
```
tmeijn, line 8: **note:** removed this as it is broken and not made equivalent to the CI check.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,134 @@ then
...
+instance_id=$(curl -sS -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/instance-id")
+region=$(curl -s -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/placement/region")
+
+# Function to send complete-lifecycle-action
+send_complete_lifecycle_action() {
+    echo "Sending complete-lifecycle-action for instance $instance_id..."
+
+    if [ -z "$region" ]; then
+    echo "Failed to retrieve AWS region."
+    exit 1
```
long-wan-ep, line 290: I think these should be indented.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,134 @@ then
...
+    fi
+    echo "Current AWS Region: $region"
+
+    # Retrieve the Auto Scaling Group name associated with the instance
+    asg_name=$(aws --region $region ec2 describe-tags --filters "Name=resource-id,Values=$instance_id" "Name=key,Values=aws:autoscaling:groupName" --query 'Tags[].Value' --output text)
+    if [ -z "$asg_name" ]; then
+    echo "Failed to retrieve Auto Scaling Group name."
+    exit 1
```
long-wan-ep, line 298: I think these should be indented.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,134 @@ then
...
+    fi
+    echo "Auto Scaling Group Name: $asg_name"
+
+    # Retrieve the lifecycle hook names associated with the Auto Scaling Group
+    lifecycle_hooks=$(aws --region $region autoscaling describe-lifecycle-hooks --auto-scaling-group-name $asg_name --query "LifecycleHooks[*].LifecycleHookName" --output text)
+    if [ -z "$lifecycle_hooks" ]; then
+    echo "No lifecycle hooks found for Auto Scaling Group: $asg_name"
+    exit 1
```
long-wan-ep, line 306: I think these should be indented.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,134 @@ then
...
+    # Retrieve the Auto Scaling Group name associated with the instance
+    asg_name=$(aws --region $region ec2 describe-tags --filters "Name=resource-id,Values=$instance_id" "Name=key,Values=aws:autoscaling:groupName" --query 'Tags[].Value' --output text)
+    if [ -z "$asg_name" ]; then
+    echo "Failed to retrieve Auto Scaling Group name."
+    exit 1
+    fi
+    echo "Auto Scaling Group Name: $asg_name"
+
+    # Retrieve the lifecycle hook names associated with the Auto Scaling Group
+    lifecycle_hooks=$(aws --region $region autoscaling describe-lifecycle-hooks --auto-scaling-group-name $asg_name --query "LifecycleHooks[*].LifecycleHookName" --output text)
```
long-wan-ep, line 303: Do you think it'd be safer to filter to only "instance terminating" hooks? I was thinking it's possible that some users have "instance launching" hooks attached to the ASG.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,134 @@ then
...
+        --auto-scaling-group-name $asg_name \
+        --lifecycle-action-result CONTINUE \
+        --instance-id $instance_id \
+        --region $region
+
+    if [ $? -eq 0 ]; then
+        echo "Lifecycle action completed successfully for hook: $hook_name"
+    else
+        echo "Failed to complete lifecycle action for hook: $hook_name"
+    fi
```
long-wan-ep, line 328: I think these should be indented.

template/gitlab-runner.tftpl
```diff
@@ -246,6 +246,134 @@ then
...
+
+# Function to monitor the existence of the Runner process
+monitor_process() {
+    while true; do
+        systemctl --no-block stop gitlab-runner.service
+        sleep 5
+        status=$(systemctl is-active gitlab-runner.service)
+        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
+          # TODO: maybe also Docker Machine Cleanup here?
+          # Currently the Lambda is resposible for this.
```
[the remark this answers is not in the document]
long-wan-ep: I think you're right that the lambda is just sending the terminate command and not waiting for termination to complete, so leaving it to the lambda should be fine.

long-wan-ep (commented):
Looks good to me overall, just some minor feedback.

template/gitlab-runner.tftpl
tmeijn: Thanks for the suggestion! Updated the query.

tmeijn:
Thanks for the review @long-wan-ep.

@long-wan-ep, @kayman-mk with the last commit I added another lifecycle hook that now truly would make this zero downtime! Previously, the ASG would immediately put the current instance in `Terminating:Wait` state and no longer accept new jobs, causing GitLab jobs to unnecessarily pending state. With this commit, the ASG now waits five minutes before putting the new instance `InService` allowing GitLab Runner to start and provision its set capacity. Only after the new instance is `InService` will it put the current instance in `Terminating:Wait` state. In theory this would allow for a smooth cutover and therefore zero downtime redeployment of the GitLab Runner.

Let me know what you all think. I can easily revert the commit if this is too much and we can address this in a separate MR, but with my testing this really works great!

long-wan-ep:
> Thanks for the review @long-wan-ep.
> 
> @long-wan-ep, @kayman-mk with the last commit I added another lifecycle hook that now truly would make this zero downtime! Previously, the ASG would immediately put the current instance in `Terminating:Wait` state and no longer accept new jobs, causing GitLab jobs to unnecessarily be in a pending state. With this commit, the ASG now waits five minutes before putting the new instance `InService` allowing GitLab Runner to start and provision its set capacity. Only after the new instance is `InService` will it put the current instance in `Terminating:Wait` state. In theory this would allow for a smooth cutover and therefore zero downtime redeployment of the GitLab Runner.
> 
> Let me know what you all think. I can easily revert the commit if this is too much and we can address this in a separate MR, but with my testing this really works great!

That looks good to me, tested it out as well.

kayman-mk:
> Some open questions.
> 
> @kayman-mk, is there a reason we do not have `terraform_docs` in pre-commit, nor as a check in CI?

This is done in the release branch, so the documentation is updated with every release. No need to do it in the feature branch.

versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 5.11"
+      version = ">= 5.26"
```
tmeijn, line 7: **Note:** Support for instance Maintenance policy was released in [5.26](https://github.com/hashicorp/terraform-provider-aws/pull/34430#issuecomment-1815493598)

kayman-mk:
@tmeijn Did a quick check in my test environment, but it was not running as expected.

In case the old Runner ran out of jobs and a new Runner was already there (did a `terraform apply` in the meantime), the old Runner was not removed. Maybe it's waiting for a timeout? When and where is the shutdown triggered? I see the old Runner reporting in CloudWatch that it is still `InService`.

Running `terraform apply` often results in failures as Terraform is unable to modify the autoscaling group in case a Runner is still running. Not sure what we can do here.

tmeijn:
> In case the old Runner ran out of jobs and a new Runner was already there (did a terraform apply in the meantime), the old Runner was not removed. Maybe it's waiting for a timeout? When and where is the shutdown triggered? I see the old Runner reporting in CloudWatch that it is still InService.

Did you wait for more than five minutes? That's how long it takes for the New instance to report InService and subsequently for the ASG to send the Terminate signal to the old instance.

> Running terraform apply often results in failures as Terraform is unable to modify the autoscaling group in case a Runner is still running. Not sure what we can do here.

Maybe because the instance refresh is still progressing?

Would you be able to provide some clear steps in how to reproduce this and I'll take a look ASAP.

kayman-mk:
I did a second test. Looked much better, but not as I expected.

My expectations:
 - after module installation I see one active Runner processing the jobs. 👍 
 - after `terraform apply` (with some changes)
   - a new Runner should appear becoming the only Runner processing jobs
   - old Runner should deregister from GitLab and no longer process jobs. It shouldn't be visible on the Runner page anymore (at least not as online/idle/running)
   - old Runner instance should wait until all jobs are finished
   - as soon as all old jobs are done, the Runner terminates immediately (no matter what the timeout is, as all jobs are done)

tmeijn:
> I did a second test. Looked much better, but not as I expected.
> 
> My expectations:
> 
>     * after module installation I see one active Runner processing the jobs. 👍
> 
>     * after `terraform apply` (with some changes)
>       
>       * a new Runner should appear becoming the only Runner processing jobs
>       * old Runner should deregister from GitLab and no longer process jobs. It shouldn't be visible on the Runner page anymore (at least not as online/idle/running)
>       * old Runner instance should wait until all jobs are finished
>       * as soon as all old jobs are done, the Runner terminates immediately (no matter what the timeout is, as all jobs are done)

@kayman-mk could you also state what is actually happening?

>       * a new Runner should appear becoming the only Runner processing jobs
>       * old Runner should deregister from GitLab and no longer process jobs. It shouldn't be visible on the Runner page anymore (at least not as online/idle/running)

This is not the intended behavior and would also not be the definition of Zero Downtime. Read also about the intended behavior here: https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/1117#issuecomment-2088214950. If the current instance would stop accepting jobs as soon as the new instance is being provisioned, you would have downtime, since there would be no processing capacity until the new instance is fully ready and has provisioned it's own capacity. That is why we wait for five minutes to allow the new instance to start GitLab Runner and provision specified capacity. Only after these five minutes the current instance will start their graceful shutdown logic.

>       * old Runner instance should wait until all jobs are finished
>       * as soon as all old jobs are done, the Runner terminates immediately (no matter what the timeout is, as all jobs are done)

Did you observe something else? If so, what was the actual behavior?

@long-wan-ep can you corroborate any of these findings?

kayman-mk:
@long-wan-ep @tmeijn  Not sure, what happened here. But I now saw the expected behavior. Strange.

I think it would be good to have a simple diagram showing the whole process on a timeline. So we know when the new Runner is registered, deregistered, ...

kayman-mk:
@tmeijn Does it make sense to add an ` aws autoscaling complete-lifecycle-action` as soon as the Runner installation is done? So we don't have to wait for the 300s timeout.

tmeijn:
> @tmeijn Does it make sense to add an ` aws autoscaling complete-lifecycle-action` as soon as the Runner installation is done? So we don't have to wait for the 300s timeout.

Yeah thought about this too, but I do not think there is an easy way to determine that GitLab Runner has provisioned the desired capacity so that is why I opted for a 'dumb' 300s timeout limit.

> I think it would be good to have a simple diagram showing the whole process on a timeline. So we know when the new Runner is registered, deregistered, ...

Sure I can work on this! Could you in the mean time do a code review? 😄

kayman-mk:
Checked my autoscaling group by accident. I found an instance with lifecycle state `Terminating:Wait` and healthy status `Unhealthy`. The Cloudwatch logs have no valuable information (last log is `monitor_runner.sh: Instance target lifecycle state is InService. No action required.`). The EC2 instance is indeed dead and shows `Terminated` on EC2 console.

Not sure how we can get rid of those instances, but waiting for the heartbeat timeout (2h in my case). Any chance for a shortcut here?

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,132 @@ then
...
+main() {
+    # Poll whether the instance is in the terminating state using instance metadata API
+    imds_token=$(curl -sSf -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 300")
+    target_lifecycle_state=$(curl -sSf -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254/latest/meta-data/autoscaling/target-lifecycle-state")
+
+    if [ "$target_lifecycle_state" = "Terminated" ]; then
+        # Monitor the existence of the Runner process
+        monitor_process
+    else
+        echo "Instance target lifecycle state is $target_lifecycle_state. No action required."
```
kayman-mk, line 358: suggestion: Use a debug flag to get a clean log file. We did this already in `user-data.tftpl`.

modules/terminate-agent-hook/variables.tf
```diff
@@ -60,3 +60,14 @@ variable "enable_xray_tracing" {
   type        = bool
   default     = false
 }
+
+variable "asg_hook_terminating_heartbeat_timeout" {
+  description = "Duration the ASG should stay in the Terminating:Wait state."
```
kayman-mk, line 65:
```suggestion
  description = "Duration in seconds the ASG should stay in the Terminating:Wait state."
```

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,132 @@ then
...
+        sleep 5
+        status=$(systemctl is-active gitlab-runner.service)
+        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
+          echo "GitLab Runner service $status. Proceeding with cleanup."
+          send_complete_lifecycle_action
+        else
+          echo "GitLab Runner Serivce still running, sleeping..."
+        fi
+        sleep 5
+    done
```
kayman-mk, line 345:
```suggestion
      systemctl --no-block stop gitlab-runner.service
      
      while true; do
        # it usually takes a while to complete the jobs, so don't poll too often
        sleep 30
        
        status=$(systemctl is-active gitlab-runner.service)
        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
          echo "GitLab Runner service $status. Proceeding with cleanup."
          send_complete_lifecycle_action
        else
          echo "GitLab Runner Service still running, sleeping..."
        fi
    done
```

As far as I understand the docs, `--no-block` does everything but returns immediately. So we can place it outside the loop.

`sleep 30` waits a little bit longer as it will take some time until all jobs are finished and I think it is better to have one line in the logs every 30s instead of every 10 seconds.

versions.tf
kayman-mk: Thanks for pointing this out.

kayman-mk (changes requested):

long-wan-ep:
> @long-wan-ep can you corroborate any of these findings?

I only saw the intended zero downtime behavior you described when I tested, I was using the instance refresh feature in the ASG.

kayman-mk:
Just noticed that I was on 7.0.0. May be the problems where releated to this fact.

I installed the latest version and will give it a new try tomorrow.

tmeijn:
> @long-wan-ep @tmeijn Not sure, what happened here. But I now saw the expected behavior. Strange.
> 
> I think it would be good to have a simple diagram showing the whole process on a timeline. So we know when the new Runner is registered, deregistered, ...

Hey @kayman-mk WYDT about something like this? Do you have a suggestion where to add this?

```mermaid
sequenceDiagram
    autonumber
    participant ASG as Autoscaling Group
    participant CI as Current Instance
    participant NI as New Instance
    ASG->>NI: Provision New Instance (status: Pending)
    Note over NI: Install GitLab Runner <br/>and provision capacity<br/>(5m grace period)
    ASG->>NI: Set status to InService
    ASG->>CI: Set status to Terminating:Wait
    CI->>CI: Graceful terminate:<br/>Stop picking up new jobs,<br/>Finish current jobs<br/>assigned to this Runner
    CI->>ASG: Send complete-lifecycle-action
    ASG->>CI: Set status to Terminating:Proceed
    Note over CI: Instance is terminated:<br/>Cleanup Lambda is triggered
```

kayman-mk:
> Hey @kayman-mk WYDT about something like this? Do you have a suggestion where to add this?

Looks great! Let's add it to [usage.md](https://github.com/tmeijn/terraform-aws-gitlab-runner/blob/feat/enable-graceful-shutdown/docs/usage.md). There is a concept section.

kayman-mk:
> Just noticed that I was on 7.0.0. May be the problems where releated to this fact.
> 
> I installed the latest version and will give it a new try tomorrow.

Looks good to me. Everything worked as expected. Let's tackle the findings from the last review and get it merged.

kayman-mk:
How to proceed here:
- make the diagram available in `usage.md`: @tmeijn 
- review findings were solved a minute ago
- kindly ask you to go over the last commits for a final review: @tmeijn @long-wan-ep 

It looks good from my side. 🚀

variables.tf
```diff
@@ -377,19 +377,21 @@ variable "runner_worker" {
...
   EOT
   type = object({
-    environment_variables = optional(list(string), [])
-    max_jobs              = optional(number, 0)
-    output_limit          = optional(number, 4096)
-    request_concurrency   = optional(number, 1)
-    ssm_access            = optional(bool, false)
-    type                  = optional(string, "docker+machine")
+    environment_variables               = optional(list(string), [])
+    graceful_terminate_timeout_duration = optional(number, null)
```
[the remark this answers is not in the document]
kayman-mk: Oh yeah, I see. Let me fix this. Should go into a separate variable.

template/gitlab-runner.tftpl
```diff
@@ -331,17 +331,19 @@ send_complete_lifecycle_action() {
...
 # Function to monitor the existence of the Runner process
 monitor_process() {
-    while true; do
-        systemctl --no-block stop gitlab-runner.service
-        sleep 5
+      systemctl --no-block stop gitlab-runner.service
+
+      while true; do
+        # it usually takes a while to complete the jobs, so don't poll too often
+        sleep 30
```
long-wan-ep, line 338: 30s feels a little long to me and could cause a bit of delay, how about 10?

long-wan-ep:
@kayman-mk Your commit look good to me overall, just 1 comment.

template/gitlab-runner.tftpl
```diff
@@ -260,6 +260,136 @@ then
...
+
+        status=$(systemctl is-active gitlab-runner.service)
+        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
+          echo "GitLab Runner service $status. Proceeding with cleanup."
+          send_complete_lifecycle_action
+        else
+          echo "GitLab Runner Service still running, sleeping..."
+        fi
+    done
+}
```
tmeijn, line 348:
```suggestion
monitor_process() {
      systemctl --no-block stop gitlab-runner.service
      sleep 5 # grace period to allow GitLab Runner to shutdown
      
      while true; do
        status=$(systemctl is-active gitlab-runner.service)
        if [ "$status" = "inactive" ] || [ "$status" = "failed" ]; then
          echo "GitLab Runner service $status. Proceeding with cleanup."
          send_complete_lifecycle_action
        else
          echo "GitLab Runner Service still running, sleeping..."
        fi
        # it usually takes a while to complete the jobs, so don't poll too often
        sleep 30
    done
}
```

WDYT? Reasoning is that there might not be pending jobs so allow GitLab Runner some time to shut it down. Then we move the sleep back to bottom so the logic executes immediately in first cycle in loop, allowing for a fast shutdown of the instance.

In the grant scheme of things this is a bit bike shedding though, so opinion loosely held.

tmeijn:
Alright, think that's it, just a couple of non-blocking comments for you @kayman-mk!

template/gitlab-runner.tftpl
kayman-mk: Let's go with the 30s (also suggested by tmeijn in his commit below) as a first guess. As the Runner doesn't accept new jobs at this point, we are fine.

kayman-mk (approved):
Guys, good to have you here! This change pushes the project really forward. Thanks @tmeijn @long-wan-ep

PR #1124, opened 2024-05-08
fix: use a valid policy for ssm access

### LINKED ISSUE #1123: Policy template instance-secure-parameter-role-policy.json is invalid
# Describe the bug

The statement:

```json
{
    "Action": [
        "ssm:GetParameters"
    ],
    "Effect": "Allow",
    "Resource": "arn:aws:ssm:*"
}
```

Is not a valid policy statement due to the `Resource` argument using an invalid ARN wildcard. The statement "arn:aws:ssm:*" is disallowed by the IAM policy editor within the AWS console. It also has, albeit inconsistently, failed to apply when running `terraform apply`.

# To Reproduce

1. Instantiate a SSM parameter to store the runner access token.
2. Instantiate the gitlab runner module by passing the above SSM parameter name to the `access_token_secure_parameter_store_name` argument of the `runner_gitlab` configuration block.
3. See error

```bash
module.runner.aws_iam_policy.ssm: Creating...
╷
│ Error: creating IAM Policy (nonprod300520813-ssm): MalformedPolicyDocument: Invalid policy
│ 	status code: 400, request id: f69871f2-4b16-4e1d-a43c-d533fb416a6f
│ 
│   with module.runner.aws_iam_policy.ssm,
│   on .terraform/modules/runner/main.tf line 593, in resource "aws_iam_policy" "ssm":
│  593: resource "aws_iam_policy" "ssm" {
│ 
╵
```

# Expected behavior

The policy generated from the template should be a valid IAM policy that can apply.

# Additional context

I'm not sure why this worked previously, as the [IAM ARN docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-arns) under the "Using wildcards and paths in ARNs" heading states that:

> You can use wildcards in the resource portion of the ARN

The strangest thing is that up until today, I never had issues with this policy being rejected. However, now it is on some of my jobs that deploy a runner, but not all of them. 🤷 

A solution that keeps the same semantics of the existing policy would be:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ssm:PutParameter",
                "ssm:GetParameter",
                "ssm:GetParameters"
            ],
            "Resource": "*"
        }
    ]
}
```

The wildcard will still apply to any parameter stored in SSM, as I believe the previous policy was attempting to do.

ryancausey:
## Description

This resolves an issue where the previous policy template did not specify a valid resource argument for the second policy statement. The modified template should now apply without error.

## Migrations required

No

## Verification

I applied this module to my runner setup and it resolved the issue.

Closes #1123


kayman-mk:
Have seen this today as well. Guess we better limit access to the exact resource.

kayman-mk:
@ryancausey Could you please have a look at my last commit? I think it is better to not allow the Runner to access all SSM parameters. I know, the `*` was there before, but I think we should fix this now.

main.tf
```diff
@@ -598,7 +598,20 @@ resource "aws_iam_policy" "ssm" {
   name        = "${local.name_iam_objects}-ssm"
   path        = "/"
   description = "Policy for runner token param access via SSM"
-  policy      = templatefile("${path.module}/policies/instance-secure-parameter-role-policy.json", { partition = data.aws_partition.current.partition })
+  policy      = templatefile("${path.module}/policies/instance-secure-parameter-role-policy.tftpl", {
```
ryancausey, line 601: I think this might be better served by using the [aws_iam_policy_document data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) rather than a template file now that it is more complex:

```hcl
data "aws_iam_policy_document" "ssm" {
  statement {
    actions = [
      "ssm:GetParameter",
      "ssm:GetParameters",
    ]
    resources = [
      for name in compact(
        [
          aws_ssm_parameter.runner_sentry_dsn.name,
          var.runner_gitlab_registration_token_secure_parameter_store_name,
          var.runner_gitlab.access_token_secure_parameter_store_name,
          var.runner_gitlab.preregistered_runner_token_ssm_parameter_name,
          aws_ssm_parameter.runner_registration_token.name
        ]
      ) : "arn:${data.aws_partition.current.partition}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter/${name}"
    ]
  }

  statement {
    actions = ["ssm:PutParameter"]
    resources = [
      "arn:${data.aws_partition.current.partition}:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter/${aws_ssm_parameter.runner_registration_token.name}"
    ]
  }
}

resource "aws_iam_policy" "ssm" {
  name        = "${local.name_iam_objects}-ssm"
  path        = "/"
  description = "Policy for runner token param access via SSM"
  policy      = data.aws_iam_policy_document.ssm.json

  tags = local.tags
}
```

This would replace the tftpl file entirely.

ryancausey (commented):
I agree that it's better to specify the exact parameters the role has access to, but I left a comment about a possible alternative approach to defining the policy JSON that I believe will be easier to follow and maintain.

main.tf
kayman-mk: Agree. Will commit this in a second. Just running a final test here.

kayman-mk: Looks good.

kayman-mk (approved):
Thanks for fixing this issue.

Issue #1128, opened 2024-05-14
Support for SSM Parameter Store Hierarchy


andreas-mueller-bb:
# Describe the bug

Using the latest version with the preregistered token workflow.

We're applying a hierarchy to our parameter names like described here: https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-hierarchies.html

Unfortunately that doesn't go well with the ssm policy created for accessing that parameter.
When the parameter is called "/tecsafe/prod/preregistered_gitlab_runner_token" the policy looks like this:

`              ~ Statement = [
                  ~ {
                      ~ Action   = [
                          - "ssm:PutParameter",
                          + "ssm:GetParameters",
                          + "ssm:GetParameter",
                        ]
                      ~ Resource = "*" -> [
                          + "arn:aws:ssm:eu-central-1:913735344111:parameter/prod-amd64-sentry-dsn",
                          + "arn:aws:ssm:eu-central-1:913735344111:parameter/prod-amd64-runner-token",
                          + "arn:aws:ssm:eu-central-1:913735344111:parameter/gitlab-runner-registration-token",
                          + "arn:aws:ssm:eu-central-1:913735344111:parameter/gitlab-runner-access-token",
                          + "arn:aws:ssm:eu-central-1:913735344111:parameter//tecsafe/prod/preregistered_gitlab_runner_token",
                        ]
`

# To Reproduce

1. Use a parameter with a leading slash
2. Apply the module
3. Check the ssm policy

# Expected behavior

As using a hierarchy is encouraged by AWS it would be awesome if this would be compatible with the module :)

# Additional context

Thanks for you work!

bwells-scripps:
Here is how we worked around the issue with our local copy. It may not be the best or most correct way, so I haven't submitted it as a PR yet:

```
---
 main.tf | 11 ++++++++---
 1 file changed, 8 insertions(+), 3 deletions(-)

diff --git a/main.tf b/main.tf
index 5a018ab..6b60f85 100644
--- a/main.tf
+++ b/main.tf
@@ -604,7 +604,7 @@ data "aws_iam_policy_document" "ssm" {
       "ssm:GetParameters",
     ]
     resources = [
-      for name in compact(
+      for name in [for nm in compact(
         [
           aws_ssm_parameter.runner_sentry_dsn.name,
           local.runner_gitlab_registration_token_secure_parameter_store_name,
[...]

PR #1146, opened 2024-07-10
fix: remove leading `/` from SSM parameter names in policy ARNs

Kqirk:
## Description

Using SSM parameter hierarchies (as described in https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-hierarchies.html) results in an error. See #1128 

Several workarounds exist, but using `trimprefix` makes them superfluous. Having a `/` as first character results in the above mentioned error.

## Verification

Checked the ARN of parameter `/test/test`. It showed up as `arn:aws:ssm:eu-central-1:123456789012:parameter/test/test` in the console. So no `//` at first place.


kayman-mk (approved):
Thanks for adding this fix to the project and getting rid of all the workarounds.

PR #1184, opened 2024-09-06
feat: allow more than one job per instance

N8ND:
## Description

We want to make the runners_capacity_per_instance variable since we want to be able to run more than 1 job on 1 EC2 instance. 

## Migrations required

No.

## Verification

Using docker_autoscaler, set the capacity_per_instance from 1 to 4 in /etc/gitlab-runner/config.toml, exec gitlab-runner restart <runner>, and verify that only 1 instance is created when running a pipeline with 4 concurrent jobs. Also verified the reported instance in the 4 jobs is the same.


kayman-mk:
Was looking for a while for this feature. Are you using the docker+autoscaler?

N8ND:
> Was looking for a while for this feature. Are you using the docker+autoscaler?

Yes indeed, using docker+autoscaler.

kayman-mk (approved):
Thanks for adding this feature.

PR #1190, opened 2024-10-03
feat: add autoscaler runner instance ready command option

maxvandermeij:
## Description

This adds the optional instance_ready_command parameter for the docker_autoscaler.

This feature was added in Gitlab 16.11. See gitlab documentation for the [runners.autoscaler]: https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runnersautoscaler-section.

Example value:
`instance_ready_command = "cloud-init status --wait"`

## Migrations required

 No

## Verification

Updated the runner_worker_docker_autoscaler by including the instance_ready_command

```
runner_worker_docker_autoscaler = {
    fleeting_plugin_version = "1.0.0"
    connector_config_user = "ec2-user"
    key_pair_name = "runner-worker-key"
    max_use_count = 100
    update_interval = "1m"
    update_interval_when_expecting = "2s"
    capacity_per_instance = var.job_capacity_per_instance
    instance_ready_command = "\"cloud-init status --wait || test $? -eq 2\""
  }
```

Confirmed this gets added correctly to the config.toml by enabling the var.debug_write_runner_config_to_file.

Result on the local_file.config_toml[0]:
```
...
[runners.autoscaler]
      plugin = "fleeting-plugin-aws"
  
      capacity_per_instance = 1
      update_interval = "1m"
      update_interval_when_expecting = "2s"
  
      max_use_count = 100
      max_instances = 1
  
      instance_ready_command = "cloud-init status --wait || test $? -eq 2"
 ....
```

If no value is given for the instance_ready_command the config.toml will default to
```
...
[runners.autoscaler]
    plugin = "fleeting-plugin-aws"

    capacity_per_instance = 1
    update_interval = "1m"
    update_interval_when_expecting = "2s"

    max_use_count = 100
    max_instances = 1
    
    instance_ready_command=""
 ...
```


variables.tf
```diff
@@ -619,6 +620,7 @@ variable "runner_worker_docker_autoscaler" {
     max_use_count                  = optional(number, 100)
     update_interval                = optional(string, "1m")
     update_interval_when_expecting = optional(string, "2s")
+    instance_ready_command         = optional(string)
```
maxvandermeij, line 623: Based on the [comment](https://gitlab.com/gitlab-org/gitlab/-/issues/408131#note_2099525433), we could potentially set the default to "cloud-init status --wait" if preferred

maxvandermeij: or "cloud-init status --wait || test $? -eq 2" according to [discussion](https://gitlab.com/hetznercloud/fleeting-plugin-hetzner/-/issues/3#note_1900817924) about new cloud init version

maxvandermeij: As I am not sure what the best default is, will leave this empty.

maxvandermeij:
Anything I can do additionally to help with the review? @npalm @kayman-mk

variables.tf
kayman-mk: I think, we should use an empty default value. In case you use your own AMI the `cloud-init` might not always be available.

maxvandermeij: Thanks for comment. 
Did some more testing with the empty "" value.  The empyt value seems to work fine for spinning up runners but only if we make sure the value is wrapped to be a string on the config.toml.

kayman-mk:
Thanks for improving the module.

Did you try to start an Agent with the variable unset? Just to make sure that it works with the default value.

maxvandermeij:
> Thanks for improving the module.
> 
> Did you try to start an Agent with the variable unset? Just to make sure that it works with the default value.

Yes did indeed confirm with the new "" default workers spin up perfectly.

kayman-mk (approved):

PR #1221, opened 2024-12-23
fix: allow changes to "runner_worker.max_jobs" for Docker Autoscaler

jonmcewen:
## Description

Remove `min_size` and `max_size` from `ignore_changes` from the ASG for Docker Autoscaler.  The reasoning described in the comment was poor; min_size is fixed at zero, so won't change, and max_size should change when runner_worker.max_jobs is changed, and will not cause immediate scale-up. If it causes scale-down, this the logical desired behaviour.


kayman-mk:
> Just wondering why an ignore changes on `desired_capacity` not required? If the desired capacity of the ASG is set to 3 by the fleeting plugin and the module is applied again, this would result in an immediate scale down, won't it? For the other values I agree with you :)

That came into my mind too.

docker_autoscaler.tf
```diff
@@ -187,13 +187,4 @@ resource "aws_autoscaling_group" "autoscaler" {
...
     }
   }
-
-  lifecycle {
-    # do not change these values as we would immediately scale up/down, which is not wanted
-    ignore_changes = [
-      desired_capacity,
-      min_size,
-      max_size
-    ]
```
kayman-mk, line 197: question: As far as I remember, these values are managed by the autoscaler and shouldn't be touched by us later.

PreNoob: Isn't only the desired size managed by the GitLab runner manager? In fact, testing it, the min size and max size where never changed by the runner manager. 

If I want to adjust the `max_size`, I first need to do that manually in the AWS Console and one additional time in the terraform config so that the runner config is adjusted.

Therefore, I am strongly in favor of removing `max_size` from the ignore list.
From my understanding, `min_size` doesn't matter that much, as the runner manager will keep the runner on the desired size according to the scheduling policy.

zillemarco: `max_size` should also not be ignored. In my situation I deployed the autoscaler config without setting it so defaulted to `0`. Then I changed the `runner_worker.max_jobs` and redeployed, but the change was ignore (hence I had to manually change the value in AWS).

jonmcewen: Now that I understand how the auto-scaling works, I have amended this PR so that only desired_size changes are ignored.  Changes to max_size should be applied based on changes to var.runner_work.max_jobs, and min_size is hard coded.

docker_autoscaler.tf
```diff
@@ -191,9 +191,7 @@ resource "aws_autoscaling_group" "autoscaler" {
   lifecycle {
     # do not change these values as we would immediately scale up/down, which is not wanted
```
jonmcewen, line 192:
```suggestion
    # do not change desired_size as this is controlled at runtime by the runner autoscaler
```

docker_autoscaler.tf
```diff
@@ -189,11 +189,9 @@ resource "aws_autoscaling_group" "autoscaler" {
   }
 
   lifecycle {
-    # do not change these values as we would immediately scale up/down, which is not wanted
+    # do not change desired_size as this is controlled at runtime by the runner autoscaler
```
jonmcewen, line 192:
```suggestion
    # do not change desired_capacity as this is controlled at runtime by the runner autoscaler
```

jonmcewen:
@kayman-mk can this be merged now please?

kayman-mk:
Looks good. Could you please update the PR description to match the change? It's used as commit message.

kayman-mk (approved):

PR #1222, opened 2025-01-06
feat!: add custom egress rules to worker security groups

ikarlashov:
Add custom egress rules to docker-autoscaler security group and remove condition to provision unused docker-machine security group.

## Description

1. By default, the module provisions a security group for Docker Autoscaler workers with egress rules that allow ALL traffic. Unlike ingress rules, egress rules are not customizable, which poses a significant security concern. This PR introduces the ability to customize egress rules for the Docker Autoscaler workers' security group by declaring a separate variable for docker-autoscaler egress rules.
2. `var.runner_worker_docker_autoscaler_asg` becomes bulky and hard to read. Considering the complexity of the ingress rules structure, it makes sense to create a separate variable `var.runner_worker_docker_autoscaler_ingress_rules` for ingress rules. This way we will follow variable convention for existing security group rules variables, i.e. `var.runner_worker_docker_machine_extra_egress_rules`. In the result of the change: `var.runner_worker_docker_autoscaler_asg.sg_ingresses` is removed and its content should be moved to `var.runner_worker_ingress_rules`.
3. Adding `var.runner_ingress_rules` to manage Ingress rules for runner-manager security group.
4. Changed  runner-manager egress rules' var name and its spec.  `var.runner_networking_egress_rules` has migrated to `var.runner_egress_rules` with a new spec.
5. Additionally, PR removes the condition that provisions an unused security group intended solely for Docker Machine setup.

## Migrations required

1. Move all docker-autoscaler ingress rules declaration from `var.runner_worker_docker_autoscaler_asg.sg_ingresses` to `var.runner_worker_ingress_rules`. 
2. `var.runner_networking_egress_rules` migrated to `var.runner_manager_egress_rules` with a new spec.
3. Move `var.runner_worker_docker_machine_extra_egress_rules` to `var.runner_worker_egress_rules`. In case you used multiple cidr_blocks, ... you have to create multiple rules.

Attention: The default value for `var.runner_worker_docker_machine_extra_egress_rules` allowing all egress traffic has been replaced by a rule allowing traffic to port 22/443 only. Don't forget to add the rules you need to `var.runner_worker_egress_rules`.

Attention: Due to the resource replacement you might see inconsistencies and disappearing security group rules. In that case delete all rules from the Runner and Runner Workers and apply the module again.

Sample egress rule:

```
runner_worker_egress_rules = [
     {
      cidr_block    = "0.0.0.0/0"
      from_port     = 443
      protocol        = "tcp"
      to_port          = 443
      description   = "Allow HTTPS egress traffic."
    },
   {
    ipv6_cidr_block = "::/0"
    description         = "Allow HTTPS Egress everywhere"
    from_port           = 443
    protocol              = "tcp"
    to_port                =  443
   }
]
```


docker_autoscaler.tf
```diff
@@ -6,9 +6,47 @@
 resource "aws_security_group" "docker_autoscaler" {
   count = var.runner_worker.type == "docker-autoscaler" ? 1 : 0
 
-  description = "Docker autoscaler security group"
+  name_prefix = "${local.name_sg}-docker-autoscaler"
   vpc_id      = var.vpc_id
-  name        = "${local.name_sg}-docker-autoscaler"
+  description = "Docker-autoscaler security group"
+
+  dynamic "egress" {
```
kayman-mk, line 13: issue: can we move this to an `aws_vpc_security_group_egress_rule` resource? See {aws_security_group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group.html)

docker_autoscaler.tf
```diff
@@ -6,9 +6,47 @@
...
+      from_port        = each.value.from_port
+      protocol         = each.value.protocol
+      security_groups  = each.value.security_groups
+      self            = each.value.self
+      to_port          = each.value.to_port
+      description      = each.value.description
+    }
+  }
+
+  dynamic "ingress" {
```
kayman-mk, line 34: issue: can we move this to an `aws_vpc_security_group_ingress_rule` resource? See {aws_security_group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group.html)

variables.tf
```diff
@@ -741,6 +733,50 @@ variable "runner_worker_docker_autoscaler_role" {
...
+      ipv6_cidr_blocks = ["::/0"]
+      prefix_list_ids  = null
+      from_port        = 0
+      protocol         = "-1"
+      security_groups  = null
+      self             = null
+      to_port          = 0
+      description      = "Allow all egress traffic for Docker-autoscaler runner workers."
+    }
+  ]
```
kayman-mk, line 777: question: any specific reason for this default? As you already wrote, it might introduce a vulnerability.

suggestion: get rid of this default.

security_groups.tf
```diff
@@ -66,7 +66,7 @@ resource "aws_security_group_rule" "runner_ping_group" {
 
 resource "aws_security_group" "docker_machine" {
   # checkov:skip=CKV2_AWS_5:Security group is used within an template and assigned to the docker machines
-  count = contains(["docker+machine", "docker-autoscaler"], var.runner_worker.type) ? 1 : 0
+  count = var.runner_worker.type == "docker+machine" ? 1 : 0
```
kayman-mk, line 69: praise: Good spot!

kayman-mk (changes requested):
Thanks for your work to improve this module. I noticed the major change, but I think we can go on as it is easy to handle for the users.

variables.tf
ikarlashov: I left it for lazy people who don't care about setting specific egresses and just want to have an ability to use the module right away. I.e. to pull docker images from Internet.
I can remove it, no problem.

security_groups.tf
ikarlashov: Thanks! :)

variables.tf
kayman-mk: Sounds good to me, yes.

variables.tf
```diff
@@ -741,6 +733,95 @@ variable "runner_worker_docker_autoscaler_role" {
...
+  type = list(object({
+    cidr_block      = string
+    ipv6_cidr_block = string
+    prefix_list_id  = string
+    from_port       = number
+    protocol        = string # Will be converted to ip_protocol
+    security_group  = string
+    to_port         = number
+    description     = string
+  }))
```
kayman-mk, line 747: question: All attributes are mandatory. Does the validation outlined in the PR really work?

thought: Make `description` and `protocol` mandatory and the others optional.

Same for `egress`.

kayman-mk:
> It would be nice to allow traffic within runner-manager ASG and Docker-autoscaler workers ASG. Basically add this to `security_groups.tf`:
> 
> ```
> resource "aws_vpc_security_group_egress_rule" "runner_manager_to_docker_autoscaler_egress" {
>   count = var.runner_worker.type == "docker-autoscaler" ? 1 : 0
> 
>   security_group_id            = aws_security_group.runner.id
>   from_port                    = 0
>   to_port                      = 0
>   ip_protocol                  = "-1"
>   description                  = "Allow ALL Egress traffic between Runner Manager and Docker-autoscaler workers security group"
>   referenced_security_group_id = aws_security_group.docker_autoscaler[0].id
> }
> ```
> 
> But if I add it, terraform doesn't detect any changes for runner's SG.

Perhaps it is already in place?

variables.tf
ikarlashov: Made `description` and `protocol` required and others - optional.

Validation does work:
```
runner_worker_egress_rules:
  - cidr_blocks:
      - "0.0.0.0/0"
    security_group: "sg-12345" # REDUNDANT
    description: "allow http egress everywhere"
    from_port: 80
    protocol: "tcpp"  # WRONG
    to_port: 80
```

got:
```
│ Error: Invalid value for variable
│ 
│   on variables.tf line 770:
│  770: variable "runner_worker_docker_autoscaler_egress_rules" {
│     ├────────────────
│     │ var.runner_worker_docker_autoscaler_egress_rules is set of object with 8 elements
│ 
│ Protocol must be '-1', 'tcp', 'udp', 'icmp', or 'icmpv6'.
│ 
│ This was checked by the validation rule at variables.tf:805,3-13.
╵
╷
│ Error: Invalid value for variable
│ 
│   on variables.tf line 770:
│  770: variable "runner_worker_docker_autoscaler_egress_rules" {
│     ├────────────────
│     │ var.runner_worker_docker_autoscaler_egress_rules is set of object with 8 elements
│ 
│ At least one destination (cidr_block, ipv6_cidr_block, prefix_list_id, or
│ security_group) must be specified for each rule.
│ 
│ This was checked by the validation rule at variables.tf:813,3-13.
```

ikarlashov:
> > It would be nice to allow traffic within runner-manager ASG and Docker-autoscaler workers ASG. Basically add this to `security_groups.tf`:
> > ```
> > resource "aws_vpc_security_group_egress_rule" "runner_manager_to_docker_autoscaler_egress" {
> >   count = var.runner_worker.type == "docker-autoscaler" ? 1 : 0
> > 
> >   security_group_id            = aws_security_group.runner.id
> >   from_port                    = 0
> >   to_port                      = 0
> >   ip_protocol                  = "-1"
> >   description                  = "Allow ALL Egress traffic between Runner Manager and Docker-autoscaler workers security group"
> >   referenced_security_group_id = aws_security_group.docker_autoscaler[0].id
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But if I add it, terraform doesn't detect any changes for runner's SG.
> 
> Perhaps it is already in place?

If you don't set `var.runner_networking_egress_rules`, then the `runner manager SG` won't have any outbound rules. As a result, the `runner manager` won't be able to connect to the `docker-autoscaler workers`. This PR already includes an ingress rule in the `docker-autoscaler workers SG` to allow traffic between the `runner manager` and the `docker-autoscaler workers SG`. But that's not enough. Specifying an egress rule to "allow SSH" in `var.runner_networking_egress_rules` is still required.

kayman-mk:
Shouldn't we add this egress rule automatically? Otherwise the applied result doesn't work.

ikarlashov:
If I add:
```
resource "aws_security_group_rule" "runner_manager_to_docker_autoscaler_egress" {
  count = var.runner_worker.type == "docker-autoscaler" ? 1 : 0

  type                     = "egress"
  from_port                = 0
  to_port                  = 0
  protocol                 = "-1"
  security_group_id        = aws_security_group.runner.id
  source_security_group_id = aws_security_group.docker_autoscaler[0].id

  description = "Allow ALL Egress traffic between Runner Manager and Docker-autoscaler workers security group"
}
```
the rule will be added to SG upon the first run and removed upon the 2nd. And so on. I've tested multiple times.

This won't work at all:
```
 resource "aws_vpc_security_group_egress_rule" "runner_manager_to_docker_autoscaler_egress" {
   count = var.runner_worker.type == "docker-autoscaler" ? 1 : 0

   security_group_id            = aws_security_group.runner.id
   from_port                    = 0
   to_port                      = 0
   ip_protocol                  = "-1"
   description                  = "Allow ALL Egress traffic between Runner Manager and Docker-autoscaler workers security group"
   referenced_security_group_id = aws_security_group.docker_autoscaler[0].id
 }
```

If we leave the default egress values of `runner_networking_egress_rules`, which is "ALLOW ALL", then user-defined egress rules in `runner_networking_egress_rules` will override "ALLOW ALL" and we still won't get an Egress connection via port 22.

So the only way to add the rule to runner-manager's SG is to refactor [security-group resource](https://github.com/ikarlashov/terraform-aws-gitlab-runner/blob/2b6b53b20d7ec58a57fafea8d480d731f562a6f4/security_groups.tf#L5-L38) by removing all `dynamic` blocks. 

And that will be another breaking change. What do you think?

kayman-mk:
Did a refactoring and notice that we also have `aws_security_group.docker_machine` which has the same problems. I think we should have a `runner_worker_ingress_rules`/`runner_worker_egress_rules` without specifying the machine type as the meaning is the same.

kayman-mk:
I'd like to test this with my current setup before merging it.

kayman-mk:
Checked with docker+machine. 

1. agent was not able to contact the workers --> security group rule added.
2. workers were not able to clone via SSH --> default egress rule extended

@ikarlashov Could you please have a look at my commits?

kayman-mk:
@Alexandre-Orisha Perfectly right! Fixed in 9.0.1

PR #1228, opened 2025-01-20
Upgrading Runner Workers AMI

rofc:
## Description

Upgrading Runner Workers AMI from Ubuntu Focal 20.04 to Ubuntu Noble 24.04

## Migrations required

No


kayman-mk:
Did you deploy this version into your account and checked that it is still working? I don't expect any problems, but just in case...

rofc:
> Did you deploy this version into your account and checked that it is still working? I don't expect any problems, but just in case... 

@kayman-mk it has been working for the last 3 days and no issue has been detected so far

kayman-mk (approved):
Thanks for updating this dependency.

kayman-mk:
Workflow complains about broken link in README.md which is not broken.

PR #1232, opened 2025-01-27
feat: enabled usage of private key with docker autoscaler

zillemarco:
## Description

This PR enables users to explicitly allow using the private key used to authenticate the runner manager with the workers when using the `docker-autoscaler` option.

This is required in order for the manager to be able to connect with the workers running Windows, as noted here: https://gitlab.com/gitlab-org/fleeting/plugins/aws#default-connector-config


zillemarco:
@npalm / @kayman-mk I can confirm that with these changes I am able to connect our runner manager to the workers (issue reported here: https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1172#issuecomment-2613930049).

I tested this on our environment, using my fork as the module's source.

main.tf
```diff
@@ -134,6 +135,8 @@ locals {
 
       runners_instance_ready_command = var.runner_worker_docker_autoscaler.instance_ready_command
 
+      use_private_key = var.runner_worker.use_private_key && var.runner_worker.type == "docker-autoscaler"
```
kayman-mk, line 138: thought: Sounds weired that we use `var.runner_worker.type == "docker-autoscaler"` here as this is the config for the autoscaler. But the same statement can be found above.

kayman-mk (approved):
Thanks for enable the module to use Windows.

variables.tf
```diff
@@ -405,6 +405,7 @@ variable "runner_worker" {
     request_concurrency = Limit number of concurrent requests for new jobs from GitLab (default 1) (request_concurrency).
     ssm_access = Allows to connect to the Runner Worker via SSM.
     type = The Runner Worker type to use. Currently supports `docker+machine` or `docker` or `docker-autoscaler`.
+    use_private_key = Use a private key to connect to the Runner Worker. Do not enable for fleeting.
```
kayman-mk, line 408: question: @zillemarco Is this comment correct?

zillemarco: @kayman-mk you're right that setting only `runner_worker.use_private_key` to `true` won't automatically enable fleeting, for `docker+machine`. With `docker-autoscaler` fleeting is turned on by design.

What do you think about this?

```suggestion
    use_private_key = Use a private key to connect the Runner Manager to the Runner Workers. Ignored when fleeting is enabled (defaults to `true`).
```

kayman-mk (approved):

PR #1235, opened 2025-01-30
ci: replace `GITHUB_TOKEN` by workflow permissions in Megalinter

kayman-mk:
## Description

According to a warning in the logs, we can safely replace the `GITHUB_TOKEN` by workflow permissions.

### DIFF
--- a/.github/workflows/feature_branch-megalinter.yml
+++ b/.github/workflows/feature_branch-megalinter.yml
@@ -1,0 +1,62 @@
+---
+name: Feature branch MegaLinter
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+
+jobs:
+  linter:
+    name: MegaLinter
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/feature_branch-megalinter.yml
```diff
@@ -0,0 +1,62 @@
...
+          # needed to avoid multiple error messages
+          TERRAFORM_TERRASCAN_ARGUMENTS: "--non-recursive"
+          # format issues fail the build
+          TERRAFORM_TERRAFORM_FMT_DISABLE_ERRORS: false
+          # it's an auto-generated file
+          MARKDOWN_MARKDOWNLINT_FILTER_REGEX_EXCLUDE: (CHANGELOG.md)
+          # it's an auto-generated file
+          MARKDOWN_MARKDOWN_LINK_CHECK_FILTER_REGEX_EXCLUDE: (CHANGELOG.md)
+          # automatically commit fixes to the feature branch
+          APPLY_FIXES: all
```
npalm, line 50: If I am correct applied fixes will not trigger any other workflow since the github bot token is used. GItHub doing this to avoid recursive CI loops. But a change could potentially break another pipeline (should. not). Wonder about your thoughts @kayman-mk

kayman-mk: I think so too, yes. But I have never seen any fix from the Megalinter.

PR #1240, opened 2025-02-09
refactor: move runner configuration into separate module

kayman-mk:
## Description

Enables us to define more than one runner per runner manager.

## Verification


kayman-mk:
Still need to check that the config file remains unchanged.

PR #1243, opened 2025-02-16
fix: do not create security group rules for `docker+machine` if `docker-autoscaler` selected

kayman-mk:
## Description

The `runner_worker_ingress_rules` and `runner_worker_egress_rules` have to be attached to the correct security group. At the moment the module always attaches the rules to the docker machine security group which does not exist when `docker-autoscaler` is used.


kayman-mk:
Problem mentioned in https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/1222#issuecomment-2658607360

PR #1250, opened 2025-02-22
docs: add `max_jobs` to docker-autoscaler example

vermut:
Otherwise it deploys unusable asg with max=0


kayman-mk (approved):
Thanks for carification.

PR #1260, opened 2025-03-04
[WIP] fix fleeting plugin installation for docker autoscaler

daveaie:
## Description

Hello everyone,

i have pushed 3 improvement to use docker autoscaler gitlub runner:

- [WIP] install fleeting plugin for gitlab runner 17.x as explained [here](https://docs.gitlab.com/runner/fleet_scaling/fleeting/) : I marked it as WIP because it might be necessary to add a condition on the GitLab Runner version used to apply the correct installation method, but I'm not sure it's needed.
- Fixed the name assignment to distinguish the runner manager from the runner worker in EC2 instances, as requested in this [issue](https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1224)
- add variable to set delete instances (runner worker) on shut down runner manager

## Migrations required

No migration required.

## Verification

I'm using my version of this module in our gitlab on premise installation since a week. 

The reason I implemented this fix is that with the old installation method of the fleeting plugin, the runners did not scale down correctly as indicated in this [issue](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/37526)


template/runner-docker-autoscaler-config.tftpl
```diff
@@ -1,11 +1,14 @@
...
   [runners.autoscaler]
-    plugin = "fleeting-plugin-aws"
+   # TODO make it variable by external variable or by gitlab_runner_version variable
+    plugin = "aws:latest"
 
     capacity_per_instance = ${runners_capacity_per_instance}
     update_interval = "${runners_update_interval}"
     update_interval_when_expecting = "${runners_update_interval_when_expecting}"
 
+    delete_instances_on_shutdown = ${delete_instances_on_shutdown}
```
kayman-mk, line 10: thought: does this make our Lambda function superfluous? We have it to kill the instances on shutdown of the agent.

kayman-mk (changes requested):
Great work!

I think we have to check the installation of the plugin. So I prefer to have this in a separate PR and merge the other changes before,

PR #1264, opened 2025-03-14
fix: add filter to s3 lifecycle rule

### LINKED ISSUE #1263: missing s3 lifecycle rule prefix
# Describe the bug

When i do a terraform plan, i have a warning saying that for the resource aws_s3_bucket_lifecycle_configuration requires a filter with a prefix (capture attached)

# To Reproduce

1. AWS provider version: 5.90.0

Image

a-moussaoui:
## Description

Eliminates terraform warning: `No attribute specified when one (and only one) of [rule[0].prefix.<.filter] is required`

Closes #1263


modules/cache/main.tf
```diff
@@ -56,7 +56,9 @@ resource "aws_s3_bucket_lifecycle_configuration" "build_cache_versioning" {
   rule {
     id     = "AbortIncompleteMultipartUploads"
     status = "Enabled"
-
+    filter {
+      prefix = var.cache_incomplete_lifecycle_prefix
```
kayman-mk, line 60: suggestion: always set this to an empty string to simplify the code and remove the variable. Or was there a specific reason to add one?

```suggestion
      prefix = ""
```

a-moussaoui: Hello, no specific reason, just for flexibility. 
I agree with you that an empty string is sufficient

kayman-mk (approved):
Thanks for quickly fixing this upcoming issue.

PR #1268, opened 2025-03-20
Added example configuration for Windows-based runners

zillemarco:
Added an example configuration to use the module to set up Windows-based runners and an example on how to create a custom Windows AMI usable with the configuration.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -219,6 +219,271 @@ module "runner" {
 }
 ```
 
+### Scenario: Use of Docker autoscaler with Windows runners
+
+Using the `docker_autoscaler` runner worker type, it makes it possible to also run Windows-based jobs in CI.
+
+This way you can use a Unix-based machine as runner manager, and spin up Windows-based runner workers when needed
+to keep costs down.
+
+A few pain-points that needs to be taken into account:
+
+- The private key is necessary in order for the manager to connect to the worker nodes
+- The default user has to be changed to `Administrator`
+- The path resolver must be changed to Powershell
+- The worker nodes must have public IPs
+- If the cache is enabled, the cache path must be changed to path valid for Windows
+- Privileged mode must be disabled because not available on Windows Docker
+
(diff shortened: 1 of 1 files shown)


zillemarco:
Hi @kayman-mk 👋 Here you can finally find the docs and examples we had talked about in https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1233 🙂 

Since I extracted the example from our setup, that's still based on the version `8.1.0` of the module but that should still be compatible with the latest version 🤞

kayman-mk:
What do you think about moving this to a new example in `examples/`? Would be easier for newcomers and the module call is always checked against the latest version. 

Thanks for sharing your configuration.

zillemarco:
> What do you think about moving this to a new example in `examples/`? Would be easier for newcomers and the module call is always checked against the latest version.
> 
> Thanks for sharing your configuration.

That's a good idea @kayman-mk, I agree. Only one thing I'd like to check: those examples are not really deployed, right? Because we'd need to create the Windows AMI for that to work otherwise.

If instead those are just checked with, I don't know, `terraform plan` to make sure the module syntax is correct, then I could just put some placeholder values 🙂

kayman-mk:
Placeholders work fine. The examples are never deployed, right.

nexonen:
Hey @zillemarco - thank you so much for doing this! I've managed to get a windows runner in our infrastructure thanks to your help.

A couple of things that I noticed:
- I had to rename the packer file to have a `.pkr.hcl` extension
- I didn't see `m6a.medium` and `m6i.medium` as valid types - smallest I could find was `.large`
- I didn't find it necessary to set `private_address_only = false` in either the runner or the worker - seemed to work fine without.

Thanks again, really appreciate all the work you did on this!

kayman-mk:
@zillemarco This PR can be merged? I will check the workflows.

kayman-mk (approved):

zillemarco:
@kayman-mk yup, should be good to merge now that the workflows passed 👍

PR #1274, opened 2025-04-10
Use current partition instead of assuming `aws`

joelryan2k:
## Description

Don't hardcode the aws partition

See https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1273

## Migrations required

No

## Verification


jryanmcc:
I cherry picked this against the 9.2.0 tag and verified that it applies against a gov-cloud account.

kayman-mk (approved):
Thanks for fixing this issue.

PR #1280, opened 2025-05-13
feat: add option to suspend process AZRebalance for docker autoscaler as recommended

mmoutama09:
## Description

Add the option to suspend processes. As indicated in the autoscaler doc AZRebalance should be suspended https://gitlab.com/gitlab-org/fleeting/plugins/aws#autoscaling-group-setup

## Migrations required

NO - I just added the option but I do not enforce it

## Verification

Tested with our prod runners.


kayman-mk:
@mmoutama09 Do we really need a separate variable for this? Can we go with a hard-coded `suspended_processes`? Would simplify the setup.

kayman-mk (approved):

PR #1285, opened 2025-06-04
fix: update cache module bucket policy

dmitrijn:
## Description

After upgrading gitlab-runners to 18.0.2 runners can't access shared cache without `s3:ListBucket` permission
Relates to https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1282

also switched to aws_iam_policy_document and added allowance to provide custom policy document

## Migrations required

No

## Verification

After adding `s3:ListBucket` additional permission error is gone.


kayman-mk (approved):
Sounds good to me. I appreciate that you inlined the policy. Looks better now.

PR #1286, opened 2025-06-06
fix: condition on the creation of the kms key

### LINKED ISSUE #1278: Plan fails if KMS key is being created at the same time as the module
# Describe the bug

If the KMS key is being created at the same time as the module, i.e. as a separate `aws_kms_key` resource or by setting `enable_managed_kms_key` the plan fails with this error:
```
╷
│ Error: Invalid count argument
│
│   on .terraform/modules/gitlab-runner/modules/terminate-agent-hook/iam.tf line 53, in data "aws_iam_policy_document" "kms_key":
│   53:   count = var.kms_key_id != "" ? 1 : 0
│
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this,
│ use the -target argument to first apply only the resources that the count depends on.
```

# To Reproduce

Try to use the module like this:
```
module "gitlab-runner" {
  source  = "cattle-ops/gitlab-runner/aws"
  version = "9.2.0"

  environment = "prod"
  runner_gitlab = {
    url                                           = var.gitlab_url
    preregistered_runner_token_ssm_parameter_name = var.token_ssm_parameter_name
  }

  enable_managed_kms_key = true

  subnet_id = var.subnet
  vpc_id    = var.vpc
}
```

# Expected behavior

Terraform plan should not fail

# Additional context

Caused by https://github.com/cattle-ops/terraform-aws-gitlab-runner/pull/1258

schniber:
## Description

this PR fixes #1278 

## Migrations required

No

## Verification

I did deploy this module with both options, i.e. enable_managed_kms_key = true and enable_managed_kms_key = false to validate the fix

### DIFF
--- a/modules/terminate-agent-hook/iam.tf
+++ b/modules/terminate-agent-hook/iam.tf
@@ -189,5 +189,6 @@ resource "aws_iam_role_policy_attachment" "spot_request_housekeeping" {
 
 resource "aws_iam_role_policy_attachment" "aws_lambda_vpc_access_execution_role" {
   role       = aws_iam_role.lambda.name
-  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
+  policy_arn = "arn:${data.aws_partition.current.partition}:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
+
 }
--- a/modules/terminate-agent-hook/iam.tf
+++ b/modules/terminate-agent-hook/iam.tf
@@ -33,14 +33,14 @@ resource "aws_iam_role" "lambda" {
 }
 
 resource "aws_iam_role_policy_attachment" "lambda_kms" {
-  count = var.kms_key_id != "" ? 1 : 0
+  count = !var.enable_managed_kms_key ? 1 : 0
 
   role       = aws_iam_role.lambda.name
   policy_arn = aws_iam_policy.lambda_kms[0].arn
 }
 
 resource "aws_iam_policy" "lambda_kms" {
-  count = var.kms_key_id != "" ? 1 : 0
+  count = !var.enable_managed_kms_key ? 1 : 0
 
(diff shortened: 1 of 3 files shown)


schniber:
@npalm : Hey, is it possible to peer review this ? 
Happy to iterate based on your inputs.

Bests.

modules/terminate-agent-hook/iam.tf
```diff
@@ -33,14 +33,14 @@ resource "aws_iam_role" "lambda" {
 }
 
 resource "aws_iam_role_policy_attachment" "lambda_kms" {
-  count = var.kms_key_id != "" ? 1 : 0
+  count = !var.enable_managed_kms_key ? 1 : 0
```
kayman-mk, line 36: question: if the root module has `enable_managed_kms_key == false`,  the `kms_key_id` was used in the policy. Now the policy attachment is no longer created and the Lambda function is not able to access the key. Anything I overlook?

schniber: Hey Kayman,

yes that's correct, in fact here if we use a string comparison on a variable emptiness, terraform is not able to run the plan which is causing the issue. so the idea was to replace  `count = var.kms_key_id != "" ? 1 : 0` by `count = !var.enable_managed_kms_key ? 1 : 0` consistently.

The variable `enable_managed_kms_key` is now also transmitted from the parent module to the child module to have the count logic working.

Thanks a lot.

schniber:
Hey @kayman-mk,

Any chance of approving this after your review ?

Thanks a lot.

Bests

meniluca:
We are also impacted by #1278 and looking forward to updating the runner to the latest version.

kayman-mk:
This change is now incompatible for those guys using a customer managed key. The module can be updated without setting the new variable to `true` resulting in an outage, I guess or at least using the wrong keys to encrypt the resources.

PR #1294, opened 2025-06-24
docs: correct docker autoscaler usage example

raids:
## Description

Small correction to the docker autoscaler usage example.

## Migrations required

No

## Verification

N/A - docs change


kayman-mk (approved):
Thanks for spotting this one.

PR #1314, opened 2025-08-07
fix: eip race condition when updating ASG

blackillzone:
## Description

Add a second EIP, when we decide to use EIP for the manager. This way, when we update the ASG, a free EIP is available to pick by the `aws-ec2-assign-elastic-ip` tool.

This fix the rolling update, and allow for a smoother transition (instead of trying to work around the removing of the EIP association), and breaking the availability of the runner.

The caveats is that we require two EIP for this use case, but I guess it's okay, as long the user is aware of it.

## Migrations required

No

## Verification

Running a deployment with a runner instance using an EIP, then applying a configuration change for this (like changing the type, or an AMI update).

```
runner_instance = {
  use_eip  = true
}
```

The second runner is starting correctly, and get an EIP associated.

## Alternative

To avoid using two EIP ressources, we may add some more logic in the `template/eip.tftpl` in order to check the EIP availability, and do a switchover if it is already associated to an existing running instance. The issue is that the runner would still be unavailable in the meantime, on Gitlab side, because it will still running the rest of the cloud init script, and won't be able to keep the communication.
I'm open to the discussion about this, but having two EIP is more reliable I think.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -50,7 +50,7 @@ locals {
   file_yum_update = file("${path.module}/template/yum_update.tftpl")
 
   template_eip = templatefile("${path.module}/template/eip.tftpl", {
-    eip = join(",", [for eip in aws_eip.gitlab_runner : eip.public_ip])
+    eip_tags = join(" ", [for k, v in local.tags : "Key=${k},Value=${v}"])
   })
 
   template_gitlab_runner = templatefile("${path.module}/template/gitlab-runner.tftpl",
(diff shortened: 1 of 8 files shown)


blackillzone:
I've updated the PR, and tested all this with `docker-autoscaler` mode. Currently, the runner agent is able to:

**At startup phase**:
- Create a new EIP, if not out of limit (default 10/region/account)
- Associate the EIP to itself

**At terminate phase**:
- Desassociate it's own EIP
- Release the EIP

**Lambda fallback**:
- In case there was any issues in the script at terminate phase (network issue for example), the lambda will cleanup any remaining free EIP, related to the runner agent (based on the tag filtering)

As I was testing in autoscaler mode, I've also include a fix for this: https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1316

It's ready for review, let me know if you think we can enhance anything in the current workflow for the EIP.

**Note:** I've not worked on backward compatibility with the old behavior, but the transition should be easy

PR #1319, opened 2025-09-09
chore: update instance-docker-autoscaler-policy

maxvandermeij:
Updates the instance docker autoscaler policy to match the new recommended iam policy according to the v1.1.0 aws fleeting plugin https://gitlab.com/gitlab-org/fleeting/plugins/aws/-/blob/ed4fb9847d2a83fa05747d40c019f0e6c7b9ebd7/README.md#recommended-iam-policy

## Description

Adds the required iam permissions for the new v1.1.0 version of the gitlab aws fleeting plugin

## Migrations required

Yes or No - If yes please describe the migration. For bigger changes please provide a migration script.

## Verification

### DIFF
--- a/policies/instance-docker-autoscaler-policy.json
+++ b/policies/instance-docker-autoscaler-policy.json
@@ -13,7 +13,8 @@
             "Effect": "Allow",
             "Action": [
                 "autoscaling:DescribeAutoScalingGroups",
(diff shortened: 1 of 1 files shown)


kayman-mk:
That's what I initially thought when reading your PR. But I think it's better to do it separately. Would you please open a new PR for that?

PR #1347, opened 2025-10-20
fix: correct default terminate lambda handler value

smelchior:
## Description

Fixes https://github.com/cattle-ops/terraform-aws-gitlab-runner/issues/1346

Terminate Lambda is not running due to wrong handler name.

## Verification
Trigger instance termination to run lambda to check it runs again properly.


kayman-mk (approved):
Thanks for fixing this. Totally overlooked the setting.

PR #1361, opened 2026-02-09
fix: retry to assign the elastic ip

lukashankeln:
## Description

Assigning the elastic ip for the managing instance will now be retried 7 times.

This avoids issues if aws is not fast enough to release the elastic ip when updating the managing runner.

## Migrations required

No

## Verification

Elastic IP assignment should still work even if aws does not release the elastic ip immediately.

### DIFF
--- a/template/eip.tftpl
+++ b/template/eip.tftpl
@@ -14,4 +14,16 @@ export PATH=~/.local/bin:$PATH
 
 pip install aws-ec2-assign-elastic-ip
 export AWS_DEFAULT_REGION=$(curl -s -H "X-aws-ec2-metadata-token: $token" http://169.254.169.254/latest/dynamic/instance-identity/document | grep region | awk -F\" '{print $4}')
-/usr/local/bin/aws-ec2-assign-elastic-ip --valid-ips ${eip}
+
+for i in {1..7}; do
+  echo "Attempt: $i for ip $eip"
+  if /usr/local/bin/aws-ec2-assign-elastic-ip --valid-ips "${eip}" 2>&1; then
+    break
+  fi
+  if [ "$i" -eq 7 ]; then
+    echo "ERROR: Failed to assign Elastic IP after 7 attempts" >&2
+    exit 1
+  fi
+  echo "Retrying in 60 seconds..."
+  sleep 60
+done


tibuntu:
@kayman-mk any chance this gets a review soon? In the past rotation of the ec2 instances after certain changes went really well for us, until we started to face the issue described above. Currently we need to manually delete the instances to ensure smooth rollout.

template/eip.tftpl
```diff
@@ -14,4 +14,9 @@ export PATH=~/.local/bin:$PATH
...
 pip install aws-ec2-assign-elastic-ip
 export AWS_DEFAULT_REGION=$(curl -s -H "X-aws-ec2-metadata-token: $token" http://169.254.169.254/latest/dynamic/instance-identity/document | grep region | awk -F\" '{print $4}')
-/usr/local/bin/aws-ec2-assign-elastic-ip --valid-ips ${eip}
\ No newline at end of file
+
+for i in {1..7}
+do
+  echo "Attempt: ---- " $i
+  /usr/local/bin/aws-ec2-assign-elastic-ip --valid-ips ${eip} && break || sleep 60
+done
```
kayman-mk:
```suggestion
for i in {1..7}; do
  echo "Attempt: $i for ip $eip"
  if /usr/local/bin/aws-ec2-assign-elastic-ip --valid-ips "${eip}" 2>&1; then
    break
  fi

  if [ "$i" -eq 7 ]; then
    echo "ERROR: Failed to assign Elastic IP after 7 attempts" >&2
    exit 1
  fi

  echo "Retrying in 60 seconds..."
  sleep 60
done
```

Did some optimizations here:
- no pause after the last try
- exit with error if it is still not working after 7 tries
- use `"${eip}"` to avoid word splitting and globbing. Just in case.
- show errors from the command
- log our errors on `stderr`

kayman-mk (changes requested):
I have never seen problems here as I am not using EIP. Thanks for fixing this.

I added some improvements and fixed some older problems in the code. Let me know if you can go with my proposal.

template/eip.tftpl
lukashankeln: Looking good to me, I updated my changes 👍

tibuntu:
@kayman-mk can this be merged soon? Causes real pain within our infrastructure.

kayman-mk (approved):

tibuntu:
@kayman-mk looks like the release pipeline failed

PR #1368, opened 2026-04-02
fix: add missing spot args for docker autoscaler

chrisgurley:
- Update variables for `docker_autoscaler_worker_asg` to expose `spot_price`
- Update `aws_launch_template` to add spot price if given

## Description

This PR updates adds `var.runner_worker_docker_autoscaler_asg.spot_price` as a new valid configuration and updates the Docker Autoscaler launch template to use this value. The launch template for docker autoscaler is updated to use the same logic as `var.runner_instance.spot_price`.

As a note, if you attempt to use both `var.runner_worker_docker_autoscaler_asg.spot_price` and `var.runner_worker_docker_autoscaler_asg.instance_requirements`, the AWS API will reject this as you cannot set instance requirements onto an ASG while a launch template is using spot pricing.

## Migrations required

No

## Verification

- Configure a runner using the module, and for the input variables, `var.runner_worker_docker_autoscaler_asg`, set `spot_price` to a reasonable number.

```hcl
    default_instance_type     = "t3.medium"
    health_check_grace_period = 30
    on_demand_base_capacity   = 0
    spot_price               = "0.17"
    spot_allocation_strategy = "lowest-price"
    subnet_ids               = local.subnets.private
  }
```

- Deploy the runner along with other required settings
- Login to the AWS account where this runner is deployed, and confirm that new spot requests are created when launching


chrisgurley:
/help

Hello, please let me know if there is anything else I can do to help get this PR into an approved state. Thanks!

kayman-mk (commented):
Good idea to amend this. Thanks!

When `spot_price` set AND `enable_mixed_instances_policy=true` with `on_demand_base_capacity > 0`:

   - ASG requests on-demand instances
   - Launch template forces market_type = "spot"

Is this a problem? Have you tried to create this setup? We might have to go with 2 differerent launch templates to fix that, if it is a problem.

PR #1381, opened 2026-06-25
fix: expose on_demand_allocation_strategy to unblock instance_requirements

dmitrijn:
`instance_requirements` (attribute-based instance selection) was already supported
by this module, but applying it would fail at the AWS API level with:

> ValidationError: The specified OnDemandAllocationStrategy is not valid.
> The prioritized allocation strategy is not compatible with instance requirements.
> Valid options are [lowest-price].

This happened because `on_demand_allocation_strategy` was not exposed by the module,
so AWS defaulted to `prioritized`, which it rejects when `instance_requirements` is used.

## Changes

- Add `on_demand_allocation_strategy` to `runner_worker_docker_autoscaler_asg` and
  pass it through to the ASG `instances_distribution` block
- Default to `lowest-price`, consistent with the existing `spot_allocation_strategy` default
- Add a validation that enforces `lowest-price` when `instance_requirements` is set,
  surfacing a clear `terraform validate` error instead of a misleading AWS 400 at apply time

## Usage

To use attribute-based instance selection:

```hcl
  enable_mixed_instances_policy = true
  on_demand_allocation_strategy = "lowest-price"
  instance_requirements = [{
    vcpu_count            = { min = 4, max = 16 }
    memory_mib            = { min = 16384, max = null }
    cpu_manufacturers     = ["amazon-web-services"]
    instance_generations  = ["current"]
    burstable_performance = "excluded"
  }]
}

### DIFF
--- a/docker_autoscaler.tf
+++ b/docker_autoscaler.tf
@@ -98,6 +98,7 @@ resource "aws_autoscaling_group" "autoscaler" {
 
     content {
       instances_distribution {
+        on_demand_allocation_strategy            = var.runner_worker_docker_autoscaler_asg.on_demand_allocation_strategy
         on_demand_base_capacity                  = var.runner_worker_docker_autoscaler_asg.on_demand_base_capacity
         on_demand_percentage_above_base_capacity = var.runner_worker_docker_autoscaler_asg.on_demand_percentage_above_base_capacity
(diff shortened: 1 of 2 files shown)


dmitrijn:
@npalm and @kayman-mk, could you please review this?

PR #1383, opened 2026-07-02
feat: support additional_tags on worker launch template

### LINKED ISSUE #1369: Support additional_tags on worker launch template (docker-autoscaler)
## Problem

When using the `docker-autoscaler` executor, there is no way to add tags exclusively to worker instances without also affecting the runner manager instance.

The module's top-level `tags` variable and the provider's `default_tags` both propagate to all resources — manager and workers alike. The `runner_instance.additional_tags` only applies to the manager ASG, but there is no equivalent for the worker launch template's `tag_specifications`.

## Use Case

AWS GuardDuty Runtime Monitoring can be configured to exclude EC2 instances tagged `GuardDutyManaged: false`. For ephemeral CI workers (short-lived spot instances), GuardDuty scanning generates significant cost with little security value. The manager instance should remain monitored.

Currently, the only options is: Exclude both manager and workers (undesirable — manager should stay monitored)

## Proposed Solution

Add an `additional_tags` field to `runner_worker_docker_autoscaler_instance`:

```hcl
runner_worker_docker_autoscaler_instance = {
  # ... existing fields ...
  additional_tags = { "GuardDutyManaged" = "false" }
}
```

These tags would be merged into the worker launch template's `tag_specifications` (instance, volume, network-interface) alongside `local.tags`.

## Alternatives Considered

- Using `default_tags` at the provider level — affects all resources including the manager
- Using `runner_instance.additional_tags` — only affects the manager ASG, not workers
- Switching GuardDuty to include-mode — requires account-level changes unrelated to the module

lorenzstorm1:
## Summary

Adds an `additional_tags` field to `runner_worker_docker_autoscaler_instance` that applies tags exclusively to the worker launch template's `tag_specifications` (instance, volume, network-interface) without affecting the runner manager instance.

This enables use cases like excluding ephemeral CI workers from AWS GuardDuty Runtime Monitoring while keeping the manager monitored.

### Usage

```hcl
runner_worker_docker_autoscaler_instance = {
  additional_tags = { "GuardDutyManaged" = "false" }
}
```

Closes #1369

## Changes

- `variables.tf`: Add `additional_tags = optional(map(string), {})` to the `runner_worker_docker_autoscaler_instance` object
- `tags.tf`: Add `worker_tags` local that merges `local.tags` with the new `additional_tags`
- `docker_autoscaler.tf`: Use `local.worker_tags` in `tag_specifications` blocks

🤖 Generated with Claude Code

### DIFF
--- a/docker_autoscaler.tf
+++ b/docker_autoscaler.tf
@@ -44,15 +44,15 @@ resource "aws_launch_template" "this" {
 
   tag_specifications {
     resource_type = "instance"
-    tags          = local.tags
+    tags          = local.worker_tags
   }
   tag_specifications {
     resource_type = "volume"
-    tags          = local.tags
+    tags          = local.worker_tags
   }
   tag_specifications {
     resource_type = "network-interface"
-    tags          = local.tags
+    tags          = local.worker_tags
   }
 
   tags = local.tags
--- a/tags.tf
+++ b/tags.tf
@@ -39,4 +39,6 @@ locals {
   runner_tags_string = join(",", flatten([
     for key in keys(local.runner_tags) : [key, local.runner_tags[key]]
   ]))
+
+  worker_tags = merge(local.tags, var.runner_worker_docker_autoscaler_instance.additional_tags)
 }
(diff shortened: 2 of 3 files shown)


lorenzstorm1:
Closing in favour of #1388, which carries this exact commit from our organisation fork instead of my personal one, so colleagues can push to the branch. No content change — please review #1388.

PR #1387, opened 2026-08-11
feat: support additional_tags on docker-autoscaler worker launch template

### LINKED ISSUE #1369: Support additional_tags on worker launch template (docker-autoscaler)

lorenzstorm1:
Closes #1369

## Problem

With the `docker-autoscaler` executor there is no way to tag worker instances without also tagging the runner manager. The top-level `tags` variable and the provider's `default_tags` propagate to everything, and `runner_instance.additional_tags` only reaches the manager ASG — there is no equivalent for the worker launch template's `tag_specifications`.

Our concrete case is AWS GuardDuty Runtime Monitoring, which can be told to skip instances tagged `GuardDutyManaged: false`. Scanning short-lived spot CI workers costs real money for little security value, while the long-lived manager should stay monitored. Today it is all or nothing.

## Change

Adds `additional_tags` to `runner_worker_docker_autoscaler_instance`, merged into the worker launch template's three `tag_specifications` blocks (instance, volume, network-interface):

```hcl
runner_worker_docker_autoscaler_instance = {
  additional_tags = { "GuardDutyManaged" = "false" }
}
```

The merge follows the existing `agent_tags` pattern in `tags.tf` — a `_merged` local, then filtered through `suppressed_tags` for consistency with the other tag locals.

The launch template resource's own `tags` attribute is deliberately left on `local.tags`, since it describes the template rather than the instances it launches.

## Compatibility

Defaults to `{}`, so existing configurations produce a byte-identical plan.

Verified against a two-runner production deployment (x86 + arm64, docker-autoscaler, mixed instances policy): with `additional_tags` unset the plan is unchanged; with it set, only the worker `tag_specifications` gain the tag and the manager launch template does not.

`terraform validate` and `terraform fmt` pass; the README variable table was regenerated with `terraform-docs`.


lorenzstorm1:
Closing as a duplicate of #1383, which implements the same feature and was opened first — I missed it when writing this. Sorry for the noise.

#1383 is the one to review. It is non-draft and mergeable.

Two small things from this branch that #1383 might be worth folding in, entirely at your discretion:

- filtering the merged worker tags through `suppressed_tags`, so the new `additional_tags` behaves like the existing `agent_tags` local rather than bypassing suppression
- regenerating the README variable table with `terraform-docs`

